PR #401, opened 2026-04-16
feat(coding): codebase hygiene, DDD framework structure, and determinism design

visionik:
## Summary

Trimmed PR — codebase hygiene rules + Rule Authority [AXIOM] only. The original DDD framework structure, determinism design doc, and Rule Ownership Map have been transformed into structural artifacts (vBRIEFs under epic #635) by PR-A (#700, merged to master as `1d3490e`); the prose surface that contradicted the determinism thesis is no longer expanded by this PR.

Per the [canonical workflow rationale](https://github.com/deftai/directive/issues/642#issuecomment-4330742436) and Visionik's V4 hierarchy: deterministic > Taskfile > vBRIEF > RFC2119 > prose.

---

## 1. Codebase Hygiene Rules

**New file: `coding/hygiene.md`** — Six sections treating ongoing health as a first-class concern:
- Dead code removal with language-specific tooling (vulture, staticcheck/deadcode, knip)
- Circular dependency detection and resolution (madge `--exit-code`, pydeps, Go compiler)
- Error hiding anti-patterns — explicit legitimate vs. illegitimate try/catch distinctions
- Legacy/deprecated code removal protocol
- DRY with guard against premature abstraction
- Comments: signal vs. noise

**`coding/coding.md`** — DRY + Dependency Direction subsection; error hiding ⊗ rules; Hygiene pointer; 5 new Anti-Patterns

**`verification/verification.md`** — New Legacy and Deprecated Code Detection section

**`core/glossary.md`** — Hygiene Terms section: Hygiene, Dead code, Error hiding, Legacy code, Circular dependency

---

## 2. Hygiene Sections in All Mainstream Language Files

Extended from 3 → 11 language files. Each gets a consistent **Types / Error handling / Dead code** section with language-specific tooling:

| Language | Types | Error Hiding | Dead Code Tool |
|----------|-------|--------------|----------------|
| Python | `Any` return, `# type: ignore` | bare `except: pass` | `vulture` |
| Go | `any`/`interface{}` | `_` ignored errors | `staticcheck`, `deadcode` |
| TypeScript | `as any`, `@ts-ignore` | empty `catch {}` | `knip`, `madge --exit-code` |
| JavaScript | untyped `.json()` | unhandled rejections | `knip`, `madge --exit-code` |
| Rust | `dyn Any`, `#[allow(dead_code)]` | `.unwrap()`, `let _ =` | `cargo clippy`, `udeps` |
[...]

PR #699, opened 2026-04-27
docs(meta/lessons): strengthen GitHub closing-keyword lesson with second-bite incident

MScottAdams:
## Summary

The lesson PR that introduced the GitHub closing-keyword false-positive lesson auto-closed its own umbrella issue a SECOND time on squash merge -- despite the PR body wrapping the offending token in inline code (single backticks) inside a paragraph that was DESCRIBING the prior incident from the PR before it. The original Rule 1 was too soft.

This PR adds Rule 3 to the lesson section, upgrading the prohibition to absolute and embedding a regex self-test as a tripwire so the lesson cannot regress.

This is a docs-only PR. No code changes.

## Discovery

GitHub's closing-keyword parser strips inline code spans (single and double backticks) as plain text for keyword detection. Backtick-wrapping renders the token as code visually, but does NOT hide it from the parser. Only genuine fenced code blocks (triple backticks) are docs-claimed to escape, and even those should not be relied on as a workaround.

## Changes

- **`meta/lessons.md`**: New Rule 3 in the existing closing-keyword section
  - Upgrades the original prohibition to absolute: no closing-keyword token followed by a literal issue number in ANY markdown context (prose, inline code, double-backtick spans, block quotes, negations, citations of prior incidents)
  - Defensive convention: placeholder `#N` / `#NNN` after the verb; separate the verb from the number when citing real prior incidents; use angle-bracket placeholders like `<keyword> #<NNN>` for shape discussion
  - Notes that lessons.md prose is regularly copied into PR bodies and commit messages (where parsing DOES happen), so the rule applies even to non-PR-body documents
  - Embeds a co-located regex self-test as a tripwire: the case-insensitive pattern `(close[ds]?|fix(e[ds])?|resolve[ds]?)\s*#\d+` MUST return zero matches against the section after any edit
- **`CHANGELOG.md`**: `[Unreleased]` -> `### Changed` entry

## Self-test

This PR body, the commit message, and the new lesson section ALL pass the regex self-test (zero matches). Run it yourself:

```text
(close[ds]?|fix(e[ds])?|resolve[ds]?)\s*#\d+
```

against any of the three artifacts above; expect zero hits. The lesson is now self-protecting -- copying any of its prose verbatim into a PR body cannot trigger an unintended auto-close.

## Related Issues

Refs the umbrella tracking issue (no number cited here on purpose, per the new rule). The umbrella was reopened a second time after the prior PR's merge; this PR should NOT close it.

## Checklist

- [x] `/deft:change <name>` -- N/A (2-file docs-only change; not cross-cutting/architectural)
- [x] `CHANGELOG.md` -- entry added under `[Unreleased]`
- [x] `ROADMAP.md` -- N/A (docs lesson, no tracked issue)
- [x] Tests pass locally -- `task check`: 2546 passed, 1 xfailed

## Post-Merge

- [ ] Verify the umbrella tracking issue stayed OPEN after merge. If the parser still bites somehow, the regex self-test will need extension to whatever new pattern slipped through, and Rule 3 will need a Rule 4.


MScottAdams:
--delete-branch

MScottAdams:
Closing as superseded.

While this PR sat unmerged, three things landed on master that change the picture:

1. PR #401 merged and added a Rule Authority [AXIOM] block to `main.md` defining a 3-tier encoding hierarchy: skill (PRIMARY, RFC2119 directives) > Taskfile target (SECONDARY, deterministic) > `meta/lessons.md` (TERTIARY, prose cross-reference only).

2. PR #700 merged with vBRIEF housekeeping.

3. PR #702 merged, codifying a deeper "Layer 3" of the same closing-keyword bug -- the persistent `closingIssuesReferences` link that survives every PR body / commit message / `--subject` / `--body-file` edit. It encodes the rule structurally:
   - `skills/deft-directive-swarm/SKILL.md` Phase 6 Step 1 (PRIMARY)
   - `task pr:check-protected-issues` + `scripts/pr_check_protected_issues.py` (SECONDARY)
   - Thin cross-reference in `meta/lessons.md` (TERTIARY only, deliberately not duplicating the rule body)

This PR is now architecturally inverted: it puts the strengthened rule in `meta/lessons.md` as primary encoding, which the new [AXIOM] explicitly demotes to TERTIARY. The "Layer 2" framing here (inline code spans don't escape the parser) is also strictly subsumed by PR #702's "Layer 3" framing (the link is persistent regardless of body content).

Salvageable bit -- the regex self-test tripwire -- is being filed as a small follow-up issue per the [AXIOM] tier 2 / 3 distinction (tiny pytest content test guarding the lesson section against re-introducing trigger patterns in prose). That part of this PR's work is not duplicated by PR #702.

Issue #1595, opened 2026-06-12
feat(codebase): generate codebase MAP from vBRIEF-owned codeStructure metadata


dbcall2:
> [!IMPORTANT]
> **Override notice (2026-06-15).** The implementation direction in this tracker has been refined by the [current-shape comment](https://github.com/deftai/directive/issues/1595#issuecomment-4712714257), which is the **living source of truth**. Where the original text below and this override conflict, **the override wins**. The original tracker text is preserved verbatim under **"Original tracker text (historical)"** for audit. Do **not** implement the superseded PR 3 / PR 4 scope as originally written (a heavy in-directive crawl/MAP engine), and do **not** create a standalone canonical `codeStructure` file.

### Directive owns contracts, not a heavy map engine

Directive must not hard-depend on any external tool or runtime to produce a codebase MAP — the same principle #1659 enforces for the `task` runner. Directive owns (canonical, in-repo, dependency-free): (1) the `codeStructure` schema, (2) the architecture-map vBRIEF **output format**, and (3) the **freshness/drift protocol**. **Map generation is a pluggable provider** with a **deterministic, dependency-free default extractor** (tree-walk + AST facts only — no network, no model calls) that always works offline. Richer providers MAY be supplied out-of-band, detected/optional; directive MUST degrade to the default when none is present. A missing provider is a graceful downgrade, never an error.

### Canonical home — one home, no standalone file (supersedes "Canonical metadata home" below)

Canonical `codeStructure` lives **only** at `plan.architecture.codeStructure` in `PROJECT-DEFINITION.vbrief.json`, **typed** in `vbrief/schemas/vbrief-core.schema.json` (type locally now; coordinate the upstream `deftai/vBRIEF` landing as a follow-up, mirroring `systemOfRecord` / vBRIEF #23). **No standalone canonical file** — the PR 2 `vbrief/architecture/*.vbrief.json` record must move into `plan.architecture.codeStructure`. A standalone path is allowed **only** for *generated projections* (banner-marked, declared in `projectionManifest[]`, drift-checked or gitignored), never canonical. `x-directive/architecture.codeStructure` remains the consumer-namespace fallback only.

**Boundedness guardrail:** `codeStructure` is canonical *intent* and scales with subsystems, not files. Keep it at subsystem granularity; keep `filePurposeOverrides[]` sparse. If it trends large that is a canonical/derived-line violation to fix — or a trigger for a typed sub-document / `$ref` mechanism, **never** sibling files.

[...]

Issue #1498, opened 2026-06-04
RFC: Directive must dogfood vBRIEF-as-source-of-truth -- source code & headers are projections, not stores


MScottAdams:
## Principle

deft's chosen direction (#1284, approved) makes the **vBRIEF the source of truth** and treats generated `.md` artifacts as rendered views (#336; #308 already did this for ROADMAP.md / PRD.md). Taken to its conclusion: **source code is itself a projection of canonical intent** — it can be regenerated, even in another language — so a **file header is an output of the build, never an input or a store.** Any datum that must survive a language rewrite cannot live in a source file; it must live in the vBRIEF, or be re-derivable from the code's AST.

## Problem

Directive does not yet live up to its own principle:
- Structural / design metadata (module ownership, architectural pattern, file → purpose) has **no canonical home** in the vBRIEF today; it lives implicitly in hand-authored code.
- Proposals keep trying to store authoritative metadata in the disposable layer — e.g. #958's hand-authored file headers — which directly contradicts the canonical direction.
- Without a canonical home plus a projection discipline, every consumer reinvents this and drift is guaranteed.

This issue tracks making **directive itself conform**: canonical structural metadata in the vBRIEF, code-adjacent artifacts (maps, headers, docs) as **regenerable, drift-checked projections**, and a defined **brownfield extraction** path.

## Direction of truth

- **Canonical (vBRIEF / PROJECT-DEFINITION):** design + structural metadata — module registry, ownership, architectural pattern, `file_scope` (already on swarm metadata), domain terminology (already via the glossary skill + `core/glossary.md`).
- **Re-derivable (code AST):** imports / coupling, language, call graph — computed on demand, never stored.
- **Projections (generated, drift-checked):** rendered `.md` (SPEC / PRD / ROADMAP today; a codebase MAP per #958 next) and, *optionally*, in-file headers — emitted by the build, validated against canonical via the existing `check_render_staleness` pattern, never hand-authored.

## Brownfield ingestion (key surface)

[...]

Issue #958, opened 2026-05-07
feat: Agent-centric codebase conventions (file headers, repo map, glossary-driven comments)


ZaenKai:
## Summary

Deft governs how AI agents *behave* but says little about how the *codebase itself* should be structured to help agents work efficiently. This proposal adds three conventions to the framework that reduce context window waste, improve search accuracy, and accelerate agent orientation in unfamiliar codebases.

## Problem

1. **No file-level metadata convention.** Agents must read 50-500 lines of a file to determine relevance. No "peek" mechanism exists.
2. **No machine-readable codebase map.** Agents spend 3-8 tool calls at session start just orienting themselves. The existing `strategies/map.md` produces human-readable docs but nothing an agent can parse cheaply.
3. **No comment quality standard for agent consumption.** Vague comments ("handles the thing") defeat vector search and force full-file reads.

### 1. Structured File Headers

Every new source file gets a structured header in the language's idiomatic format (Python module docstring, JSDoc, Go doc comment, C++ block comment) containing:
- **Purpose**: one sentence
- **Module**: domain tag matching the project's module registry
- **Dependencies**: key imports that define coupling
- **Pattern** (SHOULD): interaction pattern (singleton, handler, state machine, etc.)

Enforced by `task verify:headers` wired into `task check`. Exemptions via `.header-exempt` glob file.

### 2. Generated Codebase Map

`task codebase:map` walks the source tree, reads file headers, and produces `.planning/codebase/MAP.md` -- a condensed structural summary. Follows the existing "source produces rendered view" pattern. Staleness detected by `deft-directive-sync` at session start.

Flags: `--format md|json`, `--depth N`, `--diff` (incremental), `--include-tests`.

### 3. Glossary-Driven Comment Semantics

New `terminology` narrative on `PROJECT-DEFINITION.vbrief.json` defines project-specific domain vocabulary. `coding/coding.md` gains a `## Comment Semantics` section at ~ (SHOULD) level. `task codebase:glossary` extracts and proposes terms interactively.

## vBRIEF Integration

- **No schema changes.** All three use existing narratives (`headerConvention`, `moduleRegistry`, `terminology`) on PROJECT-DEFINITION.vbrief.json.
- **Brownfield retrofit** tracked as scope vBRIEFs through normal lifecycle folders.
- **Generated artifacts** follow the existing source-to-rendered-view pattern.

## New Tasks

- `task codebase:map` -- generate repo map
[...]

PR #1607, opened 2026-06-12
docs(codebase): record map source-of-truth decision

dbcall2:
## Summary

Records the #1595 PR 1 decision that durable codebase-structure metadata belongs in vBRIEF-owned architecture metadata, while generated codebase MAPs and optional file headers are projections.

## Related Issues

Refs #1595, #958, #1498.

## Checklist

- [x] `/deft:change <name>` - N/A; docs/vBRIEF-only PR 1 slice scoped by #1595 and confirmed with explicit implementation directive.
- [x] `CHANGELOG.md` - added entry under `[Unreleased]`.
- [x] `ROADMAP.md` - N/A; #1595 remains the multi-PR tracker.
- [x] Tests pass locally.


MScottAdams:
## Architecture feedback — aligns with RFC #1618, two corrections on the metadata home

This decision record gets the **projection half exactly right** — "generated artifacts must point back to the metadata that produced them" with **one-directional drift checks** (projection compared to truth, never the reverse), and the explicit refusal to let hand-authored headers become authoritative. That is the same source-of-truth / materialized-view separation argued in #1618, independently arrived at. No objection to landing this as a decision record.

Two corrections worth folding into the **PR-2 hand-off** (before the `codeStructure` schema is frozen):

**1. Split the truth out of the monolith.** The "Metadata Home" section puts `codeStructure` into `plan.architecture.codeStructure` inside `PROJECT-DEFINITION.vbrief.json`. That file is already 511 KB / 10,831 lines and ~97% of it is the `plan.items` registry, which has already drifted ~10% from filesystem truth (#1618 has the measurements). Adding a second large authored section to a file that #1618 argues should be *shrinking* moves the wrong direction. `codeStructure` authored intent (modules, ownership, allowedPatterns, overrides) should live in its **own git-tracked file** (e.g. `vbrief/architecture.vbrief.json` or per-module files), not a new section of the monolith.

**2. `codeStructure` introduces a new scaling axis #1618 doesn't have.** `plan.items` scales with the *number of work items*; `codeStructure` scales with the *size of the codebase*. `modules[]` / `pathOwnership[]` grow with module count, and `filePurposeOverrides[]` is potentially **O(files)** — a code-size-proportional blob welded into one JSON in a large monorepo. The `*Overrides` naming is the right instinct (store human *exceptions*, treat inferred purposes as derived) — please keep that discipline explicit in the schema so it can't quietly become a full per-file registry. The **code-derived half** (file→module inference, MAP contents, freshness/drift state) is exactly the rebuildable local-index workload proposed in #1618 (`.deft/index.db`, gitignored) — it should never touch git-tracked authored truth.

Net: philosophically aligned and safe to merge as a decision; the home + scaling concerns are design-time corrections for the schema PR, not blockers here. Cross-linking #1618 as the companion source-of-truth/projection thesis.

Refs #1618.

Issue #1577, opened 2026-06-09
triage:welcome uses unprefixed task names and fails in consumer Taskfile namespace


MScottAdams:
## Summary

`task triage:welcome -- --onboard` fails in a consumer install when the framework Taskfile is included under a namespace (for example `deft:`). The welcome script shells out to unprefixed task names, but the consumer-visible tasks are namespaced.

Observed in `deftai/cartograph` after upgrading to Deft Directive `v0.44.0`.

## Observed

From the consumer repo root, the documented/onboarded command was run via the installed include surface:

```bash
task deft:triage:welcome -- --onboard
```

The flow reached phase 3, then failed because the Python script invoked an unprefixed task:

```text
[3/6] Running `task triage:bootstrap`...
task: Task "triage:bootstrap" does not exist
  ! `task triage:bootstrap` exited 200; see stderr above.
```

It later reached phase 6 and failed similarly:

```text
[6/6] Final state:
task: Task "triage:summary" does not exist
task: Failed to run task "deft:triage:welcome": task: Failed to run task "deft:triage-welcome:welcome": exit status 2
```

In this consumer layout, the available task names are:

```bash
task deft:triage:bootstrap
task deft:triage:summary
```

The unprefixed names do not exist at the project root.

## Impact

This is first-run friction for the new session-start/onboarding flow:

- The command partially mutates policy state before failing.
- The operator is left with an incomplete bootstrap and no final summary.
- In the Cartograph run, the scope/cap choices had to be corrected manually, then `task deft:triage:bootstrap` and `task deft:triage:summary` had to be run directly.

## Expected

`triage:welcome` should resolve and call sibling tasks correctly from both framework-maintainer and consumer-installed layouts.

Possible fixes:

- In `scripts/triage_welcome.py`, call the currently exposed namespace when running inside a consumer include, e.g. `task deft:triage:bootstrap` / `task deft:triage:summary`.
- Or avoid shelling out through `task` for sibling phases and call the underlying Python functions/modules directly.
- Or pass the task namespace into the script from the Taskfile wrapper and use that consistently.

## Acceptance criteria

- In a consumer project where `.deft/core/Taskfile.yml` is included under `deft:`, `task deft:triage:welcome -- --onboard` completes all six phases without `task: Task "triage:*" does not exist`.
- In the Directive maintainer repo, the existing unprefixed `task triage:welcome -- --onboard` behavior continues to work.
- Tests cover both direct/framework and consumer-namespaced Taskfile layouts.
[...]

Issue #1523, opened 2026-06-05
bug(docs/templates): AGENTS.md ritual + main.md instruct bare task names that do not resolve under the canonical deft: namespaced include


MScottAdams:
## Summary

In a consumer project that follows the **canonical** root-`Taskfile.yml` include pattern documented in `main.md` ("Publishing deft tasks in your project root"), every **bare** `task <name>` command that the generated `AGENTS.md` managed section (and `main.md`) instructs the agent/operator to run fails with:

```
task: Task "<name>" does not exist
```

The canonical include namespaces all deft tasks under `deft:`, so the actually-resolvable names are `task deft:<name>`. But the docs/templates uniformly say `task <name>`. This is an internal inconsistency in directive's own canonical docs, not a consumer misconfiguration.

## Environment

- Consumer: `deftai/statusreport`
- Deft CLI: **v0.42.0** (sha `fffdfad`), `install_root: .deft/core`
- OS: Windows 11, pwsh 7.6.2, Task (go-task) v3, uv
- `task deft:doctor -- --session` → all green (install integrity, `skill-paths-resolve`, `manifest-agreement`, `agents-md-managed-section-fresh`). So this is a healthy, up-to-date install.

## Reproduction

1. Install directive into a consumer project and wire the root `Taskfile.yml` exactly as `main.md` recommends:

   ```yaml
   version: '3'
   includes:
     deft:
       taskfile: ./.deft/core/Taskfile.yml
       optional: true
   ```

2. Start a session and follow the `AGENTS.md` "Session-start ritual (#1149)" verbatim:
   - `task doctor` → `task: Task "doctor" does not exist`
   - `task triage:welcome` → `does not exist`
   - `task verify:cache-fresh` → `does not exist`
3. Prefix with the namespace and every one resolves: `task deft:doctor`, `task deft:triage:welcome`, `task deft:verify:cache-fresh`.

(go-task always exposes an included Taskfile's tasks as `<include-key>:<task>`. With the include key `deft:`, the only resolvable form is `task deft:<task>`. There is no bare form.)

## Root cause — directive's docs are inconsistent with directive's own recommended include

The recommended include namespaces under `deft:`, yet the agent-facing docs reference bare task names throughout:

- **`templates/agents-entry.md`** (source of the consumer `AGENTS.md` managed section) uses bare names everywhere:
  - Session-start ritual (#1149): `task doctor`, `task triage:welcome`, `task verify:cache-fresh`
[...]

Issue #11, opened 2026-02-04
Proposal: Publish Deft as an NPM + PIP CLI (agent-first distribution)


orlandobrent:
### Motivation

Today Deft is typically consumed via `git clone`, which makes installation and updates heavier than they need to be (especially across many repos/machines). Publishing Deft as package-manager CLIs provides:

- **Easy install** (`npm i -g ...`, `npx ...`, `pipx install ...`, `uv tool install ...`)
- **Version pinning** (package version == Deft version)
- **Consistent entrypoint** (`deft ...`) regardless of environment
- **No git workflow friction** (no submodules, no clone required)

### Desired user experience

Users should be able to install Deft with either ecosystem:

#### NPM (global / one-off / pinned via Volta)

**Global install**
- `npm i -g <pkg>`
- `deft --version`

**One-off execution**
- `npx <pkg> <command>`

**Project dev dependency**
- `npm i -D <pkg>`
- `npx deft <command>`

**Volta (pinned, per-user)**
- `volta install <pkg>`
- `deft --version`

**Volta (pinned, per-project)**
- Add to `package.json`:
  - `"volta": { "node": "x.y.z", "npm": "a.b.c" }`
- `npm i -D <pkg>`
- `npx deft <command>`

> Note: If we want "pin Deft CLI version" via Volta specifically, we can recommend `volta install <pkg>@<version>`.

#### PIP (recommended via uv tool / pipx)

**uv tool install (pinned, per-user)**
- `uv tool install <pkg>`
- `deft --version`

**pipx install (pinned, per-user)**
- `pipx install <pkg>`
- `deft --version`

**pip install (fallback)**
- `pip install <pkg>`
- `deft --version`

### Why keep a PIP rail

PIP is justified only if it serves users not well covered by webinstaller + npm. The strongest cases are:

- **Python-native teams** that already standardize on `pipx` / `uv tool install` for developer CLIs
- **Enterprise allowlists** where Python package distribution is approved but npm is not
- **Internal package mirrors** where PyPI-compatible distribution fits existing tooling better than direct GitHub or npm-based install flows

PIP is **not** proposed just for symmetry with npm. If shipped, it should exist because it solves a real adoption/distribution problem for those users.

In all cases, users get a `deft` command with the same behavior.

### Example CLI usage (what users would actually run)

Assuming the project-local layout and behavior is already solved (see https://github.com/visionik/deft/issues/4), the CLI should support workflows like:

**Install / initialize in a repo**
- `deft install`
[...]

PR #1655, opened 2026-06-15
fix(ritual): run session gates in-process

### LINKED ISSUE #1648: Gated session ritual invokes unprefixed task names (doctor, verify:cache-fresh) that don't resolve in consumer projects
## Summary

In a consumer project that vendors the framework under `.deft/core` and exposes tasks via a root `Taskfile.yml` include under the `deft:` namespace (the canonical pattern from `main.md`), the **gated session ritual** fails because it invokes **unprefixed** task names that do not exist in the consumer's task namespace.

Observed while running the pre-`start_agent` gate stack in `deftai/statusreport`:

- `task deft:vbrief:preflight -- <path>` runs the session-ritual gate first.
- The gated ritual's `doctor` step shells out to `task doctor` (unprefixed) → consumer only has `task deft:doctor` → Task prints "Available tasks for this project" and exits non-zero.
- After deferring `doctor`, the next step fails identically: the `cache_fresh` step invokes `task verify:cache-fresh` (unprefixed) → consumer only has `task deft:verify:cache-fresh`.

Net effect: the gate stack (session-ritual → story-ready → vbrief preflight → cache-fresh → branch) cannot pass in a consumer project without deferring `doctor` and `cache_fresh` via `task deft:session:start -- --defer ...`. The defers are an audit workaround, not a fix.

## Environment

- Consumer repo: `deftai/statusreport` (vendored install, `.deft/core`)
- Framework payload: v0.46.0 (sha 9c7de81), `install_root: .deft/core`
- Root `Taskfile.yml` includes `./.deft/core/Taskfile.yml` under `deft:` (the documented pattern)
- OS: WSL2 Ubuntu

## Repro

1. In a consumer project using the `deft:`-namespaced include, ensure a vBRIEF is in `vbrief/active/` with `plan.status == running`.
2. Run `task deft:vbrief:preflight -- <active-vbrief-path>`.
3. Observe: `Session ritual gate failed before #810 lifecycle check: session ritual gated step 'doctor' failed: task: Available tasks for this project: ...`
4. Defer doctor; re-run; observe the same failure for gated step `cache_fresh` (`task verify:cache-fresh`).

## Expected

The gated ritual should resolve framework tasks through the same namespace the consumer uses (e.g. honor a configured prefix / `deft:` include), or call the wrapped scripts directly (as the preflight target itself does via `uv --project .deft/core run python ...`), rather than assuming unprefixed `task <verb>` resolves in the consumer.

## Suggested fix directions

- Have the ritual sub-steps invoke the underlying Python entrypoints directly (consistent with how `deft:vbrief:preflight` resolves its wrapped script) instead of re-entering `task <verb>`.
- Or make the prefix discoverable/configurable so the ritual can call `task deft:doctor` / `task deft:verify:cache-fresh` in a `deft:`-namespaced consumer.

## Workaround in use

`task deft:session:start -- --defer doctor="..." --defer cache_fresh="..."` then re-run preflight. Functional but records permanent defers for steps that should simply resolve.

Related friction class: see also the consumer-side `.deft/ritual-state.json` gitignore issue filed previously (#1609).

dbcall2:
## Summary
- Convert the gated session ritual from nested `task <prefix>:...` re-entry to in-process Python entrypoint calls.
- Record `doctor.cmd_doctor` and `preflight_cache.main --allow-missing-bootstrap` in ritual state so the gate preserves the operator-facing `task doctor` / `task verify:cache-fresh` behavior without depending on Taskfile namespaces.
- Remove the now-unused Taskfile namespace resolver and prefix plumbing from session/verify ritual surfaces.
- Update AGENTS/template guidance, tests, CHANGELOG, and the completed vBRIEF for the #1659 canonical pattern.

## Verification
- `UV_FROZEN=1 task verify:branch`
- `UV_FROZEN=1 task verify:session-ritual -- --tier=gated`
- `UV_FROZEN=1 task check` (`8188 passed, 5 skipped, 9 deselected, 1 xfailed`)

## Review notes
- Scott's requested in-process-entrypoint conversion is implemented.
- `doctor.cmd_doctor([...])` is used instead of `doctor.main([...])` because `cmd_doctor` is the canonical `task doctor` behavior surface; `doctor.main` is the legacy framework-doctor surface.
- Cache freshness keeps the wrapper behavior Scott called out by passing `--allow-missing-bootstrap`.

Closes #1648
Refs #1659


MScottAdams:
Thanks for the fast turnaround on this, @dbcall2 — the root-cause writeup and the test coverage are exactly right, and this genuinely unblocks the `deft:`-namespaced consumer path that was forcing the defer workaround.

Two things converged *after* you opened this, and they're on me, not you. About two hours after this PR went up I filed umbrella #1659 ("Decouple framework runtime + guidance from the `task` runner") for npm/pip readiness (#11), with #1648 as its first child. The original issue legitimately listed both fix directions as acceptable — you picked a sanctioned one — but #1659 now pins the canonical pattern:

> invoke the underlying Python entrypoints in-process rather than reconstructing `task <prefix>:<verb>`

There's also a pattern worth surfacing: this is the third instance of the same bug class. #1523 (bare task names in docs) and #1577 (`triage:welcome` unprefixed) both got prefix-style instance fixes; #1577 was registered complete on Jun 11 (#1593), and the same class recurred as #1648 four days later. Prefix-based fixes — whether the `DEFT_TASK_PREFIX` env var (#1577) or Taskfile-include discovery (this PR) — keep the `task`-runner coupling intact and haven't held. That recurrence is exactly why #1659 mandates removing the coupling rather than re-computing the prefix.

Concretely, for the two files this PR touches (both named as Class-A items in #1659):

- `scripts/verify_session_ritual.py` gated steps → call `doctor.main([...])` / `preflight_cache.main([...])` in-process (both expose `main(argv)`) instead of re-entering `task {prefix}:doctor` / `task {prefix}:verify:cache-fresh`.
- `scripts/session_start.py` → same in-process treatment for the ritual sub-steps.

The namespace resolver itself is nice work, and there may be a place for it as a Taskfile-only convenience — but it shouldn't be the mechanism the gate depends on, and #1659 sets an acceptance criterion that new `subprocess(["task", ...])` call sites should fail CI, which this path would eventually trip.

Would you be up for converting this PR to the in-process-entrypoint approach as #1659's child #1? It's a high-leverage place to land it — it sets the canonical pattern the rest of that umbrella follows. Happy to pair on the entrypoint signatures.

dbcall2:
Yes, we can do that! And ty for the very thorough explanation.

dbcall2:
@MScottAdams thanks again for the detailed direction here. I converted this PR away from the Taskfile namespace-resolution approach and onto the #1659 in-process-entrypoint pattern.

What changed:
- `scripts/verify_session_ritual.py` now runs gated steps through Python entrypoints instead of reconstructing `task <prefix>:...`.
- `cache_fresh` calls `preflight_cache.main([...])` and preserves the framework wrapper behavior by passing `--allow-missing-bootstrap`.
- `doctor` uses `doctor.cmd_doctor([...])` rather than `doctor.main([...])` because `cmd_doctor` is the canonical `task doctor` behavior surface; `doctor.main` is the legacy framework-doctor path and failed this source checkout.
- `scripts/session_start.py` no longer threads or records task prefixes for the ritual sub-step.
- The namespace resolver and resolver-only tests were removed since the gate no longer depends on that mechanism.

Verification:
- `UV_FROZEN=1 task verify:session-ritual -- --tier=gated`
- `UV_FROZEN=1 task check` (`8188 passed, 5 skipped, 9 deselected, 1 xfailed`)

Issue #1729, opened 2026-06-18
feat(ts-migration): port the release pipeline (release*.py) to TS (Wave 4)


MScottAdams:
## Parent

#1530 (Part 1 of 2; umbrella epic #1545)

## What to build

Port the release pipeline (`scripts/release*.py` -> `deft release:*`: release / publish / rollback / e2e, CHANGELOG section extraction, GitHub release body flow) to TS.

## Acceptance criteria

- [ ] `release:*` verbs run on TS with the publish/rollback/e2e flow preserved
- [ ] CHANGELOG `[version]` section extraction + release-body flow reproduced
- [ ] Golden-output parity vs the Python oracle (cache-off)
- [ ] `task check` passes

## Standing invariants (all Part-1 children)
- The Python suite is the **parity oracle** and runs **wholesale** (not affected-sliced) until deleted.
- `task check` stays green at every step (strangler-fig; no big-bang).
- The parity gate runs **cache-off** so a golden diff is never served from cache.
- No Python module is deleted until its TS replacement covers historical bug fixtures (not just happy-path) and CI usage.
- No agent-runtime dependency (that is Part 2, #1707); no monorepo orchestrator on day one (Nx never; Turborepo only on a measured trigger).

## Type

AFK

## Blocked by

- Blocked by #1725
- Blocked by #1726
- Blocked by #1727
- Blocked by #1728

MScottAdams:
All child stories ported, wired, and merged (release core/publish/rollback/e2e). Wave 4a complete; closing umbrella. Refs #1530.

MScottAdams:
## Current shape (as of pass-1)

Last updated: 2026-07-13T22:14:29Z
Last pass type: additive
Child count: 4 (0/4)
Child-count history: pass-1: 4

### Open children

- none

### Closed children

- 1729-s1-release-core: Port the release.py core (version bump + CHANGELOG section extraction + GitHub release-body flow) to TypeScript (completed)
- 1729-s2-release-publish: Port release_publish.py (release:publish flow) to TypeScript (completed)
- 1729-s3-release-rollback: Port release_rollback.py (release:rollback flow) to TypeScript (completed)
- 1729-s4-release-e2e: Port release_e2e.py (release:e2e rehearsal) to TypeScript (completed)

### Wave order

- Wave 1: 1729-s1-release-core
- Wave 2: 1729-s2-release-publish, 1729-s3-release-rollback, 1729-s4-release-e2e

### Open questions

- none

[...]

Issue #1530, opened 2026-06-05
RFC: Migrate directive's enforcement engine from Python to TypeScript/Node.js (Part 1 of 2)


MScottAdams:
# RFC: Migrate directive's enforcement engine from Python to TypeScript/Node.js (Part 1 of 2)

**Status**: Draft for Discussion (rev 3 — tooling sub-decision)
**Date**: 2026-06-05 (rev 2: 2026-06-17; rev 3: 2026-06-18)
**Labels**: rfc, design, process, determinism, agent-experience, runtime-portability

> **Split notice (rev 2).** This RFC was originally a single proposal that bundled two separable decisions. It has been split so the language/engine move can proceed without blocking on the runtime question:
> - **Part 1 — this issue:** migrate the deterministic enforcement/tooling engine from Python to **TypeScript/Node.js** (plain Node, no monorepo orchestrator on day one). **No agent-runtime dependency.** Actionable now.
> - **Part 2 — #1707:** whether directive should *own* a durable agent runtime for its orchestration layer (RivetOS is the leading candidate). **Deferred** — a separate decision to be made after Part 1 lands.
>
> Both parts sit under umbrella epic #1545. The earlier "on the RivetOS agent runtime" framing now lives entirely in Part 2.

> **Tooling note (rev 3).** The earlier "plain Node + Nx" framing is **struck**. Day-one tooling is pnpm workspaces + `tsc -b` + `vitest --changed`; incrementality is provided by the consumer-reachable in-engine content-hash cache (#1713). A monorepo orchestrator is **not** adopted on day one, and if a measured trigger ever forces one it is **Turborepo, never Nx**. See "Tooling sub-decision" below.

> Notation: this RFC uses the deft RFC2119 markers — **!** MUST, **~** SHOULD, **⊗** MUST NOT.

## Context & Problem Statement

Deft Directive is a *layered set of standards files plus deterministic `task` tooling*. It cleanly separates two things:

- **How the AI behaves** — markdown standards (`main.md`, `languages/`, `patterns/`, `skills/`, `templates/`, the `AGENTS.md` managed section). Language-agnostic; **not** the subject of this RFC.
[...]

PR #1865, opened 2026-06-21
fix(release-e2e): resolve task release:e2e deadlock via cross-thread worker wake (#1864)

### LINKED ISSUE #1864: bug(release-e2e): runPromiseSync deadlocks task release:e2e -- Atomics.wait blocks the main event loop that must resolve the worker Promise
## Symptom

`task release:e2e` hangs indefinitely. Observed during the v0.53.0 release rehearsal (2026-06-21): the harness provisioned the temp repo, cloned, `set-origin`, `push-mirror` (all OK), the worker even created the draft `v0.0.1` on the temp repo -- then the **main thread deadlocked** in `futex_wait_queue` for 20+ min, blowing past the 600s release / 300s rollback internal timeouts, and never reached the `try/finally` teardown (temp repo was left undeleted). Process state: `Sl`, `wchan=futex_wait_queue`, no active `gh`/`git`/`pnpm` subprocess.

## Root cause

`packages/core/src/release-e2e/entrypoint.ts::runPromiseSync` (lines ~137-163) blocks the **main thread** with `Atomics.wait(ia, 0, 0)` while waiting on the Promise returned by `runEntrypointWorker`. But that Promise only ever settles via **main-thread event-loop callbacks**:

- `worker.on("message", ...)`
- `worker.on("error", ...)`
- the `setTimeout(..., timeoutMs)` timeout timer

`Atomics.wait` blocks the main event loop, so none of those callbacks can run -- including the timeout timer -- and the `promise.then` handler that calls `Atomics.store`/`Atomics.notify` is a microtask that also cannot run. The notify therefore never fires, and `Atomics.wait` blocks forever. The timeout cannot save it because the timer lives on the same blocked event loop.

The worker thread itself runs `cmdRelease` fine (hence the draft got created); only the main thread's synchronous wait is wedged.

Call chain: `runRehearsal` (sync) -> `dispatchTaskRelease` (sync) -> `callReleaseEntrypoint` -> `callReleaseEntrypointWorkerBacked` -> `runPromiseSync(runEntrypointWorker(...))`.

Why tests didn't catch it: the unit tests exercise `callReleaseEntrypointTimed` (the **async** `await` path) and/or mock the worker, so they never hit the sync `runPromiseSync` deadlock that the real e2e dispatch uses.

## Fix direction

Resolve the worker result without depending on the blocked main event loop. Canonical Node pattern for synchronous worker handoff:

- Share a `SharedArrayBuffer` Int32 signal with the worker; have the **worker thread** call `Atomics.store(signal,0,1); Atomics.notify(signal,0)` after posting its result. The cross-thread notify wakes the main `Atomics.wait(signal, 0, 0, timeoutMs)` (and the SAB-level `timeoutMs` gives a working timeout).
- Pass results structured via a `MessageChannel` and read synchronously with `receiveMessageOnPort(port)` after the wait returns `"ok"`.
- On `Atomics.wait(...) === "timed-out"`, `worker.terminate()` and return `ENTRYPOINT_TIMEOUT_EXIT_CODE`.

(Alternative: make the whole `runRehearsal` -> dispatch chain `async` and use the already-correct `callReleaseEntrypointTimed`; larger blast radius up to the e2e main.)

Add a test that drives the **sync** `callReleaseEntrypoint` worker-backed path end-to-end with a fast real worker to prove (a) it returns the result without hanging and (b) the timeout path returns `124` instead of deadlocking.

## Provenance

Introduced by the TypeScript port of the release pipeline (#1729, Wave-4a of #1530). The Python oracle used real subprocesses with OS-level timeouts; the TS port substituted an in-process worker + `Atomics.wait` and introduced the event-loop-blocking deadlock.

Refs #1729 #1530

MScottAdams:
## Summary

`task release:e2e` deadlocked on every run. Discovered during the v0.53.0 release rehearsal: the harness provisioned the temp repo, cloned, mirrored, and the worker even created the draft `v0.0.1` — then the **main thread wedged in `futex_wait_queue` for 20+ minutes**, blew past its 600s/300s internal timeouts, and never reached `try/finally` teardown.

**Root cause:** `runPromiseSync` blocked the main thread with `Atomics.wait` while waiting on the Promise returned by `runEntrypointWorker`, whose only settle paths (`worker.on("message")`, `worker.on("error")`, the timeout `setTimeout`) are **main-thread event-loop callbacks**. A blocked event loop can never run them — so the `Atomics.notify` microtask never fired and even the timeout couldn't save it. The worker thread itself ran fine (hence the draft got created); only the main thread's synchronous wait was wedged.

Introduced by the TS port of the release pipeline (#1729); the Python oracle used real subprocesses with OS-level timeouts.

## Changes

- Replace `runPromiseSync` with `runEntrypointWorkerSync`: the worker posts its result to a transferred `MessagePort` and calls `Atomics.notify` **from the worker thread** (a cross-thread wake that does not depend on the main event loop). `Atomics.wait`'s own `timeoutMs` is the working timeout backstop; `receiveMessageOnPort` reads the result synchronously.
- Worker bootstrap notifies the shared signal + posts to the transferred port when present (sync path); unchanged `parentPort` behavior for the async path.
- New `entrypoint-sync.test.ts`: deterministic regression coverage for the wake, timeout, and worker-error paths.

## Test plan

- [x] `entrypoint-sync.test.ts` (3 tests) pass in ~280ms — the pre-fix code hung indefinitely
- [x] Full release-e2e suite green (76 tests, incl. existing coverage/main/parity)
- [x] `task check` fully green (8775 Python passed; TS lane 87.65% stmts / 85.08% branches)
- [x] **End-to-end validated**: real `task release:e2e` now completes in ~10s instead of hanging; it correctly fails at `[2/13] branch == master` only because the rehearsal was run from this fix branch, then cleans up and exits (the deadlock is gone)

## Notes

- The e2e harness's `gh repo delete` cleanup needs the `delete_repo` token scope; two private rehearsal temp repos were left behind during diagnosis and need manual cleanup (`deftai/deftai-release-test-20260621215813-9fc408`, `deftai/deftai-release-test-20260621224037-f12ec2`).

Closes #1864
Refs #1729 #1530


MScottAdams:
Addressed both review findings in 5407a0c (single batch):

**Issue 1 — missing `worker.on("error")` in the sync path:** added a documented no-op crash guard. Note the sync runner blocks the main thread in `Atomics.wait` and then returns synchronously, so an `'error'` event can only be delivered on a later tick *after* the function has returned — the listener can't act on it, but it prevents Node from re-throwing a late worker failure as an uncaught exception that crashes the process. (Avoided shipping the "capture and surface the error" variant since it would be dead code under the blocked-event-loop model.)

**Issue 2 — `postMessage` on a closed port after timeout termination:** wrapped the worker's `postMessage` in try/catch in the bootstrap. When the parent times out, closes `port1`, and calls `worker.terminate()`, a slow "hang" worker that later posts to the closed port no longer throws outside `runWorkerEntrypoint`'s catch (which previously fed the missing-handler issue above).

`task check` is green (8775 passed; TS lane incl. coverage thresholds; codebase map refreshed).

PR #2009, opened 2026-06-25
fix(doctor): npm canonical upgrade path and post-freeze signpost hardening

### LINKED ISSUE #1997: Post-freeze doctor: surface npm-migration signpost on a normal run + npm payload-staleness command + Python doctor legacy-layout parity
Child of #1993 (sub-problem 5). Split out from the live Cartograph consumer report.

Three doctor gaps post-freeze:
- **Throttle hides the signpost**: on a normal (non-`--full`) run the 24h clean-window throttle (`packages/core/src/doctor/main.ts:56-78`, `constants.ts` CLEAN_WINDOW_HOURS) can skip the legacy-layout signpost entirely. The frozen-final v0.56.0 canonical-vendored layout is a known terminal state and should be detectable/surfaced without a networked `--full` run.
- **Payload-staleness still recommends the Go bridge**: `CANONICAL_UPGRADE_COMMAND` (`doctor/constants.ts:55`) / `payload-staleness.ts:177` emit `deft-install --yes --upgrade ...`. Post-freeze the canonical path is npm — update the recommendation.
- **Python doctor lacks legacy-layout**: `scripts/doctor.py` has no `legacy-layout` check, so the vendored Python doctor never emits the npm signpost (only the TS doctor does). UPGRADING claims parity; close the gap or adjust the claim.

Refs #1993, #1912, #1411.

### LINKED ISSUE #2003: doctor payload-staleness command diverges from AGENTS.md / managed-section docs
## Summary

Two different "canonical upgrade command" strings are live across surfaces, and the AGENTS.md / managed-section narrative attributes the **wrong** one to `scripts/doctor.py`.

The managed-section template (and therefore every consumer-installed `AGENTS.md`) states:

> The canonical `scripts/doctor.py` ... when behind, emits the canonical upgrade command `npm i -g @deftai/directive@latest` (#1339 / #1409 / #1912).

But the doctor's payload-staleness path actually emits a **different** command.

## Evidence

Documentation side (claims `npm i -g @deftai/directive@latest`):
- `templates/agents-entry.md` (managed-section source) → propagates to every consumer `AGENTS.md`
- `AGENTS.md` (maintainer copy)
- `skills/deft-directive-sync/SKILL.md`, `README.md`

Source side (actually emits `deft-install --yes --upgrade --repo-root . --json`):
- `scripts/doctor.py:1640` — `recommended_command = "deft-install --yes --upgrade --repo-root . --json"` (docstring at :1502 confirms this is the payload-staleness emission, #1409)
- `packages/core/src/doctor/constants.ts:55` — `CANONICAL_UPGRADE_COMMAND = "deft-install --yes --upgrade --repo-root . --json"`
- Release notes for v0.56.1 / v0.56.2 also cite `deft-install --yes --upgrade --repo-root . --json`

## Why it matters

`npm i -g ...` and `deft-install --yes --upgrade ...` are **different operations**: the former refreshes the globally-installed CLI, the latter refreshes the *vendored* `.deft/core` payload in a project. The doctor's payload-staleness check is specifically about the vendored payload, so the source command (`deft-install --yes --upgrade`) is the correct one for that emission — which means the AGENTS.md narrative citing the npm command for that exact code path is inaccurate. A consumer who copy-pastes the npm command after a payload-staleness warning updates the wrong thing and the warning persists.

## Suggested resolution

Reconcile the managed-section narrative with the doctor source. Either:
1. Correct `templates/agents-entry.md` (and re-run `task agents:refresh`) so the doctor-attributed command matches `CANONICAL_UPGRADE_COMMAND` (`deft-install --yes --upgrade --repo-root . --json`), keeping `npm i -g @deftai/directive@latest` only where it correctly refers to refreshing the global CLI; or
2. If npm is intended to be the single canonical path post-freeze, update `doctor.py` + `constants.ts` to emit the npm command and adjust the release-notes template accordingly.

A small contract test asserting the doctor-emitted command string matches the string quoted in `templates/agents-entry.md` would prevent regression (cf. the existing `tests/content/test_agents_entry_contract.py` marker mechanism).

## Discovered

Found during a routine version-staleness investigation on a vendored install (cartograph, payload `v0.56.0`).

### LINKED ISSUE #2004: doctor payload-staleness silently green when ls-remote unavailable (false 'System check passed!')
## Summary

`scripts/doctor.py`'s payload-staleness check degrades to a silent, invisible `skip` whenever its `git ls-remote` probe can't resolve a remote SHA (no network, no origin, restricted/allowlist network, or empty output). Because `skip` findings do not affect the doctor's overall verdict, the doctor prints `✓ System check passed!` even when the installed payload is genuinely behind. In any network-restricted environment (CI, sandboxed agents, corporate egress filtering) the staleness guard is effectively blind and reports a false "all clear".

## Reproduction (observed live)

Vendored install on a project (cartograph), payload `v0.56.0`; latest published was `v0.56.2` (two patches behind, same day). `doctor.py --full` output:

```
ℹ Checking payload staleness from install manifest...
ℹ payload-staleness: skip -- ls-remote produced no sha
...
✓ System check passed!
```

The only reason the staleness was caught at all was an independent `npm view @deftai/directive version` (the npm registry was reachable in the same environment where `git ls-remote origin <ref>` to GitHub was not).

## Root cause

All network/resolution failure branches emit `add_finding(..., status="skip")` and `return`, which is non-fatal and does not downgrade the green summary:

- `scripts/doctor.py:1594-1597` — `git ls-remote failed (no network or no origin)`
- `scripts/doctor.py:1624-1626` — `ls-remote produced no sha` (the branch hit here)
- `scripts/doctor.py:1627-1630` — `could not probe remote (<Exc>)`

The docstring (`:1503-1504`) documents this as intentional best-effort behavior ("Skips gracefully ... when git / network / manifest unavailable (non-fatal)"). The problem is the *consequence*: a skip is indistinguishable from a confirmed-current result in the headline verdict, so "couldn't verify" reads as "verified fresh".

## Why it matters

- The payload-staleness guard is the canonical surface that tells consumers they're behind. When it silent-skips, consumers (and headless agents that key off doctor's exit/verdict) believe they're current when they're not.
- This is most likely to fire exactly where it's most dangerous: automated/sandboxed/CI contexts that can't reach GitHub directly.

## Suggested resolution

1. **Make the skip visible in the summary.** When payload-staleness can't verify, surface a distinct advisory line in the final summary (e.g. `⚠ payload currency UNVERIFIED — could not reach remote`) rather than folding into a clean `✓ System check passed!`. Keep it non-fatal, but not invisible.
2. **Add a network-independent fallback resolver.** Post-freeze the canonical distribution is npm, and the npm registry is often reachable when GitHub `ls-remote` is not (demonstrated above). The doctor could compare the manifest `ref`/`tag` against `npm view @deftai/directive version` as a fallback when `ls-remote` yields nothing — which would have caught the `v0.56.0` vs `v0.56.2` gap.
3. Optionally distinguish "skip: no manifest / inside repo" (genuinely not applicable) from "skip: could not verify currency" (applicable but unverifiable) so only the latter raises the advisory.

## Related

Surfaced alongside #2003 (doctor-emitted upgrade-command string diverging from AGENTS.md docs) during the same version-staleness investigation.

MScottAdams:
## Summary

- Align doctor payload-staleness remediation with the post-freeze npm path: both TS and Python doctors now recommend `npm i -g @deftai/directive@latest`, matching `agents-entry.md` / AGENTS.md (#2003).
- When `git ls-remote` cannot verify currency, fall back to `npm view @deftai/directive version` and surface a non-fatal **UNVERIFIED** advisory instead of a silent green pass (#2004).
- Surface legacy-layout and canonical-vendored npm-migration signposts on throttled (non-`--full`) doctor runs; add Python doctor parity for legacy-layout detection (#1997).

Closes #1997
Closes #2003
Closes #2004

## Test plan

- [x] `TMPDIR=$(mktemp -d) task check`
- [x] New contract test: `tests/cli/test_doctor_upgrade_command_contract.py`
- [x] TS unit tests under `packages/core/src/doctor/*.test.ts`
- [x] Throttle signpost integration: `tests/cli/test_doctor_throttle.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **Vendored `run update` now points at npm instead of dead-ending** — the compatibility `run update` stub no longer tells operators to manually replace the deft directory; it signposts `npm i -g @deftai/directive@latest` and `npx @deftai/directive update` so future deposits steer consumers to the post-freeze upgrade path. Closes #1998.
+- **Doctor now recommends npm upgrades and surfaces layout advisories on throttled runs** — payload-staleness emits `npm i -g @deftai/directive@latest` (matching AGENTS.md), falls back to the npm registry when GitHub `ls-remote` is unreachable, and warns when payload currency is UNVERIFIED instead of silently passing; legacy-layout and canonical-vendored npm-migration signposts run on normal throttled doctor runs and in the Python doctor. Closes #1997, #2003, #2004.
 
 ### Removed
 
--- a/packages/cli/src/triage-cli.test.ts
+++ b/packages/cli/src/triage-cli.test.ts
@@ -6,7 +6,6 @@ import {
   buildFixtureRepo,
   diffCase,
   normalizeOutput,
-  PARITY_CASES,
   renderReport,
   resolveDeftRoot,
   runParity,
--- a/packages/core/src/doctor/branches.test.ts
+++ b/packages/core/src/doctor/branches.test.ts
@@ -106,7 +106,7 @@ describe("payload staleness", () => {
         stdout: "cafebabe refs/heads/main\n",
       }),
     });
-    expect(lines.join("")).toContain("deft-install --yes --upgrade");
+    expect(lines.join("")).toContain("npm i -g @deftai/directive@latest");
   });
 });
 
--- a/packages/core/src/doctor/checks.ts
+++ b/packages/core/src/doctor/checks.ts
@@ -4,13 +4,24 @@ import {
   type LegacyDetectSeams,
   legacyLayoutSignpostLine,
 } from "../init-deposit/legacy-detect.js";
+import {
+  detectCanonicalVendoredManifest,
+  isNpmManaged,
+  NPM_MANAGED_SENTINEL_KEY,
+  NPM_MANAGED_SENTINEL_VALUE,
+} from "../init-deposit/migrate.js";
 import { findSkillPathsInText } from "../text/redos-safe.js";
-import { GO_BRIDGE_RELEASES_URL, UPGRADING_DOC_URL } from "./constants.js";
+import {
+  CANONICAL_UPGRADE_COMMAND,
+  GO_BRIDGE_RELEASES_URL,
+  UPGRADING_DOC_URL,
+} from "./constants.js";
 import {
   isDeprecationRedirectStub,
   locateManifest,
   manifestCandidatePaths,
   manifestTagToVersion,
+  parseInstallManifest,
   parseInstallRootFromAgentsMd,
   parseManifest,
 } from "./manifest.js";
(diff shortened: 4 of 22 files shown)


MScottAdams:
@greptileai please re-review latest HEAD (e09ca6b) — addressed P1 JSON throttle signpost_findings and P2 branch-ref semver false-stale.

Issue #2022, opened 2026-06-26
Python purge readiness — execution tracker (supersedes #2001)


MScottAdams:
> **Purpose.** The pre-#1860 integration gate, rewritten clean. All decisions are
> settled; this issue is **pure execution**, ordered by dependency. When the phases
> below are green **and Phase 5 operator sign-off is recorded**, #1860 (big-bang Python
> delete) may be authorized. **Phase 5 is a hard stop** — no #1860 work until the
> conversion-hole review pass completes.
>
> **Decision record (immutable):** preserved on #2001 —
> [gate re-decision](https://github.com/deftai/directive/issues/1933#issuecomment-4811038059) ·
> [§1/§2 rescope](https://github.com/deftai/directive/issues/2001#issuecomment-4811038164) ·
> [§3 → #2013](https://github.com/deftai/directive/issues/2001#issuecomment-4810721079) ·
> [§6 bootstrap](https://github.com/deftai/directive/issues/2001#issuecomment-4810485514) ·
> [Python ledger](https://github.com/deftai/directive/issues/2001#issuecomment-4810720223)

## Settled decisions (baked in — not re-litigated here)

- **Gate:** deprecate-by-disuse (#1933 Option 1). The frozen Go `gate` subcommand stays
  in place, **unused**; the frozen binary is **not** reopened to remove it. `deft doctor`
  (node present) + agent `command not found` → `UPGRADING.md` (cold start) are the
  health/probe surfaces.
- **`migrate:vbrief` (`migrate_vbrief.py`):** out of scope here. The TS-port-vs-cutoff
  decision is owned by the **#2013** umbrella (v0.20 spec-authority model) — its output
  contract is spec-authority dependent. It is the one Bucket-A holdout: **not** in this
  issue's Phase 1, and **not** deletable by #1860 until #2013 resolves it.
- **Bootstrap:** implement the `directive bootstrap` thin launcher (not skill-only).

## Execution order (load-bearing)

**Phase 0 → Phase 1 convert → Phase 2 rewire → Phase 3 unbundle → Phase 5 audit (HARD STOP)
→ operator authorizes #1860 delete.** Phase 4 (bootstrap) runs in parallel with Phases 1–3.

The ordering is a hard dependency, not a preference: the bundled `scripts/` tree is still
load-bearing on the consumer path — `deft <verb>` shells into it via `loadPythonScriptHandler`
(6 routes in `packages/cli/src/dispatch.ts`) **and** the consumer Taskfile calls
[...]

PR #2031, opened 2026-06-26
feat: convert the policy-set write path to native TypeScript (#2022 Phase 1)

MScottAdams:
## Summary
- Converts the `policy-set` write path from a `policy_set.py` shell-out to a native TypeScript handler in `packages/cli/src/dispatch.ts` (#2022 Phase 1 — Python purge). The route no longer uses `loadPythonScriptHandler`; the now-dead `execFileSync` path is removed.
- Ports every `policy_set.py` subcommand: `wip-cap`, `subagent-backend`, `subagent-backends`, and `enforce-branches` / `allow-direct-commits`. Typed-field writes update `PROJECT-DEFINITION.vbrief.json` in place and append the `meta/policy-changes.log` audit row.
- Output, JSON serialization, audit trail, and exit codes are **byte-for-byte identical to the Python oracle** (verified against `scripts/policy_set.py` for every subcommand, including the no-confirm refusal paths).
- Builds on the already-merged pack-migrate dispatch.ts conversion (#2030); pack-migrate routes are untouched.

## Acceptance coverage (story py-purge-policy-set-ts)
- **a1** — `policy-set` routes to a native TS handler with no `loadPythonScriptHandler` call. ✓
- **a2** — a typed field write updates the field and appends an audit row to `meta/policy-changes.log`. ✓
- **a3** — `dispatch.test.ts` asserts the write path updates the field and records the audit row. ✓ (57 dispatch tests; new `native policy-set handler (#2022)` describe block)

## Test plan
- [x] `pnpm exec vitest run packages/cli/src/dispatch.test.ts` — 57 passed
- [x] Full TS suite `pnpm exec vitest run` — 7337 passed (513 files)
- [x] Python suite via `task check` — 8373 passed
- [x] Byte-exact Python-oracle parity diff for every subcommand (stdout + PROJECT-DEFINITION + audit log)
- [x] `biome check` clean on edited files

> Note: `task check` reports a pre-existing `vbrief:validate` drift on the `#2022` tracker's child references (setupghx / rewire-consumer / npm-deposit / bootstrap-launcher). That drift is orchestrator-owned and outside this story's file scope; it is unrelated to this change.

Refs #2022.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - **`deft pack-migrate` now runs without Python installed** — the five pack-migrate verbs (skills, rules, strategies, patterns, swarm-spec) are now served by native TypeScript handlers instead of shelling out to bundled Python scripts, so pack rendering works on a machine that has no Python. Output is byte-for-byte identical to the previous Python contract. Refs #2022.
+- **Editing a typed policy field no longer needs Python** — the `policy-set` write path (`task policy:wip-cap`, `task policy:enforce-branches` / `allow-direct-commits`, and the `subagent-backend(s)` surfaces) is now a native TypeScript handler instead of shelling into a bundled Python script, so changing a policy and recording its `meta/policy-changes.log` audit row works on a machine with no Python. Output and the audit trail are byte-for-byte identical to the previous Python contract. Refs #2022.
 
 ### Fixed
 
--- a/packages/cli/src/dispatch.test.ts
+++ b/packages/cli/src/dispatch.test.ts
@@ -711,9 +711,12 @@ describe("native policy-set handler (#2022)", () => {
   }
 
   function readPolicyBlock(): Record<string, unknown> {
-    const parsed = JSON.parse(readFileSync(projectDefPath(), "utf8")) as Record<string, unknown>;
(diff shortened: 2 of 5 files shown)


MScottAdams:
Pushed 9bbe13ab addressing the review findings:

**SLizard P1 (blocking) — `JSON.parse` null guards (3):** Added `typeof !== "object" || === null` guards before the `plan`/`policy` property chain in `loadProjectDefinitionForWrite` and before property access on the two test-side parses. A top-level `null`/array/scalar now yields a clean config error instead of a `TypeError`.

**Greptile P2 — whitespace `--set` value:** `wip-cap --set " 7 "` now trims before the integer check, restoring Python `int()` parity (added a parity test).

**Greptile P2 — `--project-root ~` tilde:** Added `expandUser()` to mirror `Path(...).expanduser()` so `~`/`~/...` resolves under `$HOME` as it did in Python.

**Greptile P1 — "Cached Handler Keeps Stale Output":** This is a pre-existing dispatch-wide trait, not introduced here: the handler cache binds `io` from the first `dispatch()` call for *every* core verb (including the already-merged `pack-migrate` routes that use the identical `(argv) => runX(argv, io)` closure). Production uses a single `defaultIo`, so it's inert there; only repeated programmatic calls with distinct `DispatchIo` objects are affected. Fixing it means threading `io` at invoke time across the whole cache, which is out of this story's `file_scope`. The new tests already account for it via `resetHandlerCacheForTests()` between calls.

Issue #2264, opened 2026-07-03
Resolution spine: fact-set classifier + plan() contract + global-first ladder + pin (keystone)


MScottAdams:
Part of epic #2203. Keystone child — everything else depends on this.

## Summary

Build the shared resolution spine that all three consumption contexts (matched shell, mismatched sandbox, headless product) stand on: the fact-set classifier, the `plan()` contract, the global-first engine-resolution ladder, the `package.json` version pin + `VERSION` reconciliation, and the three-band skew policy.

Absorbs #2124 (global-first ladder + committed pin) and the version half of #2199.

## Classifier + plan() (Decision 5)

- `classify()` returns an **orthogonal fact-set**, never a single enum: `hasGit, hasAppCode, hasDeftCore, deftCorePayloadVersion, hasManagedSection, managedSectionSha, hasVbrief, hasXbrief, preCutoverArtifacts, engineReachable, engineVersion, pinVersion`.
- `plan()` applies an explicit, ordered **precedence table** over the fact-set and emits exactly one recommended action + rationale.
- `plan()` output is a **versioned public JSON schema** — `{ schemaVersion, mode, files: [{ path, content, encoding }], nextAction: { command, rootCause, remediation }, warnings[] }`. `doctor` prints it; headless (#child E) serializes it; products consume it. Single source of truth (closes the #537 split-source drift risk).

## Global-first engine-resolution ladder (from #2124)

```text
1. global `deft` reachable AND >= pin?                    -> use it   (matched / new machine / post-upgrade)
2. local .deft/.cli/<platform> reachable, intact, >= pin? -> use it   (warm mismatched sandbox)
3. must install:
   a. registry up + global prefix writable                -> npm i -g @deftai/directive@<pin>
   b. registry up + global prefix NOT writable (sandbox)  -> npm install --prefix .deft/.cli/<platform>
   c. registry down                                        -> pre-staged tarball / vendored payload; else hard-fail "stage this"
```

[...]

Issue #2203, opened 2026-07-02
Epic: unified Directive adoption & consumption model (init / update / doctor)


polfer:
## Summary

One epic owning the first-run-through-upgrade lifecycle for a Directive **consumer install**, across every way an installed project is consumed: a human in a matched shell, an agent in a mismatched sandbox (execution-env != install-env), and a hosted product with no shell at all. Re-baselined from the original first-run UX epic to also own the hybrid-install hardening (formerly #2124/#2123) and the headless product contract (formerly #2253), because all three reshape the same `init`/`update`/`doctor` verbs.

Target end state:

```text
directive init      Start or adopt this project (any directory, any environment)
directive update    Refresh an existing install (self-healing, version-reconciling)
directive doctor    Diagnose only; emit one next command with its reason
```

## Problem

First-run and upgrade behavior was specified across three disconnected places that reshape the *same verbs* from different sides — UX/routing (the original #2203 children), engine/content resolution in mismatched sandboxes (#2124/#2123), and no-shell product installs (#2253, which blocks deftai/deftvisage#1036). Because they lived apart: docs over-promise, `update`/`doctor` emit competing guidance, three independent version-pins drift, and the cold-start recovery pointer is stored *inside* the very payload (`.deft/core/`) that is absent when recovery is needed.

## Operating modes (one resolver; in scope = 1 & 2)

1. **In-repo vendored** (today) — `.deft/core/` committed. Stays supported.
2. **In-repo hybrid** — global/npm engine + gitignored content; self-heals across matched/mismatched execution envs. **Primary target of this epic.**
3. **External workspace** (#1583) + **content-from-package** (#1944) — **out of scope**, remain owned by #1669. This epic's resolver/pin are designed not to preclude them.

## The unifying model

**1. `init` is `classify -> plan -> execute`, with plan separable from execution.**
- `classify()` returns an **orthogonal fact-set**, never a single enum.
- `plan()` applies an explicit, testable **precedence table** over the fact-set and emits exactly one recommended action + rationale.
- `plan()` output is a **versioned public JSON schema** — the same contract `doctor` prints and headless serializes (single source of truth; closes the #537 split-source drift risk).
- `execute` is the only environment-specific stage: full side effects (matched), engine-bootstrap-then-content (mismatched), or serialize-to-manifest with no side effects (product).

[...]

PR #2296, opened 2026-07-05
feat: first-class pnpm support for the install/upgrade path (#2197)

### LINKED ISSUE #2197: Upgrade/install path assumes npm; add first-class pnpm support
## Problem

The documented distribution/upgrade path for Deft (`directive`) now assumes **npm** as the only global package manager. The release notes and docs advertise the canonical path as:

```
npm i -g @deftai/directive@latest
deft update && deft migrate && deft doctor
```

There is **no pnpm equivalent** documented or supported. Users on a pnpm-managed setup have no first-class install/upgrade path.

## Why it matters

- `npm i -g` global installs are not compatible with a pnpm-managed environment. Mixing an npm global into a pnpm setup causes PATH/shim and store inconsistencies, so the "canonical" upgrade path is effectively unusable for pnpm users.
- pnpm is a mainstream, widely-adopted package manager. Its users are currently **blocked** from the canonical upgrade path with no supported alternative.
- Forcing npm-only guidance pushes pnpm users into ad-hoc workarounds, which undermines the reliability goals of a canonical upgrade flow.

## Observed context

- Latest release `v0.66.2` ships **zero release assets** — the project moved off standalone installer binaries in favor of npm distribution.
- The package `@deftai/directive@0.66.2` is confirmed published on the npm registry.
- Because standalone installers were dropped, npm is currently the *only* signposted way to get/upgrade the CLI, which makes the missing pnpm story a hard gap rather than a nice-to-have.

## Request

Please document and support a first-class **pnpm** install + upgrade method. Concretely:

1. Add a pnpm install/upgrade path to the release notes and UPGRADING docs, e.g.:
   ```
   pnpm add -g @deftai/directive@latest
   ```
   and/or a `pnpm dlx @deftai/directive ...` invocation and/or a project-local (non-global) equivalent for pnpm-managed repos.
2. Verify that `deft update && deft migrate && deft doctor` all work correctly when the CLI is installed via pnpm (global and/or project-local), and fix any assumptions that are hard-coded to npm's global layout.
3. Document any caveats for pnpm's global bin directory / PATH setup so pnpm users have a clean, supported flow.

This would restore parity so pnpm users are not locked out of the canonical upgrade path.

MScottAdams:
## Summary

Gives pnpm-managed consumers a first-class install/upgrade path. The documented distribution flow previously hard-coded npm (`npm i -g @deftai/directive`), locking pnpm users out of the canonical path. This is a **pure consumption-side** change — pnpm resolves the same `@deftai/directive` package from the same npm registry, so there is **no additional registry, no republish**.

Scope + locked decisions were recorded on the issue first: https://github.com/deftai/directive/issues/2197#issuecomment-4884330595

### What changed

- **New `packages/core/src/resolution/package-manager.ts`** — a pure `detectPackageManager()` (precedence: `DEFT_PACKAGE_MANAGER` env → `packageManager`/Corepack field → `pnpm-lock.yaml` → `npm_config_user_agent` → npm default) plus command renderers for the four forms (global install, project-local install, ephemeral run, upgrade one-liner). No renderer ever emits a `--registry` flag.
- **`plan()` + install rungs** thread an optional `packageManager` (default `npm`, so all existing behaviour/tests are unchanged) and render pnpm forms (`pnpm add -g`, `pnpm dlx`) when pnpm is active.
- **Locked non-goal:** the internal `.deft/.cli/<platform>` sandbox rung stays on `npm install --prefix` regardless of the detected PM — its `node_modules/.bin` layout is npm-shaped and validated by `integrity.ts`, and it is gitignored / package-manager-invisible. Enforced by a boundary test.
- **doctor:** `PNPM_UPGRADE_COMMAND` + `upgradeCommandFor(pm)`; `refresh.ts` self-heal remediation and the CLI `printHelp` surface the pnpm alternative alongside npm.
- **Docs:** README (cold-start block, Getting Started, upgrade note), `content/UPGRADING.md`, and `content/docs/BROWNFIELD.md` document the pnpm install+upgrade path and the `PNPM_HOME` / `pnpm setup` PATH caveat.

### Deferred (explicit non-goal)

pnpm-shaped vendoring of the internal `.deft/.cli/<platform>` sandbox. It stays on npm; file a follow-up only if a real pnpm-only-no-npm environment needs it.

## Test plan

- [x] New `resolution/package-manager.test.ts` — detection ladder + render matrix + no-`--registry` assertion
- [x] `plan.test.ts` — pnpm install-global / ephemeral rendering + sandbox-stays-npm boundary; npm default unchanged
- [x] `doctor/constants.test.ts` — pnpm upgrade one-liner + `upgradeCommandFor`
- [x] `readme_brownfield.test.ts` — `#2197` content-contract block for the pnpm docs
- [x] `task ts:check-lane` green
- [x] `task check:framework-source` green

Closes #2197. Refs #2264, #2203.


MScottAdams:
Addressed the SLizard review (HEAD `66743f0fac07ebfafbeba5b9de3257b5a2445d19`):

**Fixed**
- **P2 (real) — `main.ts` Corepack detection:** the doctor's package-manager detection now also reads the project `package.json` `packageManager` field, so a Corepack-managed pnpm project with no `pnpm-lock.yaml` yet gets the pnpm upgrade hint. (We intentionally still exclude the ambient `npm_config_user_agent`, which reflects the spawning shell, not the consumer's project.)
- **P1 — `PNPM_UPGRADE_COMMAND` dead export:** removed. It was only referenced from tests; `upgradeCommandFor("pnpm")` renders the same string and is wired into `payload-staleness.ts`. The test now asserts against `upgradeCommandFor`.

**False positives (no change, with evidence)**
- **P1 `constants.ts:95` — `upgradeCommandFor` "only referenced from tests":** it is imported and used in production at `packages/core/src/doctor/payload-staleness.ts` (`const upgrade = upgradeCommandFor("pnpm")`), inside the same diff. It is genuinely wired.
- **P1 `payload-staleness.ts` — "called with 5 args but accepts at most 4":** `resolveUpgradeCommand` is defined with **4** parameters (`projectRoot, manifest, isFile, pm`) and called with **4** arguments. `task ts:check-lane` (tsc) passes, which it would not if the arity were wrong.
- **Advisory `renderGlobalInstall` "global fan-out":** the function only formats a command string; there is no fan-out.

`task check:framework-source` is green.

Issue #2376, opened 2026-07-06
feat(value-feedback): enrich attribution events + trusted-org auto-enable of local collection (Refs #1709)


MScottAdams:
## Summary
Make the per-machine value-feedback ledger (#1709) good enough to aggregate later, and auto-enable LOCAL collection for company-owned (deftai) repos with zero per-repo or per-machine setup. Strictly local + no network here; the collector/push/aggregation is a deferred follow-up.

## Motivation
#1709 events today (a) aren't tagged with which repo or directive version produced them, and (b) sit behind a manual per-repo opt-in that no deftai repo has set. We want every deftai-owned repo collecting good local data automatically -- on a fresh clone, on any machine -- so we can aggregate a fleet later.

## Part A -- Event enrichment (no policy/consent implications)
Stamp each attribution record at emit time with:
- `repo` -- normalized `owner/name` via existing `inferRepoFromGit` / `parseGitHubRemoteRepo`
- `directive_version` -- `readCorePackageVersion()` at emit time (attribute value to the version that earned it)
- `install_id` -- stable random uuid persisted once in `.deft-cache/` (per machine/clone; not tied to user identity)
- `schema_version` -- for future record-shape migration

## Part B -- Trusted-org auto-enable (built-in, zero-config)
Add a resolution layer above the typed policy, evaluated at session start:
- Read the origin org via `inferRepoFromGit` (owner of `owner/name`).
- Compare it against a **built-in framework constant** `DEFAULT_AUTOENABLE_ORGS = ["deftai"]` (shipped in the framework source -- this IS how deftai "configures" it: once, in-code, not per-repo and not per-machine).
- On match, with no explicit typed `valueFeedback` flag present -> resolve `emitEvents` + `sessionLine` ON, `source="org-auto"`, **network/upstream OFF**.
- Precedence: an explicit typed `valueFeedback.enabled=false` always wins (opt-out honored).
- Optional override (generalization, NOT deftai's mechanism): env `DEFT_VALUE_AUTOENABLE_ORGS` (comma-list) extends/overrides the built-in constant, so other orgs / controlled CI envs can opt their own org in without a code change.
- Fail-safe: no origin remote, or org not matched -> stays OFF (correct default; e.g. detached CI checkouts).
- Emit the capability-cost disclosure once as a budgeted session line when auto-enabled (transparency without a prompt).

### Company policy decision (resolved)
[...]

Issue #1709, opened 2026-06-18
Epic: Directive value attribution & adoption loop (RFC)


MScottAdams:
## Problem

Consumers can't see what directive does for them. Its value is invisible when **present** (silent gates, prevented mistakes), invisible when **absent** (going off-flow feels free), and invisible when **under-used** (a consumer can "use directive" while never touching planning, cost, or decompose — leaving most of the value on the table). And when directive genuinely falls short, that signal dies in the consumer's session instead of reaching the maintainers.

## Goal

Make directive's value legible along four axes — **without** lying, nagging, or spending the consumer's tokens unless they opt in:

1. **Value delivered** — show, with evidence, what directive caught.
2. **Boundary awareness** — make going off-flow visible.
3. **Adoption awareness** — surface directive capabilities relevant to the work that the consumer isn't using (planning, cost, decompose, pre-PR, debug, glossary, lessons/packs).
4. **Gap escalation** — when directive fails or is missing something, help the consumer file an issue upstream on `deftai/directive`.

## Principles (binding)

- **Attributed-only** — every value claim points to a concrete logged event ("encoding gate caught 2 corruptions"), never "your quality improved." Silence when there's nothing to attribute.
- **Opt-in before it ever fires** — default OFF. Zero behavior, zero tokens, zero log writes until the consumer explicitly enables it. Token cost disclosed at enable time. This is the load-bearing constraint: it gates everything downstream.
- **Budgeted, never naggy** — one line max per session (respects the existing nudge budget, #1419). Detail is pull-based (`task deft:value:show`), not pushed.
- **Confirmation-gated escalation** — the agent drafts + dedups; the user approves before any upstream issue is filed (mirrors `deft-directive-article-review`'s "MUST NOT create without confirmation").
- **Consumer-only** — upstream filing targets `deftai/directive`, never the consumer's own repo, and no-ops inside the maintainer repo (detection exists: `_running_inside_deft_repo()`, `is_framework_source_context()`).

## The opt-in / token model

A single typed policy flag gates the entire feature, following the existing typed-policy pattern (`plan.policy.wipCap`, `allowDirectCommitsToMaster`):

```jsonc
plan.policy.valueFeedback: {
  "enabled": false,              // master switch — DEFAULT OFF, nothing fires
  "emitEvents": true,            // attribution ledger logging (cheap, local)
  "sessionLine": true,           // welcome one-liner readback
  "upstreamPrompt": true,        // session-end gap-escalation offer
[...]

Issue #2189, opened 2026-07-02
pr:wait-mergeable-and-merge merges through SLizard blocking findings (only gates on Greptile)


MScottAdams:
## Summary

While driving PR #2187 (issue #2182) to merge-ready via the canonical cascade surface (`task pr:wait-mergeable-and-merge -- 2187 --repo deftai/directive`), the automation merged the PR while a SLizard review was still outstanding a **blocking P1 finding**:

```
Decision: request_changes
Merge impact: blocking
Findings: 2 (P0: 0, P1: 1, P2: 0, P3: 0)
Important files: packages/core/src/doctor/network-gate.test.ts (P1, 2 finding(s), confidence 1.00)
```

`task pr:wait-mergeable-and-merge` (`scripts/pr_wait_mergeable.py`, composing `task pr:merge-ready` / `scripts/pr_merge_readiness.py`) only parses the **Greptile** rolling-summary comment for its merge-exit condition (confidence, P0/P1, HEAD-SHA freshness, errored sentinel). It has no equivalent check for the **SLizard** check-run's structured verdict (`Decision`, `Merge impact`, severity counts), even though SLizard also posts a check-run with an explicit `request_changes` / `blocking` decision.

Because the cascade uses `gh pr merge --squash --delete-branch --admin`, the `--admin` flag bypasses GitHub branch-protection required-status-check enforcement too, so nothing else caught this — the PR merged with the P1 finding unresolved.

## Impact

Any PR reviewed by both Greptile and SLizard can merge through the cascade automation (or the "merge at will" operator-approved path) with an unresolved SLizard blocking finding, as long as Greptile's own verdict is clean. This defeats the purpose of running a second reviewer.

## Suggested fix

Extend `scripts/pr_merge_readiness.py` (or add a sibling check invoked by `pr_wait_mergeable.py`) to also parse the SLizard check-run `output.summary` (or annotations) for `Decision: request_changes` / `Merge impact: blocking`, and fail merge-readiness (non-zero exit) until it's resolved or an explicit override is supplied -- mirroring the existing Greptile-body verdict gate.

## Reproduction

- PR: https://github.com/deftai/directive/pull/2187
- SLizard check-run (still failing/blocking at merge time): commit `e94ef7b0fc11a568c57d2dda4e8350e1b42b0e19`, check-run conclusion `failure`, decision `request_changes`.
[...]

PR #2377, opened 2026-07-06
feat(value-feedback): trusted-org auto-enable of local collection + event enrichment (Refs #1709)

MScottAdams:
## Summary

Implements #2376 — makes the per-machine value-feedback ledger (#1709) good enough to aggregate later, and auto-enables **local** collection for company-owned (`deftai`) repos with zero per-repo/per-machine setup. Strictly local + no network; the central collector/push/aggregation stays a deferred follow-up.

**Part A — event enrichment.** Every attribution record is stamped at emit time with `repo` (normalized `owner/name` from the git origin), `directive_version` (`readCorePackageVersion`), a stable per-checkout `install_id` (persisted once in gitignored `.deft-cache/install-id`), and `schema_version`. Enrichment is best-effort and never interrupts the emit path.

**Part B — trusted-org auto-enable.** A new resolution layer above the typed policy: when the typed `valueFeedback` block is **absent** and the origin org is in the trusted set (built-in `deftai`, extendable via `DEFT_VALUE_AUTOENABLE_ORGS`), LOCAL `emitEvents` + `sessionLine` resolve ON with `source=org-auto` and network/upstream OFF. Precedence: an explicit typed block (including `enabled: false`) always wins; a non-matching org or no origin remote stays OFF (fail-safe). Org detection is memoized to stay off the hot emit path.

**Company policy (resolved).** For `deftai`-owned repos, org membership *is* the consent for local, no-egress collection on company-owned resources — network/upstream escalation still requires explicit operator confirmation. The maintainer-repo split is preserved: in-org emit auto-enables, but session readback stays gated behind `DEFT_VALUE_SELF_DOGFOOD=1`.

**Docs (contract-locked).** The #1709 rules are reworded in lockstep across `AGENTS.md`, `content/templates/agents-entry.md`, and `content/templates/agent-prompt-preamble.md`; the operator-confirmation prohibition is scoped to network/upstream surfaces; new `#2376` markers were added to `agents_entry_contract.test.ts`; the AGENTS.md region budget was raised for the added rules.

## Out of scope (deferred)

Central collector service, network push, high-water-mark cursor, and cross-repo aggregation.

## Test plan

- [x] `task check` green (validate + lint + full test + coverage 88.79% stmts / 85.05% branches)
- [x] New source files ship with tests (`verify:forward-coverage`)
- [x] `agents_entry_contract` markers present in AGENTS.md + template
- [x] Temp-dir tests (no git remote) confirm org-auto stays OFF — existing #1709 expectations preserved
- [x] Explicit `enabled:false` overrides org-auto; non-trusted org / no remote → OFF

Refs #1709

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,12 +74,14 @@ Same `!` / `⊗` rules as the managed `## Deterministic questions runtime obliga
 
 ## Value feedback and attribution (#1709)
 
-- ! `plan.policy.valueFeedback.enabled` defaults OFF -- while false, every downstream path (emit-only ledger, budgeted session readback, upstream gap escalation) short-circuits with zero token spend. Opt-in ONLY via `task policy:enable-value-feedback -- --confirm` after the capability-cost disclosure prints. Inspect with `task policy:show --field=valueFeedback`.
+- ! `plan.policy.valueFeedback.enabled` defaults OFF for non-org repos -- while off, every downstream path (emit-only ledger, budgeted session readback, upstream gap escalation) short-circuits with zero token spend. Opt-in for any repo via `task policy:enable-value-feedback -- --confirm` after the capability-cost disclosure prints. Inspect with `task policy:show --field=valueFeedback`.
+- ! Trusted-org local auto-enable (#2376) -- for a repo whose GitHub origin belongs to a company-owned org (built-in default `deftai`; extend via the `DEFT_VALUE_AUTOENABLE_ORGS` env override), LOCAL emit + session readback resolve ON with `source=org-auto` and network/upstream OFF, with NO per-repo or per-machine confirmation: org membership IS the consent for local, no-egress collection on company-owned resources. An explicit typed `valueFeedback` block always wins (including `enabled: false`); a non-matching org or no origin remote stays OFF (fail-safe).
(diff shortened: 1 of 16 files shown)


MScottAdams:
### SLizard override rationale (#2189)

Greptile (the required review of record) is CLEAN at 5/5 on HEAD `33ed3434` (P0=0 P1=0 P2=0) after the review-fix commit. The remaining SLizard P1s are all verified false positives, so merging with `--skip-slizard`:

- **`resolveValueFeedback` "only referenced from tests" + "arity 3"** — false. It has 7 production call sites (`attribution-ledger.ts`, `adoption-emit.ts`, `friction-emit.ts`, `readback.ts` ×2, `feedback-file.ts`, `value-feedback.ts`), all passing 1 arg. SLizard only scans the PR-diff files, so it can't see the unchanged production callers, and it mis-reads the backward-compatible optional 2nd param as an arity overflow.
- **`inspectValueFeedbackField` "arity 4"** — false. It is invoked only as `inspect(data, projectRoot)` (2 args) via the heterogeneous `REGISTERED_POLICIES.map(...)` callback array, which SLizard's static graph cannot resolve to the concrete function.
- **`clearOriginOrgCache` "only referenced from tests"** — true but intentional: it is a test/hygiene cache-clear hook, mirroring the existing `clearRegistryCache` pattern.

Issue #2595, opened 2026-07-16
v0.78 update leaves migrated xbrief schema and version derivative stale


dbcall2:
Consumer upgrade from v0.72 to v0.78 reproduced two post-migration drifts: (1) xbrief/.deft-version remains 0.72.0 even after directive update reports current 0.78.0; (2) xbrief/schemas/vbrief-core.schema.json retains the vBRIEFInfo v0.6 root schema after migrate:xbrief, and a subsequent directive update no-ops instead of refreshing it. directive doctor --full reports manifest-agreement failure and stale-xbrief-schema-deposit. Also, a cache-only xbrief/.triage-cache directory caused the first migrate:xbrief --force invocation to report converged without migrating project artifacts. Refs #2034/#2110/#2149.

dbcall2:
Additional v0.78.0 deposit defect: .deft/core/Taskfile.yml contains trailing whitespace on generated ENGINE_CMD lines 364 and 463, causing git diff --check to fail in consumer pre-PR validation.

PR #2610, opened 2026-07-16
fix(installer): hybrid update/setup deposits root git hooks (#2530)

### LINKED ISSUE #2530: fix(installer): v0.75 hybrid setup does not deposit root git hooks
## Summary

Directive v0.75.0 hybrid deposits can leave consumer projects without a project-root `.githooks/` directory. The documented setup path then configures Git to use that missing directory, and the canonical hook verifier fails.

## Version and deposit mode

- `@deftai/directive` / `@deftai/directive-core` / `@deftai/directive-content`: `0.75.0`
- Consumer deposit: canonical `.deft/core` with `managed_by: 'npm'` (hybrid)
- Platform reproduced on: macOS with zsh

## Safe reproduction

Starting from a v0.75.0 hybrid consumer deposit whose project root does not contain `.githooks/`:

```sh
git config --unset core.hooksPath || true
task deft:setup
deft verify:hooks-installed
deft update
deft verify:hooks-installed
```

## Expected behavior

The documented setup or repair path should deposit the current hook templates at project-root `.githooks/pre-commit`, `.githooks/pre-push`, and `.githooks/_deft-run.sh`, set executable modes where required, configure `core.hooksPath=.githooks`, and make `deft verify:hooks-installed` pass.

If the payload is already current, `deft update` should still repair missing installer-managed root hooks rather than treating version currency as proof that the consumer scaffold is complete.

## Actual behavior

1. `deft setup` is not a registered v0.75.0 CLI verb, so consumer documentation routes through `task deft:setup`.
2. `task deft:setup` only runs `git config core.hooksPath .githooks`; it does not create or refresh project-root `.githooks/`.
3. The resulting configuration points Git at a nonexistent directory.
4. `deft verify:hooks-installed` fails with the `#1463 false-green` diagnostic.
5. `deft update` reports/skips the already-current payload and does not reconstruct the missing project-root hooks, so the documented repair path cannot self-heal this state.

The nonfunctional repository-local `core.hooksPath` setting was rolled back after reproduction; the consumer is not being left configured to a missing hook directory.

## Root-cause evidence in the shipped v0.75.0 payload

- `@deftai/directive-content/Taskfile.yml` defines `setup` as configuration plus the optional ghx check. Its hook step only reads/writes `core.hooksPath`; it has no root-hook copy/deposit operation.
- `@deftai/directive-core/dist/init-deposit/scaffold.js` contains `writeConsumerGitHooks()`, which correctly copies `.deft/core/.githooks/{pre-commit,pre-push,_deft-run.sh}` to project-root `.githooks/` and then configures Git.
- The available consumer setup task does not invoke that deposit primitive.
- The current-version fast path in `deft update` does not repair missing consumer scaffold files, including root hooks.

## Impact

Consumers can complete the v0.75 hybrid migration with no active pre-commit or pre-push enforcement. The canonical verification gate blocks setup completion, while the documented recovery commands cannot repair the state. Leaving `core.hooksPath=.githooks` configured would also disable default Git hooks silently because the directory is absent.

## Suggested acceptance criteria

- A documented v0.75-compatible repair command deposits or reconstructs the expected project-root hooks before setting `core.hooksPath`.
- Setup is transactional/fail-safe: it does not configure a missing or invalid hooks directory.
- Update/repair validates and refreshes consumer scaffold files even when the framework payload version is already current.
- A regression test covers an already-current hybrid deposit with missing project-root `.githooks/`.
- `deft verify:hooks-installed` passes after the supported setup/repair flow.

Refs #1709

MScottAdams:
## Summary
- Call `writeConsumerGitHooks` from `runRefreshDeposit` on every refresh path (including already-current / #2595-style no-op), so hybrid consumers missing project-root `.githooks/` self-heal via `deft update`.
- Make `task setup` transactional: refuse to set `core.hooksPath=.githooks` when `pre-commit`, `pre-push`, or `_deft-run.sh` are absent.
- Align `verify:hooks-installed` recovery prose and upgrade docs with `deft update` as the heal command.

Closes #2530.

## Test plan
- [x] `pnpm -w exec vitest run packages/core/src/init-deposit/refresh.test.ts packages/core/src/verify-env/verify-hooks-installed.test.ts`
- [x] `pnpm run lint`
- [ ] CI `task check` on GitHub (local swarm worktree path under `.deft-scratch` triggers a pre-existing skill-frontmatter discovery false-negative)


MScottAdams:
@greptileai review

PR #2614, opened 2026-07-16
fix(resolution): Windows-safe defaultEngineProbe for npm .cmd shims (#2606)

### LINKED ISSUE #2606: bug(resolution): defaultEngineProbe reports engine unreachable on Windows despite global npm install
## Summary

On Windows, `directive doctor` / `deft doctor` (and `directive update`'s self-heal ladder) report:

> Root cause: no Directive engine is reachable in the execution environment

…even when `@deftai/directive@0.78.0` is installed globally and both `directive` / `deft` work interactively in the same shell. Reproduced on a consumer after `deft update` (finalpilot, content v0.78.0, hybrid vendored deposit).

## Observed symptoms

1. `npm list -g @deftai/directive` → `@deftai/directive@0.78.0` (+ nested `@deftai/directive-core@0.78.0`)
2. `where.exe directive` / `where.exe deft` → `C:\Users\…\AppData\Roaming\npm\directive(.cmd)` / `deft(.cmd)`
3. Interactive `directive --version` / `deft --version` succeed and print:
   `@deftai/directive (engine: @deftai/directive-core@0.78.0)`
4. Same project: `deft doctor` Resolution section prints:
   - `Version reconciliation: engine unreachable, content 0.78.0, pin none`
   - Primary next action: `no Directive engine is reachable in the execution environment`
5. `deft update` prints the same false-negative ladder trace:
   `[deft update] engine self-heal (global-first ladder): global: absent; local: absent; install: npm i -g @deftai/directive@absent; install: deferred (no install runner supplied)`

## Root cause

`packages/core/src/resolution/classify.ts::defaultEngineProbe` / `probeEngineVersion` shells out with:

```ts
execFileSync(binary, ["--version"], {
  encoding: "utf8",
  timeout: 5000,
  stdio: ["ignore", "pipe", "ignore"],
});
```

for bare names `"directive"` then `"deft"`, **without** `shell: true` and without Windows PATHEXT / `.cmd` resolution.

On Windows, npm global bins are `.cmd` shims. Node's `execFileSync("directive")` / `execFileSync("deft")` does **not** resolve those shims → `ENOENT`. The probe returns `{ reachable: false, version: null }`.

### Reproduction (same machine, Node 24)

```js
const { execFileSync } = require("child_process");
for (const b of ["directive", "deft", "directive.cmd", "deft.cmd"]) {
  try {
    console.log("OK", b, execFileSync(b, ["--version"], {
      encoding: "utf8", timeout: 5000, stdio: ["ignore", "pipe", "ignore"],
    }).trim());
  } catch (e) {
    console.log("FAIL", b, e.code, e.message.split("\n")[0]);
  }
}
```

Result:

```
FAIL directive ENOENT spawnSync directive ENOENT
FAIL deft     ENOENT spawnSync deft ENOENT
FAIL directive.cmd EINVAL …
FAIL deft.cmd      EINVAL …
```

With `{ shell: true }` the same bare names succeed (and the existing `SEMVER_IN_TEXT_RE` would extract `0.78.0` from the banner). Direct `node <npm-prefix>/node_modules/@deftai/directive/dist/bin.js --version` also succeeds.

## Why doctor / update both lie

| Call site | How it consumes the probe |
|-----------|---------------------------|
| `doctor/main.ts::runResolutionDecision` | `classify()` → `plan()` row **3b** (`effectiveEngine === null`) → mode `blocked`, root cause string above |
| `init-deposit/refresh.ts::buildDefaultLadderFacts` | `globalEngineVersion: facts.engineReachable ? facts.engineVersion : null` → ladder rung 1 prints `global: absent` |

So one Windows-broken probe poisons both surfaces. Doctor is especially misleading: it is already running *as* the engine, yet still claims the engine is unreachable because it re-probes via subprocess instead of using in-process identity.

## Contributing factors (not primary)

- Consumer had **no committed `package.json` pin** (`pin none`). That alone would be a soft warning under plan row **3c** (`proceed` + advisory) if the engine probe succeeded — it is **not** what forces `blocked`.
- Tests for `defaultEngineProbe` only assert "never throws" (`classify.test.ts`); they do not assert Windows PATH / npm-shim resolution, so Linux CI green-washes this.

## Expected behavior

When a global npm engine is installed and on PATH (or when doctor/update are already executing inside `@deftai/directive`):

- `engineReachable: true`
- `engineVersion` parsed from the banner / package (e.g. `0.78.0`)
- doctor Resolution → `proceed` (or pin-advisory if no pin), **not** `blocked`
- update ladder → `global: 0.78.0 … -> use`, not `global: absent`

## Suggested fix direction

1. **In-process first** when already running inside the CLI/engine package (doctor/update self-identity) — avoid shelling out to discover the process that is already executing.
2. **Windows-safe PATH probe** as fallback:
   - resolve via `where` / PATHEXT, or
   - `execFileSync(process.execPath, [resolvedBinJs, "--version"])`, or
   - carefully use `shell: true` only for the known binary names (note DEP0190 if args are concatenated).
3. Add a regression test that simulates the Windows npm-shim case (or a CI Windows job assertion that `defaultEngineProbe().reachable === true` when the global engine is present).

## Environment

- OS: Windows 10/11 (`win32`)
- Node: v24.x
- Global: `@deftai/directive@0.78.0`
- Consumer: hybrid vendored `.deft/core` content `v0.78.0`, no `package.json` pin
- Source: `packages/core/src/resolution/classify.ts` (`probeEngineVersion` / `defaultEngineProbe`); consumers in `doctor/main.ts`, `init-deposit/refresh.ts`

## Related

- Resolution spine / global-first ladder: #2264 / epic #2203
- Runtime portability follow-ups: #2275

MScottAdams:
## Summary

- Fix `defaultEngineProbe` / `probeEngineVersion` so Windows npm global `.cmd` shims resolve via PATHEXT-aware PATH lookup and `spawnCommandText` shell retry (#2606).
- Prefer in-process `@deftai/directive-core` identity when doctor/update already execute inside the CLI, avoiding a false unreachable diagnosis.
- Add regression tests covering in-process identity, Windows shim resolution, and shell fallback paths.

Closes #2606

## Test plan

- [x] `pnpm -w exec vitest run packages/core/src/resolution/classify.test.ts packages/core/src/doctor/network-gate.test.ts`
- [x] `task verify:branch` + `task verify:forward-coverage`
- [x] `task check` (Windows hairline branch-coverage miss matches master baseline; Linux CI is authoritative)


MScottAdams:
@greptileai review

MScottAdams:
@greptileai review

Issue #1717, opened 2026-06-18
feat(ts-migration): TS monorepo scaffold + CI wiring (Wave 1)


MScottAdams:
## Parent

#1530 (Part 1 of 2; umbrella epic #1545)

## What to build

Stand up the TypeScript monorepo skeleton alongside the existing Python engine, with **no behavior moved** (zero parity risk). pnpm workspaces; `@deft/types` (leaf, zero-dep), `@deft/core` (domain, no I/O), and `@deft/cli` package skeletons; `tsc -b` project references; a Node CI lane (Vitest + ESLint + `tsc`) running alongside the Python pipeline. No monorepo orchestrator on day one -- incrementality comes from `tsc -b`, `vitest --changed`, and the in-engine content-hash cache tracked at #1713 (cross-reference, not duplicated here).

## Acceptance criteria

- [ ] pnpm workspaces + `@deft/types`/`@deft/core`/`@deft/cli` skeletons build with `tsc -b`
- [ ] Node CI lane (Vitest/ESLint/tsc) runs green alongside the Python pipeline
- [ ] No runtime behavior is ported in this slice (skeleton only)
- [ ] No monorepo orchestrator is introduced (no Nx, no Turborepo)
- [ ] `task check` passes

## Standing invariants (all Part-1 children)
- The Python suite is the **parity oracle** and runs **wholesale** (not affected-sliced) until deleted.
- `task check` stays green at every step (strangler-fig; no big-bang).
- The parity gate runs **cache-off** so a golden diff is never served from cache.
- No Python module is deleted until its TS replacement covers historical bug fixtures (not just happy-path) and CI usage.
- No agent-runtime dependency (that is Part 2, #1707); no monorepo orchestrator on day one (Nx never; Turborepo only on a measured trigger).

## Type

HITL

## Blocked by

None -- can start immediately

MScottAdams:
## Design pass — decisions locked (2026-06-18)

Foundational decisions for this slice, locked with the maintainer before build:

| Decision | Choice | Rationale |
|---|---|---|
| Package manager | **pnpm workspaces** | No monorepo orchestrator day-one (rev-3 RFC); `tsc -b` + `vitest --changed` cover incrementality |
| npm scope | **`@deftai/*`** | Matches the GitHub org; publishable artifact for the consumer-adoption path |
[...]

PR #2848, opened 2026-07-26
fix(hooks): realpath-aware entrypoint guard for npm bin symlinks (#2846)

### LINKED ISSUE #2846: bug(hooks): deft-hook symlink shim never runs main() (argv[1] realpath mismatch) - empty stdout blocks all Cursor edit tools and silently fails open on Claude/Grok/Codex
## Summary

In `@deftai/directive@0.85.0`, the new lightweight `deft-hook` entrypoint introduced by #2790 never executes its `main()` function when it is invoked through the symlink npm installs into the global bin directory. The process exits 0 with completely empty stdout, so every hook decision — allow *and* deny — is discarded before it reaches the host.

The cause is the module entrypoint guard in `dist/hook-bin.js`, which compares `process.argv[1]` against `fileURLToPath(import.meta.url)` as raw strings. Node does not resolve symlinks when populating `process.argv[1]`, so the two values are never equal when the file is reached through the `deft-hook` symlink.

On Cursor this is not a soft failure. The deposited `.cursor/hooks.json` sets `"failClosed": true` on both `preToolUse` matchers, and Cursor treats empty stdout under that flag as hook failure, so it blocks the tool outright. Every edit tool is unusable. On Claude, Grok and Codex the same empty stdout is indistinguishable from a normal allow, so the Tier-1 write guard silently fails open and deny decisions are lost with no diagnostic.

The policy engine itself is healthy. Running the exact same file through `node` with a real path produces correct JSON. Only the shim invocation is broken.

## Environment

- `@deftai/directive` 0.85.0, `@deftai/directive-core` 0.85.0, `@deftai/directive-content` 0.85.0
- Node v22.19.0
- macOS (darwin 25.5.0), zsh
- npm global prefix `~/.local`, so the shim is `~/.local/bin/deft-hook -> ../lib/node_modules/@deftai/directive/dist/hook-bin.js`
- Consumer repo on the vendored `.deft/core/` hybrid layout

## Reproduction

```
$ echo '{"tool_name":"Write","file_path":"/tmp/x"}' | deft-hook --host cursor --event tool.before
                                     # empty stdout, exit 0

$ echo '{"tool_name":"Write","file_path":"/tmp/x"}' \
    | node "$(readlink -f ~/.local/bin/deft-hook)" --host cursor --event tool.before
{"permission":"allow"}               # exit 0
```

The two invocations run byte-identical code. The only difference is whether `process.argv[1]` is the symlink path or the real path.

The loss of *deny* decisions is the more serious half, and it reproduces on all four hosts. Using an empty payload, which the dispatcher is designed to deny:

```
$ printf '' | node "$(readlink -f ~/.local/bin/deft-hook)" --host claude --event tool.before
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny",
 "permissionDecisionReason":"Directive denied this matched write event because the host payload omitted its tool name."}}

$ printf '' | deft-hook --host claude --event tool.before
                                     # empty stdout, exit 0
```

The same contrast holds for `--host grok`, `--host codex` and `--host cursor`: each produces a correctly shaped host-specific deny through `node` with the real path, and nothing at all through the symlink.

## Expected vs actual

Expected: invoking `deft-hook` writes a host-appropriate decision document to stdout — `{"permission":"allow"}` for Cursor allows, empty for Claude/Grok/Codex allows, and a host-shaped deny object for denials.

Actual: invoking `deft-hook` writes nothing and exits 0 regardless of the decision, because `main()` is never called.

## Impact per host

The deposited configs differ in how they interpret empty stdout, so the same defect produces two different failure modes.

Cursor deposits `"failClosed": true` on both `preToolUse` matchers — the edit matcher (`Edit|Write|WriteFile|CreateFile|MultiEdit|NotebookEdit|StrReplace|SearchReplace|Delete|DeleteFile|ApplyPatch|apply_patch`) and the subagent-dispatch matcher (`Task|SubagentStart|spawn_subagent|start_agent|CreateAgent`). Empty output therefore reads as hook failure and Cursor blocks the tool with:

```
Tool blocked because this hook is configured to fail closed (block when it fails).
Hook "deft-hook --host cursor --event tool.before" returned no output.
```

Every edit tool is blocked outright, and subagent dispatch is blocked as well. This was confirmed on two independent agent sessions in the same project before this report was written. While writing this report a third session hit it again: the attempt to write the draft to a scratch file was itself blocked with exactly the message above, which is as direct a demonstration of the impact as can be offered.

Claude (`.claude/settings.json`), Grok (`.grok/hooks/deft.json`) and Codex (`.codex/hooks.json`) were all migrated to `deft-hook` by the same change, but none of them deposit a fail-closed flag. For these hosts `renderHostDecision` deliberately returns empty stdout to signal allow and preserve the host permission flow, which means the broken shim's empty output is read as an unconditional allow. These hosts are not blocked, but they lose the guardrail entirely: denials are silently dropped and nothing surfaces the failure. That is the same fail-open-in-a-fail-closed-control shape as #2829.

Neither outcome is acceptable, and the Cursor one is arguably worse than having no guard at all, since the operator cannot edit any file and the framework reports itself as healthy.

## Root cause

`dist/hook-bin.js` in full:

```js
#!/usr/bin/env node
import { fileURLToPath } from "node:url";
import { run } from "./hook-dispatch.js";
export function main(argv = process.argv.slice(2)) {
    return run(argv);
}
if (process.argv[1] !== undefined && fileURLToPath(import.meta.url) === process.argv[1]) {
    process.exit(main());
}
```

When invoked as `deft-hook`:

- `process.argv[1]` is `/Users/<user>/.local/bin/deft-hook` (Node does not resolve the symlink)
- `fileURLToPath(import.meta.url)` is `/Users/<user>/.local/lib/node_modules/@deftai/directive/dist/hook-bin.js`

The strings never match, the guard is false, `main()` is never called, the module finishes evaluating with no side effects, and Node exits 0 having written nothing.

Notably, the equivalent guard in `dist/verify-encoding.js` carries a comment explaining that both sides are normalized through `fileURLToPath` specifically so the guard fires on Windows, where `process.argv[1]` is a native backslash path while `import.meta.url` is a forward-slash `file://` URL. That normalization is correct as far as separators go, but it addresses only one of the two ways the two values can disagree. Symlink indirection is the other, and it was not covered.

## Why `dist/bin.js` is unaffected — the useful contrast

`dist/bin.js`, which backs the working `deft`, `directive` and `deft-ts` shims, has no entrypoint guard at all:

```js
#!/usr/bin/env node
import { routeAndDispatch } from "./cli-router/index.js";
const code = await routeAndDispatch(process.argv.slice(2));
process.exit(code);
```

It executes unconditionally at module top level using top-level await. There is nothing to compare and therefore nothing to get wrong, which is exactly why `deft --version` works through its symlink while `deft-hook` does not. The guard pattern is what distinguishes the broken entrypoint from the working one, not anything about the hook dispatch logic.

## Other symlinked entrypoints share the defect

`package.json` declares five bins:

```json
{
  "directive": "./dist/bin.js",
  "deft": "./dist/bin.js",
  "deft-ts": "./dist/bin.js",
  "deft-hook": "./dist/hook-bin.js",
  "deft-verify-encoding": "./dist/verify-encoding.js"
}
```

Three of them point at the unguarded `dist/bin.js` and work. Both of the remaining two are guarded, and both are broken. `deft-verify-encoding` is silently a no-op through its symlink:

```
$ deft-verify-encoding --help
                                     # empty stdout, exit 0

$ node "$(readlink -f ~/.local/bin/deft-verify-encoding)" --help
verify_encoding: unrecognized argument: --help
                                     # exit 2
```

The guarded form is not limited to these two files. Roughly fifty modules under `dist/` use the identical `process.argv[1] !== undefined && fileURLToPath(import.meta.url) === process.argv[1]` construct. Those are latent rather than active, because they are only reachable by direct real path today, but any of them that later becomes a declared bin will fail the same way. One module, `dist/plan-sequence.js`, uses a different form — `resolve(process.argv[1]) === fileURLToPath(import.meta.url)` — which handles relative paths but still does not resolve symlinks, so it would also fail if it were installed as a bin.

Whatever fix is chosen is probably worth applying to the shared pattern rather than only to `hook-bin.js`.

## Proposed fix

Compare resolved real paths instead of raw strings, so symlink indirection cannot defeat the comparison:

```js
import { realpathSync } from "node:fs";
import { fileURLToPath } from "node:url";

function isDirectEntrypoint(moduleUrl) {
  const entry = process.argv[1];
  if (entry === undefined) return false;
  try {
    return realpathSync(entry) === realpathSync(fileURLToPath(moduleUrl));
  } catch {
    return false;
  }
}

if (isDirectEntrypoint(import.meta.url)) {
  process.exit(main());
}
```

Resolving both sides matters: the module path can itself sit under a symlinked directory, which is common with `pnpm` layouts and with `npm link`. The `try`/`catch` keeps the guard from throwing if `argv[1]` names something that no longer exists.

Extracting this into a single shared helper — something like `dist/entrypoint.js` exporting `isDirectEntrypoint(import.meta.url)` — would fix `hook-bin.js`, `verify-encoding.js` and every latent copy at once, and would give future bins a correct default. It would also retain the Windows separator normalization that the current form was written to provide, since `realpathSync` returns native separators on both sides.

An alternative worth mentioning only to reject it: dropping the guard entirely, as `bin.js` does, would work for the bins but would break the test suites that import these modules for their exports, which is what the guard exists to protect.

## Suggested regression test

The gap that let this reach release is that nothing executes the installed entrypoint and inspects its output. A structural check cannot catch it, because the config was structurally perfect the whole time. The test needs to invoke the shim the way a host does.

As a packaging or post-install smoke test:

```js
import { execFileSync } from "node:child_process";
import assert from "node:assert/strict";

// Invoke through the installed bin name, not the dist real path,
// so symlink resolution is part of what is under test.
const stdout = execFileSync("deft-hook", ["--host", "cursor", "--event", "tool.before"], {
  input: JSON.stringify({ tool_name: "Write", file_path: "/tmp/x" }),
  encoding: "utf8",
});

assert.notEqual(stdout.trim(), "", "deft-hook produced no stdout when invoked via its installed shim");
assert.doesNotThrow(() => JSON.parse(stdout), "deft-hook stdout was not parseable JSON");
assert.equal(JSON.parse(stdout).permission, "allow");
```

Two properties are worth asserting beyond non-emptiness. First, that a deny payload actually produces a deny through the shim, since a guard that allows everything would pass a non-empty-stdout check on the allow path alone. Second, that the invocation goes through a real symlink; a test that resolves the real path first would have passed against 0.85.0 and caught nothing.

A cheap and hermetic variant that needs no global install is to create a temporary symlink to `dist/hook-bin.js` in a scratch directory and invoke that, which reproduces the failure exactly.

## Suggested doctor coverage

`deft doctor --full` currently reports `agent-hooks-registration: registered and structurally valid` against a completely non-functional hook. That is accurate to what it measures and useless for this failure. `evaluateAgentHooks` in `dist/verify-env/agent-hooks.js` imports only `statSync` plus the deposit inspector and the host-hooks policy loader; it never spawns the hook command, so it cannot observe that the binary produces nothing. `deft agents:refresh` does not help either, since the deposit it writes is already correct.

A live probe would close this. Piping a synthetic payload through the configured hook command and warning when stdout is empty or unparseable would have flagged this immediately, and it would keep flagging any future breakage in the invocation path rather than only in the config shape. Given that the Cursor deposit is fail-closed, an empty-stdout hook is precisely the condition most worth surfacing as a P0 doctor finding.

## Related

- Refs #2790 — introduced the `deft-hook` entrypoint and retired `.cursor/hooks/deft-cursor-hook-adapter.mjs`, shipped in v0.85.0
- Refs #2438 / #2437 — the `deft-hook` harness and the agent-host-hooks-as-Tier-1-enforcement epic
- Refs #2779 — same user-visible `returned no output` symptom, different cause. That fix is intact and verifiably working: `renderHostDecision` does return `{"permission":"allow"}` for Cursor, as the direct-`node` reproduction above shows. The present defect sits upstream of it, discarding the output before the host ever sees it.
- Refs #2829 — fail-open in a fail-closed control; the Claude/Grok/Codex half of this report is the same family
- Refs #2838 — the other v0.85.0 adapter-retirement follow-up

MScottAdams:
## Summary
- Add shared `isDirectEntrypoint` helper that realpath-compares `process.argv[1]` and `import.meta.url`, fixing npm bin symlink invocations where `main()` never ran (empty stdout exit 0).
- Wire `deft-hook` and `deft-verify-encoding` to the helper; add hermetic symlink regression tests (Linux CI; skipped on win32 without symlink privilege).
- CHANGELOG [Unreleased] cites #2846.

## Test plan
- [x] `pnpm exec vitest run packages/cli/src/entrypoint.test.ts packages/cli/src/hook-bin.test.ts packages/cli/src/verify-encoding.test.ts`
- [x] `task verify:forward-coverage`
- [x] `task check`

Closes #2846


MScottAdams:
owner: MScottAdams
monitor_agent_id: leaf-2846-hook-symlink
head_sha: 07eaca0611d57c7005dfb65bf858d544a71e4a1b
started_at: 2026-07-26T17:31:04.800Z
expires_at: 2026-07-26T18:01:04.800Z
platform_primitive: cursor-task

MScottAdams:
## Review ack — dist build gate (post-strip)

Per operator-approved simplification in cd879310, this PR **removed** the Vitest `globalSetup` / lock / in-worker `tsc -b` harness entirely.

**How dist is obtained now**
- `packages/cli/dist/hook-bin.js` and `verify-encoding.js` are produced by the **normal build-before-test gate** (`task check` / CI `tsc -b packages/cli`), same as every other dist-backed subprocess test in this repo.
- `cli-bin-symlink-entrypoint.test.ts` fails fast in `beforeAll` with a clear message if dist is missing (`run task check` / `pnpm --filter @deftai/directive-cli run build`).
- No Vitest worker runs `tsc -b`; no cross-worktree lock; no parallel dist-write races from test orchestration.

**Residual Greptile themes addressed**
- *Prebuilt dist required* — intentional; matches repo convention (#1717: vitest aliases src; `tsc -b` is the emit gate).
- *Parallel build / lock / stale dist* — removed with the globalSetup strip; net diff no longer touches `vitest.config.ts` globalSetup beyond master.

CI green, P0=0, P1=0 on current HEAD. Confidence holdout acknowledged; proceeding under operator override for this narrow case.

PR #2872, opened 2026-07-27
fix(hooks): surface decision codes and harden Cursor tool.before spawn path (#2864)

### LINKED ISSUE #2864: bug(hooks): Cursor tool.before intermittently blocks Task spawns on macOS — host reports "exit code 1" but no hook:dispatch path exits 1
Filed as a **consumer of the framework**, not a maintainer — apologies if any of this is already understood internally. Everything below was verified against the installed npm package on one project; I have not read the framework source repo. Same install as #2826 and #2862, so those three reports can be correlated.

The framework's own denial text asks for this report ("If write tools should pass, update Directive or report the payload shape from Cursor."), so I've tried to bring the payload-shape and exit-code evidence it's asking for.

## Summary

On macOS + Cursor, the Directive-managed `preToolUse` entry that matches `Task|SubagentStart|spawn_subagent|start_agent|CreateAgent` **intermittently** blocks a sub-agent dispatch. Because the entry is `failClosed: true`, the block is hard — the tool call never runs.

Two things I can state as fact, and one I cannot:

1. **Cursor reported the hook as `failed with exit code 1`.** I could not find any code path in `@deftai/directive@0.85.0` that exits 1. `run()` in `dist/hook-dispatch.js` returns `0` for every verdict (allow *and* deny) and `2` only for an argv parse error. I verified this across nine stdin shapes (matrix below).
2. **Cursor rendered its own generic fail-closed text, not the Directive `user_message`.** On the deny path Directive emits `{"permission":"deny","user_message":"…"}` and other reports (#2713, #2860) show Cursor surfacing that text in the UI. We got the generic wrapper instead, which suggests the in-session event was a **process-level failure**, not a `deny` verdict.
3. **What made it fail is unknown to me.** I can reproduce two independent mechanisms offline that both end in a fail-closed block (empty-stdin deny; blocking-stdin read killed by the host timeout), but I cannot observe what Cursor actually delivered on the failing call. I am not going to present either as the cause.

Separately, and reproducible on demand: `deft hook:dispatch --host cursor --event tool.before` with **empty stdin** returns a `deny` verdict **with exit code 0**. That specific behaviour is already noted as *Defect 3* of the open #2860 — see the dedup section at the bottom.

## Environment

| | |
|---|---|
| CLI | `@deftai/directive@0.85.0` (engine `@deftai/directive-core@0.85.0`, content `@deftai/directive-content@0.85.0`), global npm install, prefix `~/.local` |
| Vendored content | `.deft/core` at `v0.84.0` (`managed_by: npm`, `fetched_at: 2026-07-23T23:11:50Z`) |
| `xbrief/.deft-version` | `0.84.0` |
| Host | Cursor `3.13.10` |
| Node | `v22.19.0` |
| OS / shell | macOS 26.5.2 (darwin `25.5.0`), arm64, zsh |
| Layout | consumer install, `xbrief/` v0.8 hybrid (not a framework source checkout) |
| Hook command in use | legacy `deft hook:dispatch` (see "Stale deposit" note below) |

Gate state at the time of the block was healthy: `.deft/ritual-state.json` present and recorded for this session, one brief in `xbrief/active/` with `plan.status: running`. A well-formed `Task` payload probed against the same CLI in the same working tree returns `{"permission":"allow"}`, so this is not a ritual or scope failure.

## Observed behaviour

Verbatim, from the Cursor UI:

```text
Error: Task blocked by preToolUse hook: Tool blocked because this hook is configured to fail closed (block when it fails). Hook "deft hook:dispatch --host cursor --event tool.before" failed with exit code 1
```

Verbatim, running the same command by hand in the project root with no stdin:

```json
{"permission":"deny","user_message":"Directive denied this Cursor preToolUse event because the host sent an empty payload (stdin was empty — not a session ritual or scope failure). If write tools should pass, update Directive or report the payload shape from Cursor.","agent_message":"Directive denied this Cursor preToolUse event because the host sent an empty payload (stdin was empty — not a session ritual or scope failure). If write tools should pass, update Directive or report the payload shape from Cursor."}
```

Exit code for that manual invocation: **0**.

### It is intermittent

Four `Task` dispatches earlier in the same Cursor session, same repository, same working tree, were allowed. The fifth was blocked. An immediate retry of the byte-identical call succeeded. Nothing about the payload, the project, the ritual state, or the active brief changed between the blocked call and the successful retry.

That pattern rules out a deterministic policy decision and points at either a payload-delivery race or a per-invocation process failure.

## Exit-code matrix

All rows run from the project root against the installed CLI, `--host cursor --event tool.before`. `verdict` is the `permission` field of stdout.

| # | stdin | exit | verdict | `code` implied by message |
|---|---|---|---|---|
| A | closed (`<&-`) | **0** | `deny` | empty payload / `stdinEmpty` |
| B | `< /dev/null` | **0** | `deny` | empty payload / `stdinEmpty` |
| C | `printf '' \|` | **0** | `deny` | empty payload / `stdinEmpty` |
| D | whitespace only (`'   \n'`) | **0** | `deny` | empty payload / `stdinEmpty` |
| E | `echo '{}'` | **0** | `deny` | omitted tool name / `invalid-input` |
| F | `echo 'not json'` | **0** | `deny` | not valid JSON / `parseFailed` |
| G | well-formed `Task` + `subagent_type: generalPurpose` | **0** | `allow` | `spawn-ready` |
| H | well-formed `Task` + `subagent_type: explore` | **0** | `allow` | `spawn-explore-ready` |
| I | well-formed `Write` | **0** | `allow` | `write-ready` |
| J | `--event bogus` (argv error) | **2** | — | stderr `unsupported event: "bogus"` |

Plus a 40-iteration sequential loop with a well-formed `Task` payload: **40/40 exit 0**, and a 12-way parallel burst: **12/12 exit 0, `allow`**.

**What the matrix reveals:** exit code and verdict are fully decoupled. A `deny` is indistinguishable from an `allow` by exit status — both are `0`. And nothing returns `1`. So the two candidate statements about this bug are genuinely different bugs:

- *"the hook denies on empty payload"* — true, reproducible, exits 0 (rows A–D).
- *"the hook exits non-zero"* — **not reproducible at all** on this install. Whatever Cursor observed on the failing call, it was not `run()` returning 1.

The remaining explanations for a host-reported "exit code 1" are (a) the process died before `run()` returned, (b) Cursor is reporting its own interpretation of a failure rather than a real `waitpid` status, or (c) something in the spawn environment differs from my shell. I could not discriminate between these from the consumer side.

## What the hook actually does with stdin

`@deftai/directive@0.85.0`, `dist/hook-dispatch.js`. The read is a **single, unconditional, synchronous read of fd 0**, with no retry, no timeout, and no distinction between "the writer closed without sending anything" and "nothing has arrived yet":

```js
const readStdin = seams.readStdin ?? (() => readFileSync(0, "utf8"));
const cwd = (seams.cwd ?? process.cwd)();
const { payload, context: payloadContext } = parsePayload(readStdin());
```

`parsePayload` converts a zero-length read straight into the deny context:

```js
export function parsePayload(raw) {
    if (raw.trim().length === 0) {
        return { payload: {}, context: { stdinEmpty: true } };
    }
    // … BOM strip, JSON.parse, free-form ApplyPatch synthesis, parseFailed …
}
```

`decideHook` in `@deftai/directive-core@0.85.0` `dist/hooks/dispatcher.js` then fails the tool-name lookup and denies:

```js
const toolName = hookToolName(input.payload, input.host);
if (toolName === null) {
    return deny(input, "invalid-input", null, missingToolNameMessage({ … }));
}
```

and `missingToolNameMessage` produces the exact string we saw:

```js
if (host === "cursor") {
    if (context?.stdinEmpty) {
        return ("Directive denied this Cursor preToolUse event because the host sent an empty payload " +
            "(stdin was empty — not a session ritual or scope failure). " +
            "If write tools should pass, update Directive or report the payload shape from Cursor.");
    }
    …
```

Finally `run()` returns `0` regardless of verdict — the last statement of the function, with no `return 1` anywhere above it:

```js
if (decision.code === "session-start-degraded") writeErr(`${decision.message}\n`);
…
return 0;
```

## How the fail-closed behaviour is wired

Project `.cursor/hooks.json`, deposited by Directive. Both `tool.before` entries carry `failClosed: true` and `timeout: 5`:

```json
{
  "version": 1,
  "hooks": {
    "sessionStart": [
      { "command": "deft hook:dispatch --host cursor --event session.start", "timeout": 5 }
    ],
    "preToolUse": [
      {
        "command": "node .cursor/hooks/deft-cursor-hook-adapter.mjs ApplyPatch",
        "matcher": "ApplyPatch|apply_patch",
        "failClosed": true,
        "timeout": 5
      },
      {
        "command": "deft hook:dispatch --host cursor --event tool.before",
        "matcher": "Edit|Write|WriteFile|CreateFile|MultiEdit|NotebookEdit|StrReplace|SearchReplace|Delete|DeleteFile",
        "failClosed": true,
        "timeout": 5
      },
      {
        "command": "deft hook:dispatch --host cursor --event tool.before",
        "matcher": "Task|SubagentStart|spawn_subagent|start_agent|CreateAgent",
        "failClosed": true,
        "timeout": 5
      }
    ],
    "preCompact": [
      { "command": "deft hook:dispatch --host cursor --event session.compact", "timeout": 5 }
    ]
  }
}
```

The `Task` matcher is the third entry, and it is the bare `deft hook:dispatch` command — no adapter, so the `spawnSync`-normalisation path from #2829 does not apply here. `failClosed: true` is what converts both a `deny` verdict *and* a process failure into an identical hard block, which is also why the two are so hard to tell apart from the operator's seat.

The generator that writes these entries is `@deftai/directive-core@0.85.0` `dist/init-deposit/agent-hooks.js`, which sets `failClosed: true` and `timeout: 5` on both `tool.before` groups.

## Two offline mechanisms that both end in a fail-closed block

Neither of these is a claim about what Cursor did. They are the two ways I could make the installed hook misbehave by varying only how stdin is delivered, both of which produce a blocked tool call. Harness: `child_process.spawn` of `deft hook:dispatch --host cursor --event tool.before`, cwd = project root, with a synthetic host timeout kill at 5s to match `timeout: 5`.

| # | stdin delivery | elapsed | exit | signal | result |
|---|---|---|---|---|---|
| K | non-blocking FIFO, payload written at t+300ms | 824ms | 0 | — | `{"permission":"allow"}` |
| L | non-blocking FIFO, **no writer ever attaches** | 773ms | 0 | — | `deny` — empty payload |
| M | blocking pipe, write end held open, **no data written** | 5032ms | `null` | `SIGTERM` | killed by the 5s timeout, **no stdout at all** |
| N | blocking pipe, payload written at t+1500ms | 1560ms | 0 | — | `{"permission":"allow"}` |

Row **L** is the interesting one: when fd 0 is a non-blocking descriptor with no writer attached, `readFileSync(0)` returns **zero bytes immediately** — indistinguishable, to `parsePayload`, from a host that deliberately sent nothing. The hook denies without ever waiting. Row **K** is the same descriptor with a writer that attaches 300ms later and it works fine, so the boundary between L and K is purely *whether the payload was ready when the process happened to read*. That is a race, and it is the shape of race that would produce exactly the intermittency we see.

Row **M** is the other end: with an ordinary blocking pipe, `readFileSync(0)` blocks **indefinitely** waiting for EOF. If the host never writes, the hook hangs until the 5s timeout kills it, emits nothing, and `failClosed` blocks the tool. A signal-killed process has no exit code at all, which is one way a host could end up reporting a synthesised "exit code 1".

### Timeout headroom is thinner than it looks

Measured on an otherwise idle machine, project root, well-formed payload:

| invocation | per-call | vs `timeout: 5` |
|---|---|---|
| `deft hook:dispatch …` (what's deposited here) | ~1.10s mean, **2.55s peak** over 40 runs | ~2x margin at peak |
| `deft hook:dispatch …`, 12-way parallel | 1.56–1.78s | ~3x margin |
| `deft-hook …` (the fast entrypoint from #2790/#2795) | ~0.24s | ~20x margin |

A 2.55s peak against a 5s budget is not a lot of room on a machine under real load with concurrent hook invocations. I'm not claiming we hit the timeout — only that the margin is ~2x, not ~10x, on the routed `deft hook:dispatch` path.

## Stale deposit note (may be relevant, may be a red herring)

This project's `.cursor/hooks.json` uses `deft hook:dispatch`, which `@deftai/directive-core@0.85.0` itself classifies as the **legacy** command:

```js
export const DEFT_HOOK_COMMAND_MARKER = "deft-hook";
export const LEGACY_DEFT_HOOK_COMMAND_MARKER = "deft hook:dispatch";
```

So we are on the slower legacy route rather than the current `deft-hook` entrypoint. Worth noting that on this install the current `deft-hook` entrypoint is *also* broken — it is the #2846 symlink no-op (`~/.local/bin/deft-hook -> …/dist/hook-bin.js`, `process.argv[1]` never equals `fileURLToPath(import.meta.url)`), which I re-confirmed on 0.85.0:

```console
$ echo '{"tool_name":"Task",…}' | deft-hook --host cursor --event tool.before
                                        # empty stdout, exit 0

$ echo '{"tool_name":"Task",…}' | node "$(readlink -f ~/.local/bin/deft-hook)" --host cursor --event tool.before
{"permission":"allow"}                  # exit 0
```

#2846 is closed and #2848 is the fix, so this is presumably just "fixed on master, not yet in a published release" — flagging it only because it means a consumer on 0.85.0 cannot migrate off the legacy command to get the 4x latency headroom. No action needed if #2848 ships as expected.

## Reproduction

The intermittent block itself I cannot give you a deterministic recipe for — that's the core of the report. What *is* deterministic:

**Empty-stdin deny with exit 0** (rows A–D):

```console
$ cd <deft-managed project root>
$ deft hook:dispatch --host cursor --event tool.before < /dev/null
{"permission":"deny","user_message":"Directive denied this Cursor preToolUse event because the host sent an empty payload …"}
$ echo $?
0
```

**Verdict/exit-code decoupling** (rows A–I): same command with any of the stdin shapes in the matrix; exit is `0` for both `allow` and `deny`.

**Zero-byte read on a non-blocking descriptor** (row L):

```js
// node, from the project root
import { spawn, execSync } from "node:child_process";
import fs from "node:fs";
execSync("rm -f /tmp/deftfifo && mkfifo /tmp/deftfifo");
const fd = fs.openSync("/tmp/deftfifo", fs.constants.O_RDONLY | fs.constants.O_NONBLOCK);
const child = spawn("deft", ["hook:dispatch", "--host", "cursor", "--event", "tool.before"],
  { stdio: [fd, "pipe", "pipe"], cwd: process.cwd() });
child.stdout.on("data", (d) => process.stdout.write(d));
// -> {"permission":"deny", … "the host sent an empty payload" …}, exit 0, in <1s
```

**Blocking-read hang** (row M): same script with `stdio: ["pipe", …]` and never writing to `child.stdin` — the hook does not return until killed.

**In-session**, for whatever it's worth: open a Cursor Agent chat in a Deft-managed repo on macOS with a healthy ritual + active brief, and dispatch several sub-agents in a row. Four succeeded and the fifth was blocked for us; an immediate retry succeeded.

## What I could not determine

Stating these plainly rather than guessing, because the honest unknowns are load-bearing here:

- **Why the payload would ever be empty.** This is the heart of the bug and I don't know. I never observed a live empty payload — only that the code turns a zero-byte read into a deny, and that a zero-byte read is easy to produce by varying stdin delivery. Whether Cursor ever actually delivers zero bytes on this matcher is unverified. A stdin-capture wrapper like the one in #2713 would settle it; I did not want to install one on a live project mid-session.
- **Whether the in-session block was the empty-payload deny at all.** The evidence points *away* from it: Cursor showed its generic fail-closed text rather than Directive's `user_message`, and reported an exit code that no Directive path produces. My best read is that these are two adjacent defects on the same surface and the operator conflated them (understandably — `failClosed` makes them look identical).
- **What produced "exit code 1".** Not `run()`. Beyond that I can only list candidates (process death before return, host-synthesised status, spawn-environment difference) without evidence to choose.
- **Whether the 5s timeout was involved.** The 2.55s peak is suggestive but I did not observe a timeout, and Cursor did not report one.

## Impact

Trying not to inflate this: **friction, not a hard blocker.** A retry cleared it immediately, and the underlying gates were healthy the whole time. For an operator watching the chat, the cost is one confusing error and one retry.

The reason I think it's still worth fixing:

- **A fail-closed control that intermittently blocks legitimate calls trains people to retry past it.** That is corrosive to a gate whose value is that a block means something. It is also the mirror image of #2829's finding on the adapter path — the same surface fails open there and fails closed spuriously here, and neither is the gate doing its job.
- **An agent that doesn't know to retry will derail.** A sub-agent dispatch that comes back as a hook failure looks like a policy refusal, not a transient. An orchestrator following the framework's own fail-closed discipline will reasonably stop rather than retry a gate it believes just denied it.
- **The deny message is addressed to a human, not a machine.** "If write tools should pass, update Directive or report the payload shape from Cursor" is good operator copy, but there is no machine-readable reason code on the wire — the `code` field (`invalid-input`, `stdin-empty`, `ritual-not-ready`, …) exists inside `decideHook` and is dropped by `renderHostDecision`. An agent cannot tell "your gates aren't satisfied, go run `deft session:start`" apart from "the host handed me nothing, retry".
- **It is silent.** Nothing is logged on the empty-payload path except the host-facing JSON, so there's no post-hoc way to count how often this happens.

## Suggested directions — reporter's opinion, not a request

I don't know the constraints here and any of these may be wrong for reasons I can't see. Offered in case they're useful:

1. **Emit the `code` on the wire.** `renderHostDecision` already has `decision.code`; adding it to the Cursor payload (and the other hosts) would let an agent distinguish a real policy deny from a host-integration failure without parsing English. Cheapest change, and it makes the rest of this diagnosable.
2. **Make the exit code agree with the verdict**, or document loudly that it never will. Right now `allow` and `deny` are both `0`, which means no host can use exit status as a signal and every host must parse stdout. If a `deny` exited non-zero, `failClosed` hosts would behave correctly by construction — though that's a breaking change for hosts that already parse stdout, so possibly not worth it.
3. **Treat an empty payload as allow-with-warning rather than deny**, at least on the spawn matcher. A zero-byte read is a host-integration failure, not evidence of a policy violation — the gate has no information, and denying is asserting something it doesn't know. If that's too fail-open for a security control, an explicit bounded re-read (or a short poll for the first byte) before concluding "empty" would at least distinguish "the host sent nothing" from "nothing had arrived yet", which row K/L shows is currently conflated.
4. **Consider whether `timeout: 5` is right for the legacy `deft hook:dispatch` route.** At a 2.55s peak it's a ~2x margin; the `deft-hook` route is ~20x. Once #2848 ships and consumers migrate this mostly goes away, but the migration itself (#2860 Defect 2) doesn't seem to be happening cleanly.

## Relationship to existing issues — please close as duplicate if you disagree with the split

I searched `hook:dispatch`, `preToolUse`, `empty payload`, `fail closed`, `cursor hook`, and `stdin` before filing, and there is a lot of adjacent history. Being explicit about the overlap so you can fold this wherever it belongs:

- **#2860 (open) — closest overlap.** Its *Defect 3* is exactly the empty-stdin deny with exit 0, and says "Whether this is the same path Cursor exercises is unclear." This report is partly an answer to that sentence: the deny is real and reproducible, but on our machine the *in-session* failure looks like a different path (Cursor's generic fail-closed text, a nonexistent exit code). Differences from #2860: **macOS not win32**, the **spawn/`Task`** matcher not writes, **intermittent** not "denies every write", and #2860's Defects 1 and 2 (win32 doubled drive prefix, `deft update` preserving stale entries) don't apply here. **If you'd rather have one tracker, fold this into #2860 Defect 3** — I've left a pointer comment there.
- **#2699 → #2713 (both closed) — the same `failed with exit code 1` string.** #2699 reported the byte-identical Cursor error; it was absorbed into #2713 and explicitly flagged as "may be a distinct failure mode — keep until diagnostics discriminate." #2713's diagnostics then pinned the *live* Write/Task denials on a UTF-8 BOM (→ #2734, shipped via #2735) and the residual free-form `ApplyPatch` on #2738. As far as I can tell **the exit-code-1 symptom itself was never explained** — the closing note attributes the related "no exit status" observation to a sandbox artifact, which is a different symptom. So that thread's loose end is still loose, and this report is a second, independent sighting on a different OS four releases later.
- **#2829 (closed) — the mirror-image defect.** Its analysis of `spawnSync` `status === null` under timeout kill is what made me test row M. It fixed the *adapter* path; the bare `deft hook:dispatch` entries that the `Task` matcher uses were out of its scope, and they fail *closed* under the same conditions.
- **#2846 / #2848 (closed) — the `deft-hook` symlink no-op**, re-confirmed present in published 0.85.0 above. Context only; blocks migrating off the legacy command.
- **#2790 / #2795 (closed) — hook cold-boot latency.** Relevant to the timeout-margin numbers; the legacy route this project is deposited with doesn't get that improvement.
- **#2852 (closed) — `doctor` live-probe for empty/unparseable hook stdout.** Adjacent: that's the observability gap that would have made this visible without hand-built spawn harnesses.
- Checked and **not** this: #2734/#2735 (BOM — our payloads parse), #2738 (`ApplyPatch` free-form — different matcher, and the adapter is a separate entry here), #2685/#2669/#2677 (Write tool-name inference — our probes classify correctly), #2787/#2789/#2793 (win32 path normalisation).

MScottAdams:
## Summary

Cursor `preToolUse` for Task/spawn matchers intermittently blocks legitimate spawns under `failClosed: true`. Two concrete gaps: (1) `decision.code` was dropped by `renderHostDecision`, so agents could not distinguish policy denials from host-integration failures; (2) a single empty stdin read treated delayed payloads as true-empty denies.

## Changes

- Emit `code` on Cursor allow/deny JSON (`spawn-ready`, `stdin-empty`, `invalid-input`, …)
- New decision code `stdin-empty` for true empty payloads (distinct from `invalid-input`)
- `readStdinHardened`: brief empty-read re-poll budget before concluding empty
- Document exit-code contract: **0** for rendered allow/deny, **2** for argv errors only
- Regression tests: empty stdin, delayed stdin, code-on-wire, exit 0/2

## Test plan

- [x] `vitest` packages/core/src/hooks + hook-dispatch
- [x] `task ts:check-lane` (full coverage gate, 85% branch)
- [x] `task verify:branch` / encoding / forward-coverage / scm-boundary / changelog / orphan-active

Closes #2864


MScottAdams:
ended_at: 2026-07-27T22:58:56.826Z

Issue #2566, opened 2026-07-15
bug(scope,spec): scope:complete moves scopes to completed/ but leaves specification.xbrief.json roadmap item statuses stale ([proposed] / Later:)


MScottAdams:
## Summary

Consumer report (Doug G, Slack 2026-07-15): roadmap items inside the specification artifact (`specification.xbrief.json` / colloquially `spec.xbrief.json`) still read ``[proposed]`` / ``Later:`` after the corresponding scope files have all shipped and been moved to ``xbrief/completed/``. Lifecycle folder moves succeed; the **spec's own item statuses are never reconciled**.

## Observed behavior

1. Operator (or swarm completion sweep) runs ``task scope:complete`` for each shipped scope.
2. Scope files land in ``xbrief/completed/`` with ``plan.status: completed``.
3. Matching roadmap / registry items inside ``xbrief/specification.xbrief.json`` remain ``status: proposed`` (and still surface under Later / deferred phase-tier labels when rendered).
4. Agents and humans reading the spec as the planning surface see work as still proposed even though filesystem lifecycle truth says completed.

## Expected behavior

When lifecycle tooling completes a scope that is mirrored as an item (or subItem) in ``specification.xbrief.json``, that matching entry should either:

- be updated to ``status: completed`` (and drop stale Later / deferred presentation), **or**
- be intentionally left alone with a documented contract that **lifecycle folders are authoritative** and ``specification.xbrief.json`` ``plan.items[].status`` is not — plus a validator / doctor / render path that does not present the stale statuses as current roadmap truth.

Same reconcile obligation should apply for the other terminal transitions if the matching item is present (at least ``fail`` / ``cancel``), or those should be explicitly deferred with rationale.

## Prior art / gap vs #1527

- **#1527** (closed) fixed the sibling bug for ``PROJECT-DEFINITION``: ``syncProjectDefinitionAfterScopeMove`` in ``packages/core/src/scope/project-definition-sync.ts`` flips matching ``plan.items[].status`` and rewrites ``source_path`` / ``lifecycle_folder`` after a scope move.
- ``packages/core/src/scope/transition.ts`` calls that PROJECT-DEFINITION sync on every folder move — it does **not** sync ``specification.xbrief.json``.
- This issue is the **specification-artifact** half of the same class of drift; #1527 did not close it.

## Related (not duplicates)

[...]

Issue #1527, opened 2026-06-05
scope:complete updates PROJECT-DEFINITION references but leaves registry item status stale


polfer:
## Summary

In a vBRIEF-centric consumer project, `task scope:complete -- <active-scope-vbrief>` moves the scope file to `vbrief/completed/` and updates the `PROJECT-DEFINITION.vbrief.json` reference URI to point at `completed/...`, but the matching `plan.items[]` registry entry can remain at `status: proposed`.

## Observed behavior

After completing a scope with:

```bash
task scope:complete -- vbrief/active/2026-06-04-sites-membership-api.vbrief.json
```

The lifecycle move succeeded and `PROJECT-DEFINITION.vbrief.json` references now include:

```json
{
  "uri": "completed/2026-06-04-sites-membership-api.vbrief.json",
  "type": "x-vbrief/plan",
  "title": "Sites and Site Membership API"
}
```

But the corresponding registry item still reads:

```json
{
  "id": "sites-membership-api",
  "title": "Sites and Site Membership API",
  "status": "proposed"
}
```

The same consumer project also had previously completed scope references whose registry items still showed `proposed`, suggesting this is not isolated to one vBRIEF.

## Expected behavior

When lifecycle tooling completes a scope and updates `PROJECT-DEFINITION.vbrief.json`, the matching `plan.items[]` entry should either:

- be updated to `status: completed`, or
- be intentionally left alone with a documented validation rule explaining that `plan.items[].status` is not authoritative and may diverge from lifecycle folder/status.

## Impact

Agents and operators can see contradictory project state in the same file: references point at completed scope vBRIEFs, but the scope registry still appears proposed. This makes work selection and dependency reasoning more error-prone unless agents explicitly prefer lifecycle folders over `PROJECT-DEFINITION.plan.items`.

## Acceptance criteria

- `task scope:complete` has deterministic behavior for the matching PROJECT-DEFINITION registry item.
- Either the registry status is synchronized, or docs/validation make the intentional divergence explicit.
- `task vbrief:validate` or a related guard catches unintended stale registry statuses if synchronization is the intended contract.

## Notes

This report is based on consumer-side behavior observed in a vBRIEF v0.6 project using Deft v0.41.0. No project-specific implementation details are needed to reproduce beyond having a PROJECT-DEFINITION registry item matching the completed scope id.

PR #2873, opened 2026-07-27
fix(scope): reconcile own plan.items and envelope updated on complete (#2862)

### LINKED ISSUE #2862: bug(scope): scope:complete leaves the completing brief's own plan.items[].status pending and xBRIEFInfo.updated stale
Filed as a **consumer of the framework**, not a maintainer — apologies if any of this is already understood internally. Everything below was verified against the installed package on one project; I have not read the framework source repo.

## Summary

`deft scope:complete` moves a scope brief to `xbrief/completed/` and sets `plan.status = "completed"`, but leaves the brief's **own** `plan.items[].status` on `pending` and never refreshes the document-level `xBRIEFInfo.updated`. The result is a completed brief that internally still reads as unstarted work.

This is the same reconcile obligation you already closed twice for the *mirror* artifacts — #1527 for `PROJECT-DEFINITION` and #2566 / #2592 for `specification.xbrief.json`. Both of those flip `plan.items[].status` on a terminal transition. The one artifact left unreconciled is the completing brief itself, which is the authoritative record rather than a projection.

Our operators have hand-corrected this on three consecutive completion PRs and a bot flagged it on a fourth, which is why I'm filing rather than continuing to patch it locally.

## Environment

| | |
|---|---|
| CLI | `@deftai/directive@0.85.0` (engine `@deftai/directive-core@0.85.0`), global npm install |
| Vendored content | `.deft/core` at `v0.84.0` (`managed_by: npm`, deposited 2026-07-23) |
| `xbrief/.deft-version` | `0.84.0` |
| Layout | `xbrief/` (v0.8), consumer install (not a framework source checkout) |
| OS / shell | macOS (darwin 25.5.0), zsh |

## Observed behavior

1. A brief in `xbrief/active/` has `plan.status: "running"` and `plan.items[*].status: "pending"`.
2. The work ships and the PR merges.
3. Operator runs `deft scope:complete -- xbrief/active/<brief>.xbrief.json`.
4. The file moves to `xbrief/completed/` with `plan.status: "completed"`, `plan.updated` refreshed, and `plan.metadata.completedAt` stamped.
5. **`plan.items[*].status` is still `"pending"`** on every item, including items that fully shipped.
6. **`xBRIEFInfo.updated`** still holds the timestamp from before the completion.

## Expected behavior

On a terminal transition, either:

- set the brief's own `plan.items[*].status` to the terminal status (with whatever cascade rule you prefer for `subItems`) and refresh `xBRIEFInfo.updated` alongside `plan.updated`; **or**
- document that `plan.items[].status` on a completed brief is not authoritative, and stop having deterministic reviewers flag it (see the SLizard rule below, which currently treats it as a logic error).

Either resolution is fine from a consumer's perspective. The current state is the one that costs time: the field looks authoritative, a bot says it's wrong, and the tooling won't fix it.

## Reproduction

```bash
deft scope:activate -- xbrief/proposed/<brief>.xbrief.json   # -> active/, status running, items pending
# ... ship the work ...
deft scope:complete -- xbrief/active/<brief>.xbrief.json     # -> completed/, status completed
jq '.plan.status, [.plan.items[].status], .xBRIEFInfo.updated' xbrief/completed/<brief>.xbrief.json
# "completed"
# ["pending","pending","pending","pending"]
# "<pre-completion timestamp>"
```

## Evidence from the installed code

`@deftai/directive-core@0.85.0`, `dist/scope/transition.js` (built from `packages/core/src/scope/transition.ts`), the mutation block of `runTransition`:

```js
const nowIso = utcNowIso(now);
planObj.status = targetStatus;
planObj.updated = nowIso;
if (act === "complete") {
    stampCompletionMetadata(planObj, projectRoot, nowIso);
}
```

That is the complete set of in-document mutations. `stampCompletionMetadata` (`dist/scope/capacity-stamp.js`) only writes `plan.metadata.completedAt` and `plan.metadata.capacityBucket`. Nothing in the function reads or writes `planObj.items`, and nothing touches the top-level `xBRIEFInfo` / `vBRIEFInfo` envelope — `planObj.updated` is `plan.updated`, a different field.

**The asymmetry is visible in the same function.** A few lines later `runTransition` calls:

```js
syncProjectDefinitionAfterScopeMove(data, resolvedPath, destPath, vbriefRoot, targetStatus);
syncSpecificationAfterScopeMove(data, resolvedPath, destPath, vbriefRoot, targetStatus);
```

Both route into `dist/scope/registry-artifact-sync.js`, which does exactly the thing the brief itself doesn't get:

```js
if (i.status !== targetStatus) {
    i.status = targetStatus;
}
```

So on a completion the framework reaches out and reconciles item status in two derived registries while leaving the source artifact stale. From the outside that reads like an oversight rather than a decision, but I may be missing the rationale.

## Occurrences in one consumer repo (deftai/BestiMax)

Three PRs whose entire diff is the hand-correction, plus a bot finding:

| PR | Brief | Manual fix-up in the diff |
|---|---|---|
| [#119](https://github.com/deftai/BestiMax/pull/119) | `2026-07-22-land-dependabot-patch-minor` | `items[].status: pending -> completed`, `xBRIEFInfo.updated` bumped |
| [#121](https://github.com/deftai/BestiMax/pull/121) | `2026-07-22-land-eslint-10` | same two corrections |
| [#138](https://github.com/deftai/BestiMax/pull/138) | `2026-07-26-rls-boundary-hardening` | same two corrections, on four items |

PR #138's commit message names the gap directly:

> `deft scope:complete` flips plan.status and stamps metadata.completedAt but leaves item statuses and xBRIEFInfo.updated behind, so those are corrected here.

**A deterministic reviewer already treats this as a defect.** SLizard's review on [BestiMax PR #127](https://github.com/deftai/BestiMax/pull/127) raised rule `directive-2584-recall`:

> **directive-2584-recall**: [Logic Error] Active xBRIEF plan.status is still "running"/"active" and all plan.items are pending/proposed even though this PR implements the work.

Worth being precise about what that finding proves, since it is adjacent rather than identical: it fired on the *implementation* PR, where the brief was still active, so it is about the implementing agent not updating item statuses. Combined with this report the picture is that **no surface in the toolchain ever transitions a brief's own `items[].status`** — not the implementer, not the lifecycle command. An operator following the documented flow exactly still ends up with a stale corpus and a bot finding.

Three repetitions of an identical manual correction, plus an automated reviewer flagging the same field, is what pushed this from "our habit is wrong" to "the tooling has a gap".

## Impact

Low severity, steady cost. Nothing breaks and no gate fails — `deft check` and `vbrief:validate` both pass on a brief with completed status and pending items, because the fields are schema-independent.

What it costs:

- **Every completion needs a hand-edit**, or the corpus accumulates briefs that are internally contradictory.
- **Reviewers flag it**, so each completion PR carries a finding the operator must dismiss or fix.
- **Item-level status becomes unusable as a data source** — you can't answer "which planned items actually shipped" across the corpus, because the answer is "all of them say pending".
- **It is invisible in the moment.** `scope:complete` exits 0 with a success message; the staleness is only found later, by a reviewer or a reader.

## Prior art in this repo

- **#1527** (closed) — *"scope:complete updates PROJECT-DEFINITION references but leaves registry item status stale."* Same defect, mirror artifact, fixed.
- **#2566 / #2592** (closed) — *"scope:complete moves scopes to completed/ but leaves specification.xbrief.json roadmap item statuses stale."* Same defect, second mirror artifact, fixed. #2566's framing applies verbatim here: *"Lifecycle folder moves succeed; the item statuses are never reconciled."*
- **#1674** (open) — *"task scope:cancel doesn't stamp cancellation metadata (cancelledAt / cancellationReason / item status / Disposition)."* This is the closest existing issue and it already names the exact gap, including the asymmetry:

  > `plan.items[*].status` — left as `proposed` (etc.) instead of `cancelled` when the parent is cancelled. (The existing PROJECT-DEFINITION sync updates *that* plan's items, but not the scope vBRIEF's own `plan.items[]`.)

  But #1674 is scoped to `cancel`; its proposed-fix item 4 asks whether the same symmetry should extend to `fail`, and does not mention `complete`. **If you'd rather widen #1674 to cover all terminal transitions than carry a second issue, that seems cleaner and I'm happy for this one to be closed as folded in.** I filed separately because `complete` is by far the highest-traffic verb and #1674 has sat at P3 since June with no comments.
- **#1618** (open RFC) — *"Separate source-of-truth from materialized views — demote plan.items."* If `plan.items` is heading for demotion, that changes the right fix here (documenting non-authority rather than reconciling), which is the other reason I'd rather surface this than keep patching locally.

## Suggested fix

Mirroring the shape of #1527 / #2592, inside `runTransition`:

1. On terminal transitions (`complete`, and by symmetry `fail` / `cancel` per #1674), set the brief's own `plan.items[*].status` to the terminal status, recursing into `subItems`.
2. Decide and document the cascade rule for items that legitimately did **not** ship. A blanket flip to `completed` would overwrite a deliberate `cancelled` or `deferred` item. The conservative rule — only advance items still in a non-terminal state (`pending` / `proposed` / `running`), leave already-terminal items alone — preserves operator intent and is what our hand-edits have been doing by eye.
3. Refresh `xBRIEFInfo.updated` (v0.8) / `vBRIEFInfo.updated` (v0.6) in the same write as `plan.updated`, matching whichever envelope the document actually carries. #2346 / #2363 is the cautionary precedent here: a writer that stamped `vBRIEFInfo` unconditionally corrupted v0.8 files and blocked a release.
4. If instead the intended contract is that a completed brief's item statuses are *not* authoritative, document that and retire the `directive-2584-recall` reviewer rule — the two currently contradict each other.

## Two adjacent observations, recorded for context only

Neither is part of this report; happy to file either separately if useful.

**(a) `plan.completedNote` is rejected by the conformance gate, and the canonical replacement isn't obvious from the error.** The two briefs completed by the PRs in the table above (BestiMax #119 and #121) use bare `plan.completedNote`, which the #1620 gate now flags: *"detected 4 bare key(s) ... Every vBRIEF key MUST be spec-core, x-directive/-namespaced, x-vbrief/-namespaced, or x-xbrief/-namespaced."* Reading `dist/policy/plan-extensions.js` makes clear this is **intended** tightening under #1650, and that the canonical target is `x-directive/completedNote` with `dist/category-b-namespace/index.js` as the codemod. So: not a bug, and I'm not filing it. The friction is only that the gate's remediation text lists three namespace prefixes generically without naming `x-directive/completedNote` as *the* answer for this specific key — our operator guessed `plan.metadata.completedNote`, which is conformant but off-canon. A one-line hint in the message would have prevented that.

**(b) The conformance gate appears not to run on consumer installs.** On a consumer project, `deft verify:vbrief-conformance` with no arguments prints `OK: vBRIEF validation passed` and exits 0, while `deft verify:vbrief-conformance --all` on the identical tree exits non-zero with 4 findings. Looking at `dist/vbrief-validate/main.js`, `cmdVbriefValidate` only routes to `runConformance` when `argv[0] === "conformance"`, or when `--staged` / `--all` is present — a bare invocation falls through to the schema validator (its `--help` even prints `usage: vbrief_validate.py ...`). Separately, `dist/check/gate-lists.js` lists `verify:vbrief-conformance` in `FRAMEWORK_CHECK_GATES` but not in `CONSUMER_CHECK_GATES`. Net effect on our install: the only path that actually runs the #1620 scan is the pre-commit hook's `--staged` mode, which sees newly staged files only — so four pre-existing violations have sat in the tree unreported. This may well be deliberate scoping; flagging in case the bare-verb fallthrough is not.

## Also from this install

The same project filed #2826 on 2026-07-24 (empty open-issue inventory treated as stale, blocking write hooks) — since closed by #2830, thank you. Noting it so you can see these come from the same consumer install and framework line, in case that's useful for correlating reports.

MScottAdams:
## Summary

`scope:complete` (and terminal `fail` / `cancel`) set `plan.status` and `plan.updated` but left the completing brief's own `plan.items[].status` on pending and never refreshed the envelope `updated` stamp. Mirror registry sync already fixed PROJECT-DEFINITION (#1527) and specification (#2566); the source brief itself was the remaining gap.

## Changes

- On terminal transitions in `runTransition`, advance non-terminal own `plan.items` / nested `subItems` and `items` to the terminal status when still `pending` / `proposed` / `running`
- Preserve already-terminal item statuses (`cancelled` / `failed` / `completed` / `blocked`, etc.)
- Stamp `xBRIEFInfo.updated` (v0.8) or `vBRIEFInfo.updated` (v0.6) in the same write as `plan.updated` without creating the other envelope key
- Unit tests for mixed statuses, envelope stamp (v0.8 + v0.6), fail/cancel symmetry, and edge cases
- CHANGELOG [Unreleased] cites #2862

## Test plan

- [x] `pnpm exec vitest run packages/core/src/scope/transition.test.ts`
- [x] `pnpm exec vitest run packages/core/src/scope/`
- [x] `task check:merge`

Closes #2862

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`scope:complete` reconciles the completing brief's own `plan.items` and envelope `updated` (#2862).** Terminal transitions (`complete` / `fail` / `cancel`) now advance non-terminal own items (and nested `subItems`) to the terminal status while preserving already-terminal items (`cancelled` / `failed` / `completed`), and stamp `xBRIEFInfo.updated` or `vBRIEFInfo.updated` alongside `plan.updated`. Closes #2862.
+
 ### Removed
 
 ## [0.86.0] - 2026-07-27
--- a/packages/core/src/scope/transition.test.ts
+++ b/packages/core/src/scope/transition.test.ts
@@ -89,6 +89,203 @@ describe("runTransition", () => {
     expect(data.plan.metadata.completedAt).toMatch(/Z$/);
   });
 
+  it("advances non-terminal own plan.items and stamps xBRIEFInfo.updated on complete (#2862)", () => {
+    root = makeRepo();
+    const path = join(root, "xbrief", "active", "mixed-items.xbrief.json");
+    const staleEnvelope = "2026-01-01T00:00:00Z";
+    writeFile(path, {
+      xBRIEFInfo: { version: "0.8", updated: staleEnvelope },
+      plan: {
+        title: "mixed",
+        status: "running",
+        updated: staleEnvelope,
+        items: [
+          { title: "pending-item", status: "pending" },
+          { title: "proposed-item", status: "proposed" },
+          { title: "running-item", status: "running" },
+          { title: "cancelled-item", status: "cancelled" },
+          { title: "failed-item", status: "failed" },
+          { title: "already-completed", status: "completed" },
+          {
+            title: "parent-with-sub",
+            status: "pending",
+            subItems: [
+              { title: "sub-pending", status: "pending" },
+              { title: "sub-cancelled", status: "cancelled" },
+            ],
+          },
+        ],
+      },
+    });
+    const fixed = new Date("2026-07-27T15:30:00.000Z");
+    const result = runTransition("complete", path, fixed);
+    expect(result.ok).toBe(true);
+    const dest = join(root, "xbrief", "completed", "mixed-items.xbrief.json");
+    const data = JSON.parse(readFileSync(dest, "utf8")) as {
+      xBRIEFInfo: { updated: string };
+      plan: {
+        status: string;
+        updated: string;
+        items: Array<{
+          title: string;
+          status: string;
+          subItems?: Array<{ title: string; status: string }>;
+        }>;
+      };
+    };
+    expect(data.plan.status).toBe("completed");
+    expect(data.plan.updated).toBe("2026-07-27T15:30:00Z");
+    expect(data.xBRIEFInfo.updated).toBe(data.plan.updated);
+    expect(data.xBRIEFInfo.updated).not.toBe(staleEnvelope);
+    const byTitle = Object.fromEntries(data.plan.items.map((i) => [i.title, i]));
+    expect(byTitle["pending-item"].status).toBe("completed");
+    expect(byTitle["proposed-item"].status).toBe("completed");
+    expect(byTitle["running-item"].status).toBe("completed");
+    expect(byTitle["cancelled-item"].status).toBe("cancelled");
+    expect(byTitle["failed-item"].status).toBe("failed");
+    expect(byTitle["already-completed"].status).toBe("completed");
+    expect(byTitle["parent-with-sub"].status).toBe("completed");
+    expect(byTitle["parent-with-sub"].subItems?.[0].status).toBe("completed");
+    expect(byTitle["parent-with-sub"].subItems?.[1].status).toBe("cancelled");
+  });
+
+  it("stamps vBRIEFInfo.updated on terminal transition for v0.6 envelopes (#2862)", () => {
+    root = makeRepo();
+    const path = join(root, "xbrief", "active", "v06.xbrief.json");
+    writeFile(path, {
+      vBRIEFInfo: { version: "0.6", updated: "2026-01-01T00:00:00Z" },
+      plan: {
+        title: "v06",
+        status: "running",
+        items: [{ title: "a", status: "pending" }],
+      },
+    });
+    const fixed = new Date("2026-07-27T16:00:00.000Z");
+    expect(runTransition("complete", path, fixed).ok).toBe(true);
+    const dest = join(root, "xbrief", "completed", "v06.xbrief.json");
+    const data = JSON.parse(readFileSync(dest, "utf8")) as {
+      vBRIEFInfo: { updated: string };
+      plan: { updated: string; items: Array<{ status: string }> };
+    };
+    expect(data.vBRIEFInfo.updated).toBe("2026-07-27T16:00:00Z");
+    expect(data.vBRIEFInfo.updated).toBe(data.plan.updated);
+    expect(data.plan.items[0].status).toBe("completed");
+  });
+
+  it("advances non-terminal own items on fail and cancel (#2862)", () => {
+    root = makeRepo();
+    const failPath = join(root, "xbrief", "active", "fail-items.xbrief.json");
+    writeFile(failPath, {
+      xBRIEFInfo: { version: "0.8", updated: "2026-01-01T00:00:00Z" },
+      plan: {
+        title: "fail",
+        status: "running",
+        items: [
+          { title: "p", status: "pending" },
+          { title: "c", status: "cancelled" },
+        ],
+      },
+    });
+    expect(runTransition("fail", failPath).ok).toBe(true);
+    const failed = JSON.parse(
+      readFileSync(join(root, "xbrief", "completed", "fail-items.xbrief.json"), "utf8"),
+    ) as { plan: { status: string; items: Array<{ status: string }> } };
+    expect(failed.plan.status).toBe("failed");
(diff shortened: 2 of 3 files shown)


MScottAdams:
ended_at: 2026-07-27T22:54:17.965Z

Issue #3081, opened 2026-08-03
epic(evals,swarm): claim honesty & framework empiricism — metric-first, benchmarks, verify-the-verifier


MScottAdams:
## TL;DR

Directive needs **honest metrics and independent verification** so optimizations and go/no-go claims can be falsified — not only "looks better."

This epic packages claim-honesty work that is **related to** the agent cost stack (#3078) but **must not block** that ladder:

1. **Metric-first design** (and GEPA-class optimization when a metric exists) — #492
2. **Framework empiricism** — with_skill vs without_skill benchmarks — #1584
3. **Methodology audit on decision-grade analysis** — #2477
4. **Verification independence** (corrective vs self-coupled loops) — #1499

**ROI headline:** fewer cargo-cult "wins," fewer single-agent analytical ship decisions, and a path to evidence that Directive guidance changes agent behavior. Wave structure separates **principle docs** from **harness infrastructure** so the epic can land value without waiting on a full eval product.

**Out of this epic:** agent cost decision ladder (#3078: overhead → form → tier); local/CPU serving (#3079). Self-killing eval *note* on #3065 stays Wave A of #3078 and cites #492.

---

## Why a separate epic (not under #3078)

| Concern | Cost stack #3078 | This epic |
|---|---|---|
| Primary question | What to fix before uptiering a model? | How do we know a claim/optimization is real? |
| Typical delivery | Guidance in `context.md` / patterns | Guidance + eval harness + swarm process |
| Size | Days–weeks, docs-first | Weeks–months if #1584 is in scope |
| Cost direction | Reduce tokens / raise binding accuracy | Sometimes **spend** a second agent (#2477) for honesty |

Putting #1584 or #2477 on #3078 exit would make the cost ladder epic never close. Comment-only links were insufficient as a program home — this tracker owns the list and waves.

---

## Rationale / package ROI (directional)

| Lever | Child | Expected effect |
|---|---|---|
| Metric before prompt | #492 | Optimizations become measurable; preferred hacks can fail |
| Self-killing exam hygiene | #492 + note on #3065 | Metrics that cannot reject the author's fix are theater |
| with_skill vs without_skill | #1584 | Empirical evidence Directive changes behavior (ADR-001 path) |
| Methodology second agent | #2477 | Decision-grade analysis not accepted from one confident report |
| Verification independence | #1499 | Corrective loops vs self-coupled "verify" that rubber-stamps |

[...]

Issue #1584, opened 2026-06-11
feat(evals): evals/shared-benchmark.json — directive framework benchmark manifest with with_skill vs without_skill paired cases


visionik:
feat(evals): evals/shared-benchmark.json — directive framework benchmark manifest using skill-eval-harness

## Summary

Directive has no empirical evidence that it improves agent behavior. The ADR-001 empirical validation vBRIEF (2026-05-01-742-adr-001-empirical-validation) references deft-agent-bench as planned benchmark infrastructure, but no benchmark has shipped. This issue tracks creating a concrete benchmark manifest at evals/shared-benchmark.json that tests directive's core behavioral rules using skill-eval-harness (https://github.com/adewale/skill-eval-harness).

## The harness fit

skill-eval-harness is runner-agnostic: it prepares tasks as JSONL, expects outputs in a standard directory layout, then grades and aggregates. The with_skill vs without_skill paired variant maps directly to the directive evaluation question:

- without_skill: agent receives the task prompt with no AGENTS.md in the working directory
- with_skill: agent runs in a directive-configured directory where AGENTS.md loads main.md and the relevant skills

The harness supports: repeated-run statistics, holdout/holdback anti-overfitting splits, objective assertions (contains, regex, file_exists, script), qualitative judge handoff, ablation variants, and Anthropic-compatible grading.json exports.

## What the benchmark manifest should cover

Core behavioral rules to test with the with_skill vs without_skill paired comparison:

1. Implementation intent gate (#810) -- does directive prevent the agent from writing code before a spec/vBRIEF exists?
   Assertion: without_skill agent starts coding; with_skill agent asks for a plan first.

2. Branch policy -- does directive prevent the agent from committing directly to main?
   Assertion: with_skill output contains branch creation; without_skill output may not.

3. task check gate -- does directive ensure the agent runs task check before committing?
   Script assertion: check whether the agent invoked task check in its output.

4. Anomaly-first summaries -- does the anomaly-first summary rule in main.md change agent PR descriptions?
   Judge assertion with rubric: does the summary lead with risks/anomalies before good news?

5. vBRIEF lifecycle compliance -- does directive keep scope vBRIEFs in the correct lifecycle folder?
   File assertion: vbrief/active/ contains the expected file after activation.

## Required adapter

[...]

Issue #3218, opened 2026-08-09
ci(smoke): greenfield-python-free-smoke red on master — verify:consumer-check-contract composition after #3145


MScottAdams:
## Summary

**`greenfield-python-free-smoke` is red on `master` and on unrelated docs/feature PRs.** Failure is **not** story-specific product content; it blocks confidence in release/smoke and forced admin-merge on otherwise-green docs PRs (#3212, #3216, #3217 class).

## Failure signature

Workflow: **`greenfield-python-free-smoke`**

Flow: pack install → `directive init` (consumer deposit) → consumer `task deft:check` / engine check path.

Dies on:

```text
verify:consumer-check-contract [ci-workflow]
  CI workflows do not invoke verify:consumer-check-contract or a full check entrypoint that composes it
  remediation: Add verify:consumer-check-contract to the consumer CI workflow composition
    (Taskfile check deps and/or CI workflow invoking deft check / task verify:consumer-check-contract).
    See content/docs/consumer-check-contract.md (#3145).

task: Failed to run task "verify:consumer-check-contract": exit status 1
check: fast gate verify:consumer-check-contract failed (exit 201); skipping remaining gates (#3188)
```

Exit **201** on the fast gate → entire smoke job red.

## Evidence (recent)

| Run context | Example |
|-------------|---------|
| `master` | https://github.com/deftai/directive/actions/runs/31332772403 (`greenfield-python-free-smoke` **failure**) |
| `master` | https://github.com/deftai/directive/actions/runs/31332413576 |
| Docs PR branch (goal-gate #852) | https://github.com/deftai/directive/actions/runs/31332495527 |

Also observed on:

- #3212 decision-log dispose (docs-only; admin merge after smoke red)
- #3216 tool-surface grammar (#3085)
- #3217 goal-gate determinism (#852)

On those PRs, **TypeScript / merge-gate / Greptile** could be green while smoke stayed red → operators treat smoke as non-required for branch protection but still **fleet signal noise**.

## Suspected root cause

After **#3145** consumer-check-contract / scope-provenance work, **greenfield consumer scaffolding** (Taskfile / CI deposit from `directive init`) does **not** compose `verify:consumer-check-contract` the way the gate expects when smoke runs `deft:check`.

So either:

1. **Fixture / deposit bug** — greenfield PROJECT-DEFINITION / generated Taskfile / CI snippets omit the contract composition, **or**
[...]

Issue #3095, opened 2026-08-03
feat(policy,review): dogfood Greptile CLEAN requires 5/5; consumers get per-project min-confidence lever


MScottAdams:
## Summary

Tighten the **maintainer (directive dogfood) Greptile CLEAN bar** from the current **confidence > 3** (i.e. **4/5 or 5/5** + no P0/P1 + SHA match + terminal check) to **confidence == 5/5 required**, while giving **consumers a per-project policy lever** to raise (or keep) the confidence floor as they need.

Today agents and gates treat **4/5 with 0 P0/P1** as merge-ready. That is too loose for **framework maintenance** on `deftai/directive`, where residual 4/5 holdouts often still name real contract gaps (as seen on recent review cycles chasing 5/5). Consumers may want the same bar, a looser default, or a project-specific floor — that should be **typed policy**, not hard-coded only in prose.

## Current behavior (SoT pointers)

Clean gate is effectively:

- Greptile SHA match on HEAD
- terminal `Greptile Review` check-run
- **confidence > 3** (4/5 or 5/5)
- **P0 = 0 and P1 = 0**
- CI green (where `pr:watch` / merge-ready enforce it)

Documented / implemented in (non-exhaustive):

- `content/skills/deft-directive-review-cycle` Step 6 — confidence > 3
- `content/templates/swarm-greptile-poller-prompt.md` — clean threshold `confidence > 3`
- `content/tools/greptile.md` — exit condition must be > 3
- `packages/core/src/pr-watch/constants.ts` — CLEAN requires confidence > 3
- `packages/core/src/pr-merge-readiness/evaluate.ts` — same

### A) Directive maintenance (this repo / dogfood)

- ! CLEAN / merge-ready for **framework source** requires **Greptile Confidence Score: 5/5** (not merely > 3).
- ! Still require **0 P0 and 0 P1**, SHA match, terminal check-run, and existing CI merge-ready rules.
- ! Agents MUST continue fix batches (or explicit blocked handback) while confidence is 4/5 or lower even if findings list is empty of P0/P1.
- ⊗ Exit review-cycle / `pr:watch` CLEAN / swarm Phase 5→6 on **4/5** for directive dogfood.

### B) Consumers — per-project lever

- ! Typed policy (e.g. under `plan.policy` on `PROJECT-DEFINITION` / xBRIEF) such as:
  - `plan.policy.review.minGreptileConfidence` (integer 1–5), **or**
[...]

PR #3220, opened 2026-08-09
feat(evals): pin framework version per eval cell — no mixed-version runs (#3215)

### LINKED ISSUE #3215: feat(evals,release): pin and record framework version per eval cell - no mixed-version runs (extends #3081)
**Evidence:** an internal benchmark run (2026-08-09; private section-31 note) recorded directive's framework version as mixed within a single eval cell - different directive versions ran inside one treatment. The benchmark's own provenance display caught it; directive's eval discipline should have prevented it.

## Problem

A score attributed to directive is meaningless if the cell mixes framework versions. Every comparative claim the empiricism track wants to make (extends #3081, manifest home #1584) depends on version-pure cells, and nothing today enforces or even records purity.

## Proposal

Eval harness integration records the resolved framework version (directive doctor output, or the package.json pin) in per-run metadata at run start. A cell-level check flags or refuses aggregation when runs within one treatment disagree on version. Wire the recorded version into the #1584 shared-benchmark manifest shape so any external benchmark can consume it.

Small scope, high leverage: it protects the credibility of every future empirical claim the framework makes about itself.

MScottAdams:
## Summary

Eval cells must be version-pure so scores attributed to directive are not mixed-version noise (extends #3081; manifest home #1584).

- Record resolved framework version pin at `eval:run` start (package.json or override) on each golden-run ledger row
- Cell-level purity: default **refuse** aggregation when runs in one treatment disagree; optional **flag** policy
- `eval:report` surfaces version purity evidence; refuses when a version cell mixes pins
- Wire pin into `#1584` `evals/shared-benchmark.json` shape when that manifest is present
- Module: `packages/core/src/eval/version-pin.ts` (+ tests)

## Test plan

- [x] `npx vitest run packages/core/src/eval/version-pin.test.ts packages/core/src/eval/run.test.ts packages/core/src/eval/report.test.ts`
- [x] `task verify:scope-provenance` clean
- [x] `task ts:check-lane` (full suite + coverage) green
- [ ] CI green on PR
- [ ] Greptile clean

Closes #3215


MScottAdams:
ended_at: 2026-08-09T21:25:00.898Z

MScottAdams:
Note: Greptile residual was confidence 4/5 with P0=0 P1=0 after fix batches (persist pin + ledger-before-pin). Latest push adds explicit no-dangling-pin tests. Framework dogfood minGreptileConfidence=5 (#3095). Sibling greenfield smoke is #3218 scope.

Issue #3282, opened 2026-08-11
bug(check,doctor): done-gate toolchain preflight - directive check burns the turn budget bootstrapping go-task/pnpm/CLI dist in consumer containers, and gate failures do not name their cause


ape-deft:
**TL;DR: this is primarily a TELEMETRY request, gated on an env var.** When `DEFT_RUN_SUMMARY_PATH` is set, directive appends a JSONL run summary to that path - dial selection at session start, every dial transition as it fires, every check/verify invocation with named causes - written event-by-event so it survives runs that die mid-session. stdout is touched only on failure of an explicitly requested write (one warning line) or when the var is set to `-`. Unset = silent. The secondary deliverable is the preflight/diagnosability fix: toolchain preflight at session start, gate failures that name cause + remedy, and a degraded-mode skip report, so agents stop burning ~45% of turns reverse-engineering `directive check` in toolchain-incomplete containers.

**SPEC v2 - consolidates the comment thread below; read this body as authoritative, comments as history.**

**Evidence (internal benchmark transcripts, private - no benchmark name or scores on public surfaces):** in two agent runs on a non-CC harness (directive 0.99.0, fresh consumer containers), 44-47% of ALL agent turns went to making `directive check` pass rather than to the task:

- `deft verify:branch` exits 1 with no diagnostic; both agents independently read orchestrator source inside node_modules to learn why.
- - The `task` (go-task) binary was absent from the consumer container - the check runner simply was not there.
- - - `toolchain:check-consumer` required pnpm (absent; corepack bootstrap) and a built CLI dist (missing; the engine attempted self-build).
- - - - After bootstrapping, check exited 0 - catching zero product defects. Both runs then failed their official verifier on product issues the gate never examined, with final product verification squeezed into the last 1-2 turns.
Likely multiplier: #3225 (advisory-verdict blocking) made previously-advisory failures mandatory. Prior art all closed (#2410 pnpm hardcode, #1659 runner decoupling, #1519 consumer self-tests, #1648 unprefixed task names) - so regression or coverage gap on the toolchain:check-consumer path.

## Fix

**(1) Preflight at session start, not done time.** The session ritual (or doctor) verifies the done-gate's own toolchain (task binary, package manager, CLI dist) and either remedies or declares degraded mode - one turn, up front. Never bootstrap tooling the PRODUCT does not need in order to satisfy the framework's own gate.

[...]

Issue #3225, opened 2026-08-09
bug(review-cycle,swarm): advisory-only reviewer verdicts are invisible to mechanical merge-readiness - babysit and merge gates must parse verdict text (extends #3095)


ape-deft:
**Live case (2026-08-09, private org docs PR):** a reviewer bot's summary stated a should-not-merge verdict (confidence 3/5) in its comment body, while never setting the formal Changes-Requested review state. GitHub's merge box showed Ready to merge the entire time; all checks were green. A shepherding agent following mechanical signals alone would have merged against explicit reviewer advice. It did not - but only because the babysit brief happened to require reading bot prose. The safety came from agent discipline, not from any gate.

## Problem

Directive's merge-readiness discipline treats formal review state plus checks as the clean signal. But reviewer bots on repos without enforced review wiring (no blocking CheckRun, no request_changes) express verdicts only in comment text and confidence scores. Mechanical mergeability is therefore an insufficient clean-signal on such repos. #3095 solved this for the directive repo itself (Greptile CLEAN requires 5/5, wired into policy and CI), but the portable review-cycle / babysit path has no equivalent - an agent shepherding a PR on any other org repo sees green and merges.

## Proposal

The review-cycle merge-ready check MUST parse advisory verdicts, not just formal state: extract reviewer-bot confidence and verdict language from comment bodies (the #1282 body-extraction helper is the precedent; #1101 already reads a CheckRun as a secondary clean signal) and treat a should-not-merge verdict or sub-threshold confidence as blocking regardless of the mechanical Ready-to-merge state. Add the explicit rule to the babysit brief and skill: mechanical mergeability is necessary, never sufficient. Recommend repo-level review-gate wiring for org repos that lack it, so the formal state and the prose verdict cannot diverge silently.

## Refs

Extends #3095 (per-repo min-confidence lever). Composes #1282 (verdict body extraction) and #1101 (CheckRun secondary signal). Related: #2308 (review-surface precedence), #1590 (decision-ready boundary and owner brief).

Issue #3214, opened 2026-08-09
feat(policy,session): ceremony dial - scale ritual depth by task size x model tier x project shape


ape-deft:
**Evidence:** internal benchmark evidence (section-31 strategy note, 2026-08-09, private - no benchmark name or scores on public surfaces). Findings: directive's fixed process overhead (session ritual, lifecycle, gates, change proposals) costs roughly 3-7x turns and dollars-per-point on small tasks with zero quality gain on strong models, and measurably reduces correctness on frontier-model short-horizon non-project tasks. On mid-tier stacks the same structure raises scores (best-of-frameworks there). Directive is also the only framework with identical scores across harnesses - the portability investment shows up as data.

## Problem

Ceremony is constant while task size, model tier, and task shape vary. A frontier model on a small task pays the full ritual and gets nothing back; a mid-tier model on a terminal task benefits. Nothing in directive today selects ritual depth deterministically.

## Proposal

A typed policy field (Rule Authority: deterministic over prose): ritualDepth = f(effort estimate, model tier, project-shaped detection, host capability). Composes existing tracked pieces - effort estimate schema in #1581 gives the task-size input, effort-based model routing in #1976 and per-call selection in #818 give the tier input, the capability descriptor in #1461 gives the session-start surface, strategies/rapid.md already exists as the light path, and the minimal AGENTS profile proposed in #3014 covers the non-project case. What is genuinely new is only the deterministic selection policy that composes them.

## Acceptance sketch

plan.policy.ceremonyDial (or sibling) as a typed field; the default maps S-tasks on frontier models to the rapid path. Non-project context detection degrades to the #3014 minimal profile automatically. Session ritual, gate depth, and lifecycle writes scale with the selected depth. Override always available; policy changes audited like other policy flags.

Composes #1581 #1976 #818 #1461 #3014 - not a new subsystem.

MScottAdams:
## Design note for the implementation cohort (#3214 + #1581 + #3215): the effort-estimate input has an ordering constraint

#1581's estimates live on scope vBRIEF plan items — they exist **after planning**. But the ceremony this dial wants to skip (session ritual, deposit loading, gate setup) fires **before planning**. A naive wiring — "read effort estimate, select ritual depth" — **deadlocks at session start** with no estimate to read.

### Two resolutions, composable

[...]

Issue #3156, opened 2026-08-06
feat(security,verification): gate-integrity rule - a failing gate must not be fixed by editing the gate (extends #3145/#782)


MScottAdams:
**Parent epic:** #3179 (wave 1) — self-improving under gates (propose-not-apply).  
**Extends:** #782 (fixed evaluator outside edit surface), #1499 (verification independence), #2436 (SkillOpt control stack).  
**Related:** #2741 (memory/learning), #3164 (stance — **shipped** PR #3202), #1200 (safety via formal gates).  
**Source:** section-31 `strategy/2026-08-06-prime-agent.md`; Prime Agent / Continual Harness writeups.

## Field evidence

Prime Agent's `/refine` loop packaged cheating as a reusable skill: in Factorio it found a resource-spawn exploit, and the same self-improvement loop that codifies good tactics codified the exploit. Framing: "self-improvement has no moral compass, only reward."

This is the concrete failure case #782 / #1499 / #2436 argue against — observed in a shipped MIT harness, not only hypothesized.

## Scope after verification (important)

**Refine-loop-internal** protection (proposer cannot edit its own reward / slow-update / validator region) is already substantially specified on **#2436**. Do **not** re-implement that stack here.

This issue owns the **general gate-integrity rule** for Directive product and process gates:

- ⊗ Fix a **failing gate** by editing the **gate definition** (or weakening the check) to go green.
- ! Fix the **product, process, or test under test** — or deliberately change the gate via the normal propose/PR path with explicit rationale.

Extends #3145 / verification-independence themes: the evaluator lives outside the editable surface under test.

## Proposed change

1. Document one sharp rule (e.g. `meta/security.md` and/or verification / pre-pr docs):  
   - ⊗ A self-modification, refine, or fix loop MUST NOT clear red by mutating the gate, verifier, reward, or required check that is failing.  
   - ! Gate definition changes go through issue/PR + review (same as constitution-tier under #3164).  
2. Cite Factorio / Continual Harness as motivating case study (pointer OK if #2436 or #782 already holds detail).  
3. Where practical: content contract or pre-pr checklist item so the rule is discoverable (full CI automation optional follow-up).

## Explicit non-goals

- Implementing full SkillOpt (#2436) under this issue  
- Host hook enforcement for self-mutating hosts (#3162)  
- Replacing #1200 design-principle docs (complementary)

## Acceptance criteria

- [ ] Gate-integrity rule documented in canon (security / verification / main pointer as appropriate)
[...]

PR #3326, opened 2026-08-13
docs(analysis): name the 0.92-0.99 hard-task regression cause (#3321)

### LINKED ISSUE #3321: bug(regression): two hard-task regressions introduced at 0.99 survive 0.101 - field-solved and directive-solved at 0.92, and the done-gate turn-sink explanation is now disconfirmed
**TL;DR: two hard tasks that directive solved at 0.92 broke at 0.99 and are still broken at 0.101. Both are solved by every comparison arm on the same model, including no-framework. The done-gate turn-sink explanation we landed on in the 0.99 forensics is now disconfirmed — 0.101 cut gate invocations by roughly an order of magnitude and both tasks still fail. Something else in the 0.92 -> 0.99 delta is responsible and is still shipping. Ask is a bisect.**

**Evidence (internal benchmark, private; one mid-tier stack, two hard non-project tasks, n=1 per task per version but four observations across two versions):**

| | 0.87 | 0.92 | 0.99 | 0.101 |
|---|---|---|---|---|
| task A | — | pass | **fail** | **fail** |
| task B | pass | pass | **fail** | **fail** |

Base rates on the same model and bank, which is what makes this attributable rather than noise:

- **Task A is solved by 11 of 14 arms**, including the no-framework control and every comparison framework on that stack. Directive 0.92 solved it. Directive 0.99 and 0.101 are two of the three failures.
- **Task B is solved by the no-framework control, two comparison frameworks, and directive's own 0.87 and 0.92.** Directive 0.99 and 0.101 fail.

These are not capability walls. The model solves both tasks unaided. Directive solved both before 0.99 and does not solve either now. That is a framework-attributable regression that has survived two releases.

## What this rules out

The 0.99 forensics attributed these two failures to the done-gate turn sink: 44–47% of all agent turns spent making `directive check` pass in a toolchain-incomplete container, crowding product verification into the last 1–2 turns. #3282 fixed that — 0.101 telemetry shows **2–8 gate invocations per trial**, down from consuming nearly half of all turns, and mean turns dropped 20% with cost down 18%.

**Both tasks still fail.** So the turn sink was real, was fixed, and was not the cause — or at minimum not the sole cause. The 0.99 forensic conclusion needs amending and the actual defect is still unidentified and still shipping.

This also means #3284 (product-first done-gate, AC runs first and never degrades) and #3267 (literal acceptance-command verification) did not recover these tasks, despite both being aimed squarely at silent product-verification failure. Either they are not firing on these tasks, or the stated-AC ladder does not reach what is actually broken. That is decidable from telemetry once #3319 lands.

## Ask

**Bisect 0.92 -> 0.99.** That range is the only place the defect can be. It is a large range and this is deliberately not a thin-slice issue — it is an investigation, and the first deliverable is a named cause, not a fix.

Suggested order, cheapest first:

1. Diff the gate / verification composition between 0.92 and 0.99 for the affected profile. #3225 (advisory-verdict blocking, made previously-advisory failures mandatory) is the leading suspect on prior-art grounds and was already flagged as a likely multiplier in #3282.
2. Check whether the ceremony dial (#3214) changed effective behaviour for this profile across that range, independent of the #3274 cold-start work that came later.
3. Reproduce locally against the two task shapes rather than waiting on external run cadence — both are ordinary hard engineering tasks, not benchmark-specific.

## Acceptance sketch

- A named cause in the 0.92 -> 0.99 range, with the mechanism explained, not just a bisect commit.
- A regression test at the appropriate boundary so the class cannot silently return.
- The 0.99 forensic conclusion amended wherever it is recorded, since "the done-gate turn sink caused these two failures" is now known to be wrong.
- If the cause turns out to be an intentional behaviour change, an explicit decision recorded on why it is worth the two tasks — per #3156 the resolution must not be to relax the gate that surfaced it.

Relates #3282 (turn sink, fixed — and the fix is what disconfirmed the old explanation), #3284 / #3267 (did not recover these), #3319 (escalation telemetry needed to decide "not firing" vs "insufficient"), #3225 (leading suspect), #3214 (ceremony dial), #3156 (gate integrity).

### LINKED ISSUE #3325: feat(envelope,ceremony): bank mid-run product pass on one-shot hard tasks — do not force merge-ready re-verify (follow-up from #3321)
## Follow-up from #3321: close the verify-to-ship gap on one-shot hard tasks

**Do not relax #3145, #3225, or consumer `task check` (#3156).** Decision: `xbrief/decisions/` keep-3153-3145-3225-despite-3321.

### Named cause (already landed)

See `docs/analysis/2026-08-13-3321-092-099-hard-task-regression.md`.

`#3153` made `drive-to: merge-ready` the default consumer unit of work. After a mid-run product pass, agents enter ~25 turns of lifecycle + toolchain, then re-verify at the end. `#3214` classifies a vanilla deposit as `project`, so mid × hard selects elevated ceremony. `#3145` added three consumer check gates inside that tail. `#3225` is not the inserter (PR merge-ready detector only).

#3284 / #3267 / #3285 did not recover the two tasks: either they did not fire (`none_stated`) or the end-of-run re-verify still overwrote the banked pass.

### Ask

Give one-shot product tasks a path that **banks a mid-run product pass** and ships without a merge-ready / elevated-ceremony tail that forces a second self-authored verify.

Suggested slices (cheapest first):

1. Honor #3285 when the official or stated product check has already passed: do not re-enter full `directive check` / pre-pr / review-cycle solely because AGENTS.md defaults to merge-ready.
2. A task-kind / envelope signal for one-shot hard tasks (not “every deposit is a project”). Explicit `non-project` already maps to minimal; vanilla `directive init` currently cannot reach it.
3. Compose #3322 (product-oracle integrity) and #3323 (derived AC) so the end-of-run check cannot replace a mid-run pass by repairing the comparison apparatus.

### Out of scope

- Editing `packages/core/src/check` or dropping #3145 gates to chase a score
- Reverting #3225
- Reproducing the private benchmark in CI

Relates #3321 (named cause), #3282 (amended), #3153, #3214, #3285, #3322, #3323, #3324, #3156.

MScottAdams:
## Summary

Investigation of the 0.92 → 0.99 hard-task A regression (#3321). Named cause with mechanism; no check-composition rollback (#3156).

**Named cause:** `#3153` made `drive-to: merge-ready` the default consumer unit of work and inserted the post-product verify-to-ship gap. `#3214` classifies a vanilla deposit as `project`, so mid × hard selects elevated ceremony. `#3145` added three consumer check gates inside that tail. End-of-run re-verify then overwrote the mid-run product pass.

**`#3225` is not the inserter.** The detector runs only on `pr:merge-ready` / `pr:watch`. The 0.99 forensic (#3282) turn-sink-as-cause conclusion is amended (disconfirmed by 0.101).

## This PR

- Analysis: `docs/analysis/2026-08-13-3321-092-099-hard-task-regression.md`
- Decision: keep #3153 / #3214 / #3145 / #3225 (`xbrief/decisions/2026-08-13-keep-3153-3145-3225-despite-3321.decision.json`)
- Ceremony-dial test locks deposit-as-project + mid × L → elevated
- Follow-up process fix: #3325
- Amendment comments on #3282 and #3321

Does not edit `packages/core/src/check` or release.

## Test plan

- [x] `task verify:investigation` on the forensic ledger
- [x] `vitest run packages/core/src/policy/ceremony-dial.test.ts` (37 passed)
- [x] `task verify:ac` (none stated / project floor)
- [x] Fast preflight of `task check` (DEFT_CHECK_MODE=full) through `verify-strategy-output`
- [x] `task ts:check-lane` exit 0 (coverage statements 88.43% / branches 85.06%)

Closes #3321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Named cause for the 0.92→0.99 hard-task regression (#3321).** `#3153` made `drive-to: merge-ready` the default consumer unit of work and inserted the post-product verify-to-ship gap. `#3214` deposit-as-project × mid × hard selects elevated ceremony; `#3145` added three consumer check gates inside that tail. `#3225` is not the inserter (review-verdict detector runs only on PR merge-ready). Amends the 0.99 #3282 turn-sink-as-cause conclusion (disconfirmed by 0.101). Analysis: `docs/analysis/2026-08-13-3321-092-099-hard-task-regression.md`. Ceremony-dial test locks deposit-as-project + mid×L→elevated. Does not relax #3145 / #3225 / check (#3156). Follow-up #3325. Closes #3321. Refs #3282, #3153, #3214, #3145, #3225, #3322, #3323, #3324.
+
 - **Overlap dispose via `decision:write` + list-before-recommend (#3315).** Portfolio-priority documents human overlap-cluster dispose (every member in `relatedIssues`, required `revisitTrigger`) and parks already-disposed pairs after `task decision:list -- --issue N`. Dedicated ledger deferred until a pass re-litigates despite a dispose record (#3310). Closes #3315. Refs #3310.
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **Named cause for the 0.92→0.99 hard-task regression (#3321).** `#3153` made `drive-to: merge-ready` the default consumer unit of work and inserted the post-product verify-to-ship gap. `#3214` deposit-as-project × mid × hard selects elevated ceremony; `#3145` added three consumer check gates inside that tail. `#3225` is not the inserter (review-verdict detector runs only on PR merge-ready). Amends the 0.99 #3282 turn-sink-as-cause conclusion (disconfirmed by 0.101). Analysis: `docs/analysis/2026-08-13-3321-092-099-hard-task-regression.md`. Ceremony-dial test locks deposit-as-project + mid×L→elevated. Does not relax #3145 / #3225 / check (#3156). Follow-up #3325. Closes #3321. Refs #3282, #3153, #3214, #3145, #3225, #3322, #3323, #3324.
+- **Named cause for the 0.92→0.99 hard-task regression (#3321).** Default merge-ready (#3153) inserted the post-product verify-to-ship gap; ceremony dial + deposit-as-project (#3214) and the #3145 consumer-check trio lengthen that tail. #3225 is not the inserter. Amends the 0.99 #3282 turn-sink-as-cause conclusion. Gates kept (#3156). Follow-up #3325. Closes #3321.
 
 - **Overlap dispose via `decision:write` + list-before-recommend (#3315).** Portfolio-priority documents human overlap-cluster dispose (every member in `relatedIssues`, required `revisitTrigger`) and parks already-disposed pairs after `task decision:list -- --issue N`. Dedicated ledger deferred until a pass re-litigates despite a dispose record (#3310). Closes #3315. Refs #3310.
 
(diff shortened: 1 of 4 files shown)


MScottAdams:
ended_at: 2026-08-13T02:11:05.400Z

Issue #3193, opened 2026-08-07
core-guard: content-aware package pin + lock follow-through (tighten #3127 path allowlist)


ape-deft:
## Problem
#3127 / PR #3138 correctly expanded `deft-core-guard` so a normal framework upgrade is **one PR**: deposit + `@deftai/directive` pin/lock + `.deft/GENERATION.json`.

The v1 implementation is **path-only** for `package.json` and lockfiles. That matches the #3127 note (“path-level allow is OK for v1; stricter pin-only is nicer if cheap”).

Residual risk: a PR that mixes `.deft/core/**` with **unrelated** package/lock edits (other deps, scripts, metadata) can still pass the guard because those paths are unconditionally exempted when present.

Consumer Greptile reviews on upgrade PRs now flag this as P2 (path-only) or, when operators hand-roll content checks in the consumer workflow, as a long P1 chase that does not belong in app repos.

## Goal
Tighten the **deposited** guard so package/lock co-travel is content-constrained:

| File | Pass only when |
| --- | --- |
| `package.json` | Diff is `@deftai/directive*` **dependency-key** pin change only (not scripts/settings that merely contain the substring) |
| lockfile(s) | Follow-through for that pin (identity + necessary transitive/resolution lines), **not** unrelated direct product dependency identity or metadata-only product edits riding the exemption |
| `.deft/GENERATION.json` | Remains path-allowlisted as upgrade stamp |

## Must still fail
- `.deft/core/**` + true app/product source
- `.deft/core/**` + package/lock changes that are not the Directive pin unit

## Non-goals
- Removing the mixed core+app guard
- Requiring two stacked upgrade PRs again
- Per-consumer hand-rolled guard forks (fix ships in the deposit)

## Acceptance
- [ ] Deposit + Directive pin + lock follow-through + GENERATION (+ installer-managed paths) **passes**
- [ ] Deposit + unrelated package.json field / unrelated lock identity or product metadata **fails**
- [ ] Documented in UPGRADING next to the one-upgrade-PR shape
- [ ] Classifier/guard tests cover pin-only vs mixed package cases (npm + at least one lock format used in CI)
- [ ] No consumer needs to patch `deft-core-guard.yml` to land a normal upgrade

## Context
- Parent: #3127 (path allowlist v1), #1430 (guard origin), #3117 (GENERATION)
- Observed: consumer upgrade reviews rate path-only as residual weakness; incomplete consumer-side content forks thrash review without fixing product SoT

## Refs
[...]

PR #3348, opened 2026-08-14
fix(deposit): loadable deft-core-guard emits no-mixed-core-and-app (#3345)

### LINKED ISSUE #3345: deft-core-guard workflow never emits no-mixed-core-and-app (invalid load; breaks classic BP)
## Summary
The deposited `.github/workflows/deft-core-guard.yml` does not successfully load as a normal Actions workflow. Consumers that still have **classic branch protection** requiring the check context `no-mixed-core-and-app` stay **merge-blocked forever**, even when the PR is a clean deposit-only SO#3 upgrade and Greptile/CI are green.

## Specimen
- Repo: `deftai/statusreport` PR #173 (Directive deposit → 0.102.0)
- Classic BP on `master` requires: `no-mixed-core-and-app`, Greptile Review, Socket Security ×2, server-quality ×3, frontend
- Org ruleset only requires Greptile (green)
- Sibling seats without classic BP (enterprize / cartograph / finalpilot) merged fine on the same wave

## Evidence
1. Workflow registry **name** stays the path string `.github/workflows/deft-core-guard.yml` instead of `deft-core-guard` → GitHub never successfully adopted the workflow `name:`.
2. All historical runs are `event=push`, **0s**, **0 jobs**, no logs (instant fail at load/dispatch).
3. Workflow declares only `on: pull_request`, but the required job `no-mixed-core-and-app` **never appears** on PR check runs.
4. Same zombie pattern on enterprize deposits; harmless there only because those repos lack classic BP requiring that context.
5. File contains a very long shell line (~5k chars) in the allowlist grep; may contribute to load failure (unconfirmed sole cause).

## Impact
- Maintain bots (e.g. `ape-deft`) cannot merge clean deposit PRs on repos with classic BP requiring `no-mixed-core-and-app`.
- Admins can still merge when `enforce_admins=false` (statusreport case).
- The guard is **not actually enforcing** mix detection on PRs while BP still demands its check name.

## Expected
1. `deft update` / deposit ships a **valid** workflow that loads with name `deft-core-guard`.
2. On `pull_request`, job **`no-mixed-core-and-app`** runs and posts a check run with that exact context.
3. Optional: document that consumers using classic BP must either keep this check green or remove it if they rely only on the org Greptile ruleset.

## Workarounds (consumer)
- Admin merge / bypass while workflow is broken
- Remove `no-mixed-core-and-app` from classic BP until fixed
- Prefer org ruleset-only protection (matches enterprize/cartograph/finalpilot)

## Env
- Directive pin wave: **0.102.0**
- Host: 2026-08-13
- Reporter: ape-deft / Number One (Deft ops)

MScottAdams:
## Summary

Fix deposited `.github/workflows/deft-core-guard.yml` so GitHub Actions loads it with name `deft-core-guard` and the `no-mixed-core-and-app` job posts on `pull_request`.

**Root cause:** the #3193 pin-content Python heredoc body was emitted at column 0, which terminated the YAML `run: |` block. Actions never adopted the workflow name (registry stayed the path string; runs were 0s/0 jobs). Classic branch protection requiring `no-mixed-core-and-app` merge-blocked forever (statusreport PR #173).

**Fix:**
- Indent every Python body line into the run block (`CORE_GUARD_RUN_INDENT`)
- Emit installer-managed allowlist as one ERE per line + `grep -vE -f` (no ~5k-char shell line)
- Mix detection semantics unchanged
- Tests assert load shape; CHANGELOG + UPGRADING re-deposit note

Consumers need `deft update` / re-deposit to receive the fixed workflow.

Closes #3345

## Test plan

- [x] `pnpm exec vitest run packages/core/src/init-deposit` (346 passed)
- [x] `task verify:ac` (3 clauses verified + literal vitest command)
- [x] Biome clean on touched files
- [ ] CI green on this PR


MScottAdams:
ended_at: 2026-08-14T01:48:00.279Z

PR #3333, opened 2026-08-13
feat(verification,policy): flag product-oracle method-change pass

MScottAdams:
## Summary

Extend #3156 to product oracles: a red verification may be resolved only by a product change or an independently re-derived oracle.

- Run-summary `verification` event records check id, method fingerprint, and outcome.
- fail then method-change then pass on one check id is flagged.
- verify:ac treats comparison-method mutation as unresolved unless independent re-derivation is recorded.
- Build and pre-pr gain the MUST/forbid pair via pack source + packs:render.

## Related Issues

Closes #3322

## Checklist

- [x] `/deft:change` — N/A (swarm story with existing scope xBRIEF)
- [x] `CHANGELOG.md` — added entry under `[Unreleased]`
- [x] `ROADMAP.md` — N/A
- [x] Tests pass locally (`task check` green)

## Test plan

- Targeted vitest on packages/core/src/verify-ac and packages/core/src/run-summary
- task packs:render
- task check (verify:ac first, then suite)

Issue #3322, opened 2026-08-13
feat(verification,policy): gate integrity for product oracles - a red verification may not be self-adjudicated by editing the comparison; pass-after-fail-with-method-change is a flagged event


MScottAdams:
**TL;DR: gate integrity (#3156) protects framework gates from being edited to go green. Nothing protects the agent-derived PRODUCT oracle. Transcript evidence shows the exact failure: a final equivalence check contradicted an earlier pass, the agent repaired the comparison apparatus instead of the product, got green, and shipped a confident success claim that the official verifier failed. Ask is one rule plus one telemetry event: a verification discrepancy may never be self-adjudicated by modifying the verification method.**

**Evidence (internal benchmark transcript, private):** on a hard task whose acceptance criterion is output equivalence, the agent produced a correct-looking solution and verified equivalence mid-run ("perfect match"). It then spent ~25 turns on lifecycle and toolchain work, and re-verified at the very end. The final check **diverged** ("Results differ now! That's concerning."). The agent attributed the divergence to a corrupted comparison artifact from an earlier run, rebuilt the comparison a different way, observed a match, and shipped — final message: output "matches the original (same order and values)". The official verifier scored 0. Budget exhaustion is ruled out: the run ended at roughly a third of max turns.

The structural defect: **the same agent that produced the artifact adjudicated the discrepancy in its own favor by editing the oracle, not the product.** This is precisely the move #3156 forbids for framework gates ("clear red by editing the gate definition"), applied to the improvised product oracle where no rule currently reaches.

## Rule (thin, fail-closed — #3265)

1. **A red verification result may be resolved only by (a) a product change, or (b) an independently re-derived oracle** — both sides of the comparison rebuilt from scratch, by a different method than the one that failed. In-place repair of the failing comparison (fixing the reference file, changing the diff invocation, re-generating one side) followed by a pass is **not** a pass — it is an unresolved discrepancy and MUST be surfaced as such in the done report (#1006 fail-loud; Agent Trap Defenses' surface-the-anomaly).
2. **Pass-after-fail-with-method-change is a named, detectable event.** The #3282 run summary gains a verification event: `{check id, method fingerprint, outcome}`. A sequence `fail → method fingerprint changes → pass` on the same check id is flagged in the summary and in `verify:ac` output. This composes with #3319 (decision events) and #3320 (denominator) — same emitter, one more event type.
[...]

PR #3349, opened 2026-08-14
feat(verify-ac): namespace product-oracle check ids per active scope

### LINKED ISSUE #3337: feat(verify-ac): namespace product-oracle check ids per active scope (#3322 residual)
## TL;DR

#3322 / PR #3333 flags fail-then-changed-method-pass on one verification check id. Every active xBRIEF currently emits the same constant \erify:ac\ check id under one session. A fail from story A and a pass from story B can look like one unresolved method-change sequence (false deny). Track that isolation here so #3333 does not grow again.

## Why this is separate

PR #3333 already covers:

- verification events with method fingerprint + outcome
- fail-closed unless independent re-derivation
- in-process evaluation when \DEFT_RUN_SUMMARY_PATH=-
The leftover Greptile 4/5 is **multi-active-scope pairing**, not the field self-adjudication #3322 was written for. #2881 / #3273: do not expand #3333 mid-review.

## Ask (thin, fail-closed)

1. Each active scope emits a distinct check id (story id or xBRIEF path), so attempts do not pair across briefs in the same session.
2. Same-scope fail then changed-method pass still flags.
3. Cross-scope fail/pass is not an unresolved discrepancy.

Out of scope: changing #3322 oracle rule; empty-resolution #3334; rung-2 #3323.

## Acceptance sketch

- Two active briefs in one \erify:ac\ run cannot produce a cross-scope method-change flag.
- Same-brief fail then different-method pass still fails closed without independent re-derivation.
- Tests pin both cases. CHANGELOG cites this issue. Refs #3322 #3333.

MScottAdams:
## Summary

Namespace product-oracle `check_id` values per active scope so multi-active `verify:ac` under one session does not false-deny across briefs.

- `verifyAcCheckId(scopeKey)` returns `verify:ac/<scope>` (fallback `verify:ac` when scope unknown)
- `emitVerifyAcAttempts` accepts `scopeKey` / `checkId`
- `evaluateVerifyAcFromPlan` / `FromPath` thread `plan.id` or path stem as `oracleScopeKey`
- Tests pin: cross-scope fail/pass is not flagged; same-scope fail then different-method pass still fails closed
- CHANGELOG cites #3337

## Related Issues

Closes #3337
Refs #3322 #3333

## Checklist

- [x] CHANGELOG.md entry under Unreleased
- [x] Tests pin both cross-scope and same-scope cases
- [x] verify:ac green on active scope
- [x] Biome and tsc on changed paths green

## Test plan

- pnpm exec vitest run packages/core/src/verify-ac (43 pass)
- pnpm exec vitest run packages/core/src/product-first-done-gate (32 pass)
- task verify:ac on active scope xBRIEF
- CI suite on this PR


MScottAdams:
ended_at: 2026-08-14T01:55:34.240Z

Issue #2881, opened 2026-07-28
bug(process): babysit/review must not expand active story scope past xbrief AC — open follow-up or amend before coding


MScottAdams:
## Description

When Greptile or another reviewer raises a design-adjacent concern during babysit of a PR for story S, agents implement multi-commit redesigns in-tree without promoting a new scope xBRIEF or amending S plan.items. That corrupts story accounting (what the active brief authorized vs what shipped), blocks merge on confidence while original acceptance criteria may already be met, and teaches agents that review is a second unbounded implementation mandate. Recurrence: PR #2871 babysit expanded #2869 Finding 3 (writeVbrief containment) into issue-emit create/stamp durability and pending-URL ledger work that later had to be extracted as #2880.

## Acceptance

- **Review-cycle / babysit skills: out-of-AC findings default to follow-up issue or brief amendment, not silent code expansion.**: Given an active story xBRIEF with fixed plan.items, when a Greptile finding is outside those Acceptance criteria, when the skill is followed, then the agent files or proposes a follow-up scope (or amends the brief with operator consent) before implementing a new subsystem.
- **Document P0-in-touched-surface carve-out vs new-subsystem-needs-story rule.**: Given the skill text, when read, then P0 security fixes on already-touched files may land in the same PR, while new ledgers/idempotency protocols/cross-cutting reliability contracts require a separate or amended story.
- **Clarify confidence-only holds (0 P0/P1) vs blocking findings for babysit behavior.**: Given review-cycle or pr:merge-ready guidance, when confidence is below threshold with zero P0/P1, then the skill does not mandate unbounded redesign; it offers follow-up, residual-risk note, or operator override paths.
- **Document issue:emit related-ref footgun (github-issue type marks already tracked).**: Given docs or skill notes for issue:emit, when an author adds a related-issue reference, then guidance prevents using emit-tracking github-issue type for non-primary related work (or emit only honors a designated primary ref).

## Traces

missing_traces_justification: process incident from PR #2871 babysit; acceptance items are the authoritative trace until skill paths are named in implementation.

MScottAdams:
## Related

- AppSec containment batch: #2869
- PR where babysit expanded past AC: #2871
- Product follow-up extracted from that expansion: #2880
- Scope xBRIEF: `xbrief/proposed/2026-07-28-babysit-must-not-expand-active-story-past-xbrief-ac.xbrief.json`

## Recurrence (one paragraph)

[...]

PR #3353, opened 2026-08-14
fix(run-summary): seed file-destination seq from JSONL line count (#3350)

### LINKED ISSUE #3350: bug(run-summary): seq is 1 on every emitted line - per-process emitter never seeds from the target file; seqSeed has no callers (#3320 post-close residual)
**TL;DR: every line in collected field run-summary JSONL carries `seq: 1`. The schema documents "monotonic per-emitter sequence (1-based)", but every real event comes from a separate CLI process, each constructing a fresh emitter. The `seqSeed` reuse mechanism exists in `packages/core/src/run-summary/emit.ts` but has no callers. A field that promises ordering and never delivers will get built against; ordering currently falls back silently to `ts`.**

(Originally reported as a comment on #3320 after it closed — re-filed so it has an open home.)

**Evidence:** collected field telemetry (internal benchmark, private) shows `seq: 1` on 100% of lines across all trials on both releases that have shipped the emitter. At HEAD: `emitRunSummaryEvent` ("one-shot helper for call sites that do not hold an emitter") constructs `new RunSummaryEmitter(...)` per call; grep for `seqSeed` finds only its own doc comment.

## Fix (pick one, thin)

- **Option A (preferred):** when the destination is a file, seed `seq` from the current line count of the target file at emitter construction (the file is append-only JSONL; count-then-append gives a monotonic per-session sequence with no extra state). Stdout destination keeps per-process seq and the docs say so.
- **Option B:** drop the field and document `ts` as the order key.

Either way the schema comment stops promising something the wire never delivers.

## Acceptance sketch

- A session producing N events into one `DEFT_RUN_SUMMARY_PATH` file yields seq 1..N (option A), or the field is gone and docs name `ts` (option B).
- Test covers the multiple-CLI-invocation case, not just one emitter instance.

Relates #3320 (shipped the denominator; this defect was reported there post-close), #3282 (emitter).

### LINKED ISSUE #3361: bug(run-summary): seq lock live-holder reclaim needs ownership check (#3350 residual)
## Summary

PR #3353 landed Option A file-destination seq seed plus a sibling lock and stale-lock reclaim. Greptile on HEAD cc7da373995d81633bf0b4848358f6ba98509e70 still holds at Confidence 3/5 with advisory should-not-merge.

## Residual

Inline P1: if a live writer spends longer than the 2s wait window recounting or appending, another process deletes the lock by mtime and both can append the same seq. Greptile also wants fail-closed acquire (no proceed without the lock).

## Why not in #3353

- Fail-open is the #3282 run-summary contract (emission never throws / never changes exit codes).
- Owner-token / live-PID reclaim is a new lock protocol, not the thin Option A count-then-append story.
- Review-cycle #2881: do not invent a new reliability subsystem mid-babysit.

## Suggested follow-up (thin)

- Write a lock owner token (pid + nonce) and reclaim only when that process is dead.
- Keep fail-open after a bounded wait; do not block CLI exit codes.
- Test live-holder vs dead-holder reclaim.

Refs #3350 #3353 #3282.

MScottAdams:
## Summary

Seed run-summary seq from the existing append-only JSONL line count when the destination is a file. Each CLI process used to construct a fresh emitter, so every collected line carried seq: 1. File destinations now continue 1..N across constructors and one-shot emitRunSummaryEvent calls. Stdout destination stays per-process, and the schema comment matches that wire behavior.

#3361 residual: `.seq.lock` stores a pid+nonce owner token. Reclaim is allowed only when that owner process is dead or the token is unknown. A live holder is never stolen by mtime. Release unlinks only when the token still matches this holder. Fail-open after the bounded wait is unchanged (#3282).

## Related Issues

Closes #3350
Closes #3361

## Checklist

- [x] /deft:change <name> — N/A: focused bugfix in packages/core/src/run-summary/** plus append-only CHANGELOG; not cross-cutting or architectural
- [x] ROADMAP.md — N/A: not a tracked ROADMAP item
- [x] Tests pass locally (pnpm exec vitest run packages/core/src/run-summary 37/37; task check exit 0)

## Test plan

- [x] Multi-constructor file dest yields seq 1..N
- [x] emitRunSummaryEvent one-shot calls continue seq
- [x] Stdout dest stays per-process (each constructor starts at 1)
- [x] Default-path session_start truncate resets seq to 1
- [x] Live-holder reclaim does not steal the lock even when mtime is old
- [x] Dead-holder and unknown-token reclaim succeed
- [x] Release unlinks only when the owner token still matches


MScottAdams:
ended_at: 2026-08-14T17:28:55.234Z

MScottAdams:
Review-cycle dual-stop: fix batch 2/3 on HEAD cc7da373995d81633bf0b4848358f6ba98509e70. Reclaimed abandoned seq locks older than the wait window (#3350 Greptile P1). MCP unavailable -- gh api REST fallback.

MScottAdams:
Confidence-only hold on prior HEAD cc7da373 (0 P0/P1, Confidence 3/5, advisory should-not-merge). Residual live-holder lock reclaim is tracked as #3361 and is out of this #3350 Option A story (#2881). Fail-open after a bounded wait stays the #3282 contract.

Merge-path rebase onto master (#3354) only; no lock-ownership protocol added. New HEAD 855212f8. MCP unavailable -- used gh api REST fallback for review comments.

Issue #3462, opened 2026-08-18
design(update,dry-run): make dry-run fidelity a property of construction, not helper discipline -- side-effect architecture for deft update


MScottAdams:
# design(update,dry-run): make dry-run fidelity a property of construction, not helper discipline — side-effect architecture for `deft update`

Refs: #3437 (consumer bug, stays open) · #3392 / #3458 (chokepoint) · #3452 (tripwire arc) · PR #3453 (halted at pre-committed stop)

> **Amended pass (2026-08-18): simplified per directive KISS/low-token values.** v1 of this body ran a four-candidate survey with an N≥3 panel. That was surface inflation (#1140). This version names one presumptive mechanism, ⊗-rejects the alternatives with reasons, and charters N≥1 critic to refute the presumption. Accepted outcome lands as an ADR in `docs/decisions/` plus a `## Current shape` comment here. **Decision gate unchanged: no further product code on #3453 until the Current shape lands.**

## Summary

PR #3453's collect-only dry-run failed to stabilize after three batches. Five bypass instances (inventory below). The fifth — `writeConsumerGitHooks` running `git config core.hooksPath` on dry-run — proves the write surface is not one category, so any architecture relying on every helper remembering a discipline (mode flag, root parameter) keeps leaking. Choose the mechanism whose fidelity holds by construction.

## KISS bar (evaluate everything against this first)

The consumer's question (#3437) is **blast radius**: "what will `deft update` touch?" The failed promise was **execution parity**: "the plan equals exactly what the writer would do." Parity is what generated all five bypasses. A deterministic answer to the consumer's question does not require simulating the writer.

## Presumptive mechanism — pure-read plan, honestly labeled

The refresh's file semantics are "make dest match projected content." So the plan is a pure function of two trees, plus a static declaration:

1. **Version skew** from two reads (manifest `.deft/VERSION` vs content `package.json`) — already implemented on the halted branch, cherry-pickable independently.
2. **File plan** = diff(projected content tree, current dest tree) over the installer-managed path set. Pure reads. No writer code executes; no mode, no port, no scratch dir. Deterministic by construction.
3. **Honest label** for what the diff cannot see, printed every time: `+ projection repairs and side effects not in this count: git config core.hooksPath, hook registration` — a static list maintained next to the side-effecting helpers, enforced by a content-contract test that fails when a new side-effect category appears in the deposit path without a matching label entry.

[...]

PR #3468, opened 2026-08-18
feat(update): dry-run dest plan via side-effect port (ADR-004)

### LINKED ISSUE #3437: [framework-gap] deft update --dry-run reports 'State: current / no changes written' immediately before the real run rewrites 1,585 files
## Summary

deft update --dry-run reports 'State: current / no changes written' immediately before the real run rewrites 1,585 files

## Context

Consumer repo on a vendored deposit whose recorded install manifest had drifted to v0.78.0 while the deposited content was newer; engine 0.103.0 at the time. The gated session ritual failed on agent_hooks registration drift across claude/cursor/grok/codex. The remediation it prints is 'run deft update (or directive init) to refresh project hooks', which reads as a narrow hook-registration repair -- so --dry-run was run first specifically to preview the blast radius before touching anything.

## Expected behavior

--dry-run should classify the file-level effect before it happens: a changed-file count and the top-level paths involved, or at minimum 'manifest 0.78.0 -> content 0.103.0, N files will be rewritten'. A caller running --dry-run specifically to preview blast radius should not be told 'no changes written' when 1,585 files are about to be rewritten.

## Actual behavior / gap

deft update --dry-run printed the header '[deft update] dry-run -- classified plan (no changes written)' with State: current, Mode: proceed, Root cause: 'engine 0.103.0 reachable; no pin to reconcile against', and JSON 'dry_run: true, update_state: current'. Nothing signalled a pending file-level rewrite. The real deft update then staged 1,585 files: 1,485 under .deft/, plus .claude/, .cursor/, .grok/, a newly created .codex/, .github/skills/, AGENTS.md, .githooks/pre-commit and pre-push, .gitignore, .prettierignore, xbrief/schemas/ and xbrief/.deft-version. It reported previous_version 0.78.0 -> content_version 0.103.0.

## Session notes

Impact: a routine documentation-only release cut became an unplanned framework deposit upgrade that had to be split onto its own branch to keep it out of the release commit. That branch is now permanently unmergeable in this repo -- Greptile is a required status context on main, it declines review above 500 changed files, and it posts its refusal as a PR comment rather than a status, so the required context can never go green. 'State: current' is especially misleading when the recorded manifest and the deposited content disagree by 25 minor versions; that skew is arguably the headline the dry-run should lead with. Separately, a --dry-run --json field carrying the planned changed-file list would let consumers gate on blast radius in CI. Minor adjacent nit found while filing this: repeated --context flags on feedback:file silently keep only the last value rather than accumulating.

---
_Filed via `task feedback:file` from a directive consumer project (Refs #1709 value-feedback gap escalation)._

MScottAdams:
## Summary

`deft update --dry-run` dest plan is now recorded port calls (ADR-004), not a collect-only walk and not a dest-vs-content-package matcher.

- Dest write, remove, chmod, and dest-mutating git/exec go through `containedWrite` / `containedRemove` plus `containedChmod` / `containedDestExec` in `packages/core/src/fs/contained-write.ts`.
- Dry-run binds that port in record mode. The planned dest set is the recorded calls.
- The three inventoried seams fold into the port: `GitHooksSeams.setHooksPath`, `copyContent`/`replaceTree`, `runOrgForceOn`.
- Static ban covers dest-mutating `execFileSync("git", …)` on the update path. Allowlist is the port module.
- Planted `xbrief/vbrief.md` dry-run fixture stays byte-identical.
- Printed exclusions label names OpenClaw `$HOME` skills until that site folds.

Closes #3437
Refs #3462
Refs ADR-004
Prior art: PR 3453 (collect-only, replaced — do not revive)

## Test plan

- [x] `pnpm exec vitest run packages/core/src/init-deposit packages/core/src/fs/contained-write.test.ts`
- [x] Dry-run zero-mutation fixture with planted `xbrief/vbrief.md`
- [x] Contract test: dest-mutating git outside the port fails
- [x] Exclusions label content-contract


MScottAdams:
owner: grok-build-3462-leaf
pr: 3468
head: 639409806ab25d58f68064ff5179baccd2dba40e
mode: parent-retained
review_cycle: in_progress:3468#parent-retained
fix-batch: 2 P1 dest-root symlink guard

Issue #1704, opened 2026-06-17
RFC: gate throughput without quality loss — affected-test selection, merge queue, caching, and the full-gate-at-merge invariant


MScottAdams:
## Problem

On a large project, `task check` becomes a throughput bottleneck. Directive is at the pessimal point for this today:

- **No incremental mode.** The `verify:*` gates deliberately disabled go-task `sources:`/`generates:` caching (to avoid swallowing CLI flags), so every gate re-runs in full every time.
- **No merge queue.** CI is plain `on: pull_request` + `push: master`; the heavy lane is re-implemented as discrete steps (ruff / mypy / pytest / go), and the full `task check` is never invoked as a single enforced command (see "enforcement gap" below).
- **Swarm multiplies it.** Every swarm worker runs the full `task check` in its worktree before commit, and it doubles as the consent gate (`skills/deft-directive-swarm/SKILL.md`). Cost scales as roughly `O(commits × full-gate)`, plus the #1681 concurrent-basetemp footgun.

This RFC proposes how to raise throughput **without** quality loss. It is a companion to #1703 (framework-performance measurement) -- they share the load-bearing metric -- and it interacts materially with #1530 (Python -> TypeScript / RivetOS-Nx migration), which changes the *tooling* for every lever below.

## The invariant that makes it safe

> **Every change MUST pass the full gate at least once before it merges. You may change the *frequency* and *location* of expensive checks, never whether they run before merge.**

Get that right and throughput is a scheduling/caching problem, not a quality tradeoff. The target cost model:

`O(commits × full-gate)`  ->  `O(merges × full-gate) + O(iterations × cheap-proxy)`

Fast local/iteration lane uses a cheaper proxy (affected tests + static gates); the full gate becomes the **merge chokepoint** -- mandatory and complete.

### 1. Affected-only test selection for the iteration lane (biggest lever)
Run only tests affected by changed files during iteration; run the whole suite at the merge gate.
- Tooling: **pre-migration** = `pytest-testmon` / coverage-based import-graph map. **Post-#1530 (Nx)** = `nx affected` is native.
- Quality guard: Test Impact Analysis misses dynamic dependencies (config-driven tests, reflection, dynamic imports) -- so it is NEVER the final word. The **full suite runs at merge**; the fast lane is convenience only.

### 2. Merge queue / batching
Run the full gate once per *batch* against the would-be-merged state (GitHub merge queue / Bors / Zuul).
[...]

Issue #747, opened 2026-04-29
feat(scripts,tasks,hooks): detection-bound branch gate — verify:branch + git hooks reading PROJECT-DEFINITION policy field


MScottAdams:
## Summary

Implement detection-bound enforcement of the existing `!` rule from `AGENTS.md`:

> `Always work on a feature branch -- never commit directly to master/main unless the user explicitly instructs it or PROJECT-DEFINITION.vbrief.json narratives contain Allow direct commits to master: true`

Today the rule is prose-only (tier 4 on the detection-strength axis from #634). It demonstrably fails to bind: in a 2026-04-29 session reproducing #136's symptom, the strongest available agent ducked the rule on a benign-looking JSON enrichment task and would have committed directly to `master` without prompting. This issue ships the layered tier-3 enforcement mechanism that operationalizes the rule.

## Relationship to other issues

- **Companion to #746** — #746 owns the *policy surface* (typed field, interview elicitation, skill-level defense, reconfigure UX). This issue owns the *mechanical enforcement* (hooks, `verify:branch`, scripts, CI). #746 explicitly declared this gate out-of-scope so the two issues stay reviewable independently.
- **Hard prerequisite: #746 acceptance criterion A** — typed `policy.allowDirectCommitsToMaster` boolean on `PROJECT-DEFINITION.vbrief.json` plus schema/migration. Without the typed field, this gate's check is fragile (string-vs-boolean ambiguity, narrative-key drift). Implementation does not start until #746 part A lands.
- **Motivated by #136 (reopened)** — Warp's auto-loading of `AGENTS.md` does not give framework rules the binding strength of Global Rules; even after `AGENTS.md` is loaded, rules in conversation history degrade in salience. Detection-bound enforcement does not depend on the agent attending to a rule at decision time.
- **Independent of #742 (ADR-001)** — this gate operates on the detection-strength axis; #742 governs the encoding-form axis. The gate's enforcer reads a typed field on `PROJECT-DEFINITION.vbrief.json` regardless of whether the underlying rule is encoded as `AGENTS.md` prose or future rules-pack vBRIEF.
- **Prior art (closed but evidently incomplete)** — #186 ("Agent skips pre-implementation checklist after loading AGENTS.md") and #171 ("Agents must not commit/push directly to master -- add hard gate to framework") both closed COMPLETED, yet the failure mode reproduces today. This issue is the determinism-tier escalation those prior issues stopped short of.

## Layered design

[...]

PR #3516, opened 2026-08-19
fix(ci): exempt verify:branch on merge-gate CI checkout (#3499)

### LINKED ISSUE #3499: bug(ci): master post-merge merge-gate is structurally red — verify:branch refuses the default branch
## Summary

The `Merge gate (Blacksmith primary)` job fails on **every** post-merge run on `master`, and has for at least 12 consecutive runs. The cause is not a test failure — the suite is fully green — it is `verify:branch` refusing to operate on the default branch.

## Evidence

Run `32258367489`, head `586c5692`:

```
Test Files  1013 passed (1013)

=============================== Coverage summary ===============================
Statements   : 89.09% ( 141116/158389 )
Branches     : 85.35% ( 46194/54121 )
Functions    : 96.29% ( 6428/6675 )
Lines        : 89.09% ( 141116/158389 )
================================================================================

task: Failed to run task "check:merge": task: Failed to run task "check:framework-source":
      task: Failed to run task "verify:branch"
##[error]Process completed with exit code 201.
```

Every test passes and all four coverage metrics clear the 85% floor. The gate that fails is `verify:branch`.

Consecutive failing master runs (all `CI` workflow, `Merge gate`):

| SHA | when |
|---|---|
| `586c5692` | 2026-08-19 13:29 |
| `16076833` | 2026-08-19 13:01 |
| `afc3a21d` | 2026-08-19 03:12 |
| `68341617` | 2026-08-19 02:55 |
| `b489da78` | 2026-08-19 02:48 |
| `5c9ab11e` | 2026-08-19 01:28 |
| `96384229` | 2026-08-19 01:03 |
| `ddf601c1` | 2026-08-19 00:31 |
| `8c1ea4fc` | 2026-08-18 23:03 |
| `1907fe76` | 2026-08-18 22:02 |
| `cbae6e53` | 2026-08-18 21:02 |
| `58234460` | 2026-08-18 20:45 |

The window extends earlier than the oldest row listed; 12 is where the query stopped, not where the failures begin.

## Cause

`verify:branch` enforces the #747 branch-protection policy: refuse to operate on the configured default branch. On a **pull request** run the checkout is a feature branch, so the gate passes — which is why required PR checks are green. On a **post-merge** run the job checks out `master` itself, so the same gate inside `check:framework-source` refuses.

`check:merge` is documented as the merge-chokepoint SoT (#1704). Running it on the default branch therefore always trips a gate whose purpose is to prevent working on the default branch.

## Why this matters

1. **The master signal is noise.** A permanently-red master conditions everyone to ignore it, so a genuine post-merge regression would look identical to the standing failure.
2. **It weakens the #1704 premise.** "CI is the merge SoT, so local full `task check` is optional" holds for PR runs, but the post-merge confirmation that the merged result is green never completes.
3. **It is invisible from the PR side.** Every PR merged during 2026-08-18/19 showed `Merge gate (task check) pass`; master went red immediately after each merge.

## Proposed change

Decide what `check:merge` means on the default branch, then make it consistent:

- **Option A — exempt the gate in CI.** `verify:branch` guards against *authoring* on the default branch; a read-only CI verification checkout is not that. Pass the documented exemption for the merge-gate job only, never for the local authoring path.
- **Option B — drop `verify:branch` from `check:framework-source`.** It is a policy/posture gate about where you are committing, not a property of the code under test, so it may not belong in the composed check at all.
- **Option C — do not run `check:merge` post-merge on master.** If the PR run is the authority, the post-merge job is redundant and should be removed rather than left failing.

⊗ Silencing the job or marking it non-required without deciding which of these is intended — that converts a structural failure into a hidden one.

## Acceptance

- A post-merge `master` run of the merge gate either passes or does not run.
- The chosen option is recorded, so it is clear whether post-merge master verification is expected to exist.
- A genuine post-merge regression is distinguishable from the standing state.

Refs #1704 (merge chokepoint SoT), #747 (branch protection), #3479

MScottAdams:
## Summary

Option A only for #3499. The post-merge master merge-gate was structurally red because `verify:branch` refuses the default branch. That gate is an authoring-posture check; a read-only CI verification checkout is not authoring.

This sets the documented exemption `DEFT_ALLOW_DEFAULT_BRANCH_COMMIT=1` on the two `task check:merge` steps (`merge-gate-primary` and `merge-gate-failover`) only. Workflow-level `env:` is unchanged, so TS/Go lanes and tests do not inherit the bypass. Local `task check` and git hooks stay fail-closed on master.

Options B and C are out of scope. The choice is recorded in `xbrief/decisions/`.

## Acceptance

- Post-merge master merge-gate can pass `verify:branch` via the documented env.
- Local authoring path is unchanged.
- Chosen option is recorded (Option A).

Closes #3499
Refs #1704, #747


MScottAdams:
ended_at: 2026-08-19T18:23:59.212Z

PR #3520, opened 2026-08-19
fix(release): derive build-dist archive from git-tracked files

### LINKED ISSUE #3490: fix(release): build-dist packs untracked files — denylist model ships host-local state into the archive
## Summary

`buildArchive` (release Step 8) walks the working tree with a **basename denylist** and no notion of which files are tracked. Every untracked file whose directory is not explicitly listed is packed into the release archive.

Measured on this host right now: **39,680 untracked files under `.claude/` alone**, zero of them tracked. That includes `.claude/settings.local.json` and full nested source checkouts under `.claude/worktrees/<agent-id>/`.

#3487 / PR #3489 adds `.claude` to the shared exclusion set, which removes today's largest instance. **The class remains:** the next untracked directory that isn't on the list ships too.

## Evidence

`packages/core/src/release/build-dist.ts`:

```
89      names = readdirSync(dir).sort();
94      if (excludes.has(name)) continue;
```

That is the entire filter. No `git ls-files`, no `.gitignore` awareness, no tracked-file check.

```
$ git ls-files .claude | wc -l
0
$ find .claude -type f | wc -l
39680
```

The count grows with use — it was ~28,700 earlier in the same session and rose as agent worktrees were created. Any host-specific untracked tree scales the archive without bound.

## Why a denylist is the wrong model here

This is a **source distribution archive**. The set of things that belong in it is "files this project tracks", which git already knows exactly. Enumerating what to leave out means:

- every new tool that drops a working directory needs a framework release to be excluded
- correctness depends on developer-host layout, so the artifact differs between machines
- the failure is silent — an over-large archive is not obviously wrong at build time

An allowlist derived from `git ls-files` inverts this: unknown directories are excluded by default, and the archive is reproducible across hosts.

## Security note, stated carefully

I did not read file contents and am not claiming a disclosed secret. But the shape is unfavourable: `.claude/settings.local.json` is local, host-specific, and never intended for distribution, and agent worktrees can contain arbitrary working state. A release cut from a developer machine — which is how this project cuts releases, since the release verbs run locally — embeds whatever untracked state that machine happens to hold. Worth treating as a release-integrity issue with a privacy edge rather than a pure size bug.

## Proposed change

1. **Derive the archive file set from `git ls-files`** (plus any deliberately-generated build outputs the release needs, added explicitly). Keep the exclusion set as defence in depth, not as the primary mechanism.
2. **Fail closed on surprise:** if the resolved file set contains anything untracked and not on an explicit generated-output allowlist, error rather than pack it.
3. **Assert reproducibility:** a test that an archive built from a checkout with extra untracked directories present matches one built without them.

## Acceptance

- Archive contents are identical whether or not `.claude/`, `.deft-scratch/`, or any other untracked tree exists in the checkout.
- Adding a new agent-host working directory to a developer machine cannot change a release artifact.
- Building with an unexpected untracked file in the resolved set fails rather than shipping it.

## Notes

Found while reviewing PR #3489 (#3487), which correctly adds `.claude` to the shared exclusion set. That fix is right and should land on its own; this issue is the underlying model.

Refs #3487 / PR #3489 (adds `.claude`; the immediate mitigation)
Refs #2953 / #2954 (added `.deft-scratch` to the same denylist for the same reason — the pattern repeating is the point)

MScottAdams:
## Summary

- Derive the release archive from git ls-files plus an explicit generated-output allowlist.
- Keep the basename denylist as defence in depth, not the primary filter.
- Fail closed if the resolved set contains an untracked surprise.
- Tests: extra untracked host trees (.claude/, .deft-scratch/, .new-agent-host/) cannot change archive contents.

Closes #3490


MScottAdams:
ended_at: 2026-08-19T23:39:04.735Z

PR #3530, opened 2026-08-20
feat(verify): report uncovered changed branches in forward-coverage (#3514)

### LINKED ISSUE #3514: feat(verify): forward-coverage is new-file-existence only, not diff coverage — the compensating control for a lowered floor does not exist
## Summary

`verify:forward-coverage` (#1310) is documented in its own header as **"v1 = NEW-file granularity only"**. For each newly added source file it asserts that a corresponding test file was added in the same diff. That is a file-existence check, not diff coverage.

It does **not**:

- measure coverage of changed code
- inspect modified (non-new) files at all
- enforce any per-diff coverage percentage

This matters now because the global floor is being lowered to 75 (#3512), and diff-level enforcement was the stated justification for doing so. That justification does not currently hold.

## The uncovered path

Modify an existing, poorly covered file:

- no new source file → `verify:forward-coverage` passes
- aggregate barely moves on a large codebase → the global floor passes

Nothing gates it. Since most work is modification rather than new-file creation, this is the common case, not the edge case.

## Why it is being filed now rather than later

[Google's guidance](https://testing.googleblog.com/2020/08/code-coverage-best-practices.html) is explicit that project-wide targets above 90% are not worth chasing, while **per-commit coverage of ~90% for new code is reasonable**. That split is the evidence-backed part of coverage gating, and it is the half Directive does not implement.

[Inozemtseva & Holmes (ICSE 2014)](https://dl.acm.org/doi/10.1145/2568225.2568271) found aggregate coverage only weakly predictive once suite size is controlled — which is a good argument for a low global floor **provided** the diff gate is real. With the floor at 75 and the diff gate at file-existence, neither control is doing much.

## Proposed change

Extend forward-coverage from file-existence to **diff coverage**: of the lines/branches added or modified in this change, what fraction is covered by the suite. Gate on that.

Design questions this issue should settle:

- **Threshold for new code.** Google suggests ~90% per-commit. That is far above the 75 global floor by design — new code is cheap to cover, legacy is not.
- **Lines, branches, or both?** Branches is the binding metric globally; for a diff gate, lines may be the more actionable signal.
- **Where does the data come from?** The v8/AST coverage report already produced by the suite, intersected with the diff. No new instrumentation should be needed.
- **Warn or fail, and when?** Failing closed on day one would block any PR touching legacy files. Warn-first with a reporting period is the safer path, mirroring `verify:telemetry-coverage`.
- **Does it run locally or in CI?** It needs a coverage report, so it is suite-adjacent. On the #1704 iteration lane this should not be a per-commit local cost.

## Acceptance

- A change that modifies an existing file and adds uncovered branches is reported.
- A change adding a well-covered new file passes, as today.
- The existing new-file-existence assertion is preserved or subsumed, not dropped.
- The gate's threshold and the global floor are documented as serving different purposes, so neither is mistaken for the other.

## Known, accepted gap in the meantime

#3512 lowers the global floor to 75 with this control absent. That is a deliberate, recorded tradeoff — current branch coverage is 81.23%, so roughly 6pp of aggregate erosion could occur before anything fires. `coverage:hotspots` is the interim watch. This issue closes that gap properly.

Refs #3512, #1310, #3480 / PR #3510

MScottAdams:
## Summary

`verify:forward-coverage` now intersects `coverage/coverage-final.json` with the diff and reports uncovered added/modified branches. New-file existence stays fail-closed. Diff coverage is warn-first (`--enforce` fail-closes).

The 90% per-change branch threshold is not the 75 global floor: 90% covers new code, 75 is the aggregate collapse detector.

## Related Issues

Closes #3514

## Checklist

- [x] `/deft:change <name>` -- N/A: swarm-cohort story with operator through-merge consent; covered by `task check`
- [x] `CHANGELOG.md` -- added entry under `[Unreleased]`
- [x] `ROADMAP.md` -- N/A (issue not listed)
- [x] Tests pass locally (`task check` exit 0)

## Test plan

- Unit tests in `packages/core/src/forward-coverage` cover modified uncovered branches (warn-first and `--enforce`), well-covered new files, preserved existence failures, and unchanged-line non-reports.


MScottAdams:
ended_at: 2026-08-20T05:39:42.095Z

Issue #1250, opened 2026-05-20
bug(triage,policy): task triage:welcome materializes plan.policy.wipCap, violating #1186 Deliverable 1


MScottAdams:
## Summary

`task triage:welcome` materializes `plan.policy.wipCap=10` into `vbrief/PROJECT-DEFINITION.vbrief.json` and appends an entry to `meta/policy-changes.log`. This directly violates the existing policy contract documented in #1186 Deliverable 1 ("deft accepts the framework default 10 -- the typed field is omitted from PROJECT-DEFINITION") and breaks the maintainer-design test `tests/content/test_consumer_config_example.py::test_policy_omits_wip_cap`.

## Reproduction (2026-05-20 dogfood session)

1. Fresh `vbrief/PROJECT-DEFINITION.vbrief.json` with no `plan.policy.wipCap` key.
2. Run `task triage:welcome` (or `task triage:welcome -- --no-subprocess`). Accept default `wipCap=10` when prompted.
3. Observe `vbrief/PROJECT-DEFINITION.vbrief.json` diff: `+      "wipCap": 10` materialized at the end of `plan.policy`.
4. Observe new file `meta/policy-changes.log` with row `2026-05-20T12:51:16Z actor=triage-welcome field=plan.policy.wipCap value=10 previous=None changed=true`.
5. Run `task check` (or `pytest tests/content/test_consumer_config_example.py::test_policy_omits_wip_cap`). Result:

```
AssertionError: plan.policy.wipCap must remain unset -- deft accepts the framework default 10 per #1186 Deliverable 1
```

The PR carrying this materialization (#1249) is unmergeable until the materialization is undone.

## Expected

`task triage:welcome` step 4 (wipCap prompt) should:
- Read the current `plan.policy.wipCap` if present.
- If unset and the operator confirms the framework default, EXIT THE STEP WITHOUT WRITING THE FIELD. The default is read from framework defaults; consumers do not need an explicit setting.
- Only write the typed field when the operator changes the value to something other than the default, AND only if `#1186` permits the consumer to opt out of the default.
- Skip the `meta/policy-changes.log` audit row when no actual change is being made.

## Actual

The ritual writes the field unconditionally on operator confirmation, regardless of whether the value matches the framework default. The `meta/policy-changes.log` audit row is appended even when `value=10 previous=None` (no actual change).

## Impact

- Breaks `task check` on any session that runs `task triage:welcome` against the framework's own repo (the dogfood case).
[...]

Issue #746, opened 2026-04-29
feat(setup,skills,policy): no-feature-branch opt-out — interview disclosure, skill-level guard, and reversible reconfiguration


MScottAdams:
## Summary

When a user opts out of feature-branch enforcement during the setup interview (`Allow direct commits to master: true` in `PROJECT-DEFINITION.vbrief.json`), they lose access to a significant slice of deft's capabilities — but the framework currently doesn't (a) elicit the choice explicitly, (b) disclose the cost at decision time, (c) defend at skill-invocation time when an incompatible skill is invoked, or (d) provide an easy path for the user to change their mind later. This issue tracks the full surface needed to make the opt-out a first-class, reversible, transparent choice rather than a silent footgun.

Spawned from the rule-binding investigation around #136 (reopened) and the proposed detection-bound branch gate (separate issue forthcoming).

## Background

`AGENTS.md` already canonizes the opt-out:

> `! Always work on a feature branch -- never commit directly to master/main unless the user explicitly instructs it or PROJECT-DEFINITION.vbrief.json narratives contain Allow direct commits to master: true`

Once a detection-bound branch gate ships (planned in `task check` + git hooks + CI), the opt-out becomes mechanically respected. But honoring the opt-out at the gate layer surfaces three downstream questions:

1. How is the opt-out elicited from the user in the first place? Currently the `deft-directive-setup` skill does not explicitly ask.
2. What happens when an opted-out user invokes a skill that fundamentally requires feature branches (swarm, review-cycle, pre-PR, release)? Currently the skill would either fail confusingly or proceed and corrupt state.
3. How does an opted-out user opt back in (or vice versa) after their project's collaboration model changes?

## Affected skills (capability cost of opting out)

A no-branch project cannot run any of these in their canonical form:

- `skills/deft-directive-swarm` -- requires worktrees + per-agent feature branches by design (Phase 2 Step 1 creates branches).
- `skills/deft-directive-review-cycle` -- requires a PR; PRs require a non-default head ref.
- `skills/deft-directive-pre-pr` -- prepares a branch for PR.
- `skills/deft-directive-release` -- requires the merge-cascade discipline that branches enable.
- Any future skill that assumes a feature-branch workflow.

What still works on a no-branch project:

- vBRIEF authoring (`/deft:change`, `/deft:run:interview`, `/deft:run:speckit`, `/deft:run:discuss`, `/deft:run:research`, `/deft:run:map`).
- `task check` (validate + lint + test + verifiers).
[...]

Issue #57, opened 2026-03-17
Add GitHub Actions CI workflow for linting and tests on PRs and pushes


jsgoecke:
## Status

**Partially implemented** — minimal CI workflow landed in `fix/72-vbrief-generation` branch (commit 3405ccd).

### What's done
- `.github/workflows/ci.yml` — runs `task check` (ruff lint, mypy type-check, pytest 794 tests) on all PRs and master pushes
- Python 3.11 + uv + Task runner
- Gates merges until all checks pass

### What remains
- [ ] **Branch protection**: After first merge to master, enable required status checks on `master` in GitHub Settings → Branches (requires CI to have run once on master first)
- [ ] **Go job**: `go test ./cmd/deft-install/` + `go build` for platform matrix (linux/amd64, darwin/arm64, windows/amd64)
- [ ] **Coverage reporting**: pytest `--cov` with threshold enforcement in CI

### Related
- #128 — CI vBRIEF schema sync check (depends on this issue)
- #115 — Strengthen spec validation gate (spec_validate.py now has schema checks; CI freshness check still needed)

## Original Description

Add GitHub Actions CI workflow for linting and tests on PRs and pushes to master.

Jobs:
1. Python: ruff check, mypy tests/, pytest tests/ with coverage
2. Go: go test ./cmd/deft-install/ + go build for platform matrix (linux/amd64, darwin/arm64, windows/amd64)

PR #3536, opened 2026-08-20
fix(policy): do not append no-op rows to policy-changes.log (Closes #3528)

### LINKED ISSUE #3528: policy-changes.log holds two contradictory no-op conventions; 63 of 236 rows record nothing (#1250 vs typed-policy writers)
## Summary

`meta/policy-changes.log` carries two contradictory no-op conventions. The typed-policy verbs append an audit row even when nothing changed; the #1250 welcome path suppresses it. Neither is recorded as a decision. The "always append" side dominates, and 63 of the ledger's 236 rows now record that nothing happened.

The operational cost is that a read-only-in-effect verb dirties a tracked file, which trips release Step 1 pre-flight.

## Observed

During the v0.105.0 cut, `task check` ran a `product-signal:enable` and left the tree dirty with:

```
2026-08-20T00:28:20Z actor=task product-signal:enable productSignal.enabled=true
  sinkRepo=deftai/product-signal previous={"enabled":true,"sinkRepo":"deftai/product-signal"}
```

New value equals previous value.

Ledger composition (236 rows):

| Rows | Verb | Real transitions |
|---|---|---|
| 64 | `product-signal:enable` | **1** (the initial `previous=null` on 2026-07-23) |
| 45 | `policy:enforce-branches` | 45 — genuine |
| 45 | `policy:allow-direct-commits` | 45 — genuine |
| 82 | assorted `actor=deft` one-offs | genuine |

The 90 enforce/allow rows are repetitive but legitimate — they are the release ritual flipping branch protection off and back on, and each is a real state change. The no-op population is the 63 `product-signal:enable` rows.

## This is deliberate, which is the actual problem

`packages/core/src/policy/product-signal.ts:256-276` computes `changedFlag`, guards the PROJECT-DEFINITION write on it, then appends unconditionally:

```ts
const changedFlag =
  previousNormalized.enabled !== nextBlock.enabled ||
  previousNormalized.sinkRepo !== nextBlock.sinkRepo;
policyBlock.productSignal = nextBlock;
if (changedFlag) {
  atomicWriteProjectDefinition(path, data);   // write is guarded
}
...
appendAuditLog(projectRoot, parts.join(" "));  // append is not
```

and says so to the operator:

```
no-op: value already matched (audit entry still appended for trail).
```

The same shape and the same string appear in `ceremony-dial.ts:756`, `value-feedback.ts:405`, and `require-human-merge.ts:422`. So the typed-policy family consistently chose "record the attempt."

#1250 chose the opposite for the wipCap welcome path (`packages/core/src/triage/welcome/onboard.ts:115`), where accepting the default deliberately does **not** materialize the field or write a row.

Both positions are defensible. Recording attempts has real value for a security-relevant ledger — "who tried to flip this and when" is not nothing. Suppressing no-ops keeps the trail readable. What is not defensible is holding both in the same file with no decision explaining which applies where.

## Why it matters beyond tidiness

1. **Tree dirtying before a release.** The append happens during `task check`, so a verb that changed no state leaves a tracked file modified. Release Step 1 is `git status --porcelain`, so this must be reverted or committed before every cut. It surfaced exactly that way on v0.105.0.
2. **Signal dilution in a security artifact.** This ledger is the audit trail for `allowDirectCommitsToMaster` transitions (#746/#747) and was hardened against symlink redirection in #2470. It is meant to be read during incident review. A quarter of it saying "nothing changed" is the failure mode audit logs die of.
3. **The `previous=` field already carries the information.** A reader can distinguish a real transition from a no-op by comparing new against `previous` — but only by parsing every row, which is what suppression or an explicit token would avoid.

## Proposed change

Pick one convention and record it. Either direction closes this:

**Option A — suppress no-ops.** Guard `appendAuditLog` on `changedFlag` in the four typed-policy writers, matching #1250. Simplest, and makes the ledger all-signal.

**Option B — keep recording attempts, mark them.** Add an explicit `changed=true|false` token to the row and keep appending. `content/UPGRADING.md:868` documents exactly this token for the `actor=triage-welcome` rows, so the format precedent already exists in-tree. A reader greps `changed=true` for the transition history. This preserves the "who tried" value that the current copy is reaching for.

**Either way**, the no-op append must stop dirtying the tree during `task check`, or `product-signal:enable` should not run there at all.

Option B is the better fit if the "for trail" intent is genuine, because it keeps the evidence and makes it filterable. Option A is right if that intent was never load-bearing.

⊗ Rewriting or pruning existing rows. The ledger is append-only by design (#746) and hardened (#2470); historical no-op rows stay.

## Acceptance

- One convention applies to every writer of `meta/policy-changes.log`, recorded as a decision (`task decision:write`) naming which and why.
- `#1250`'s welcome path and the typed-policy verbs agree.
- A `product-signal:enable` that changes nothing does not leave a tracked file modified after `task check`.
- If Option B: rows carry a machine-readable changed token, and `content/UPGRADING.md:868`'s existing `changed=` documentation is generalized to cover them.
- Existing rows are untouched.

## Affected writers

- `packages/core/src/policy/product-signal.ts:275`
- `packages/core/src/policy/ceremony-dial.ts:756`
- `packages/core/src/policy/value-feedback.ts:405`
- `packages/core/src/policy/require-human-merge.ts:422`
- `packages/core/src/policy/org-force-on-migration.ts:565`
- shared sink: `packages/core/src/policy/resolve.ts:208` (`appendAuditLog`)

Refs #746, #747, #1250, #1186, #2470, #2980

MScottAdams:
## Summary

One convention for meta/policy-changes.log: record real transitions only. Typed-policy writers and the welcome path skip the append on no-op, stamp changed=true on real writes, and leave historical rows byte-identical. A no-op product-signal:enable no longer dirties the tree during 	ask check.

Decision: xbrief/decisions/2026-08-20-policy-changes-log-records-real-transitions-only-written-rows-st.decision.json.

## Related Issues

Closes #3528

Refs #1250, #746

## Checklist

- [x] /deft:change <name> — N/A (single-issue bugfix; not a /deft:change proposal)
- [x] ROADMAP.md — N/A (issue not listed)
- [x] Tests pass locally (	ask check exit 0)

## Post-Merge

- [ ] **Verify issue auto-close**: After squash merge, confirm referenced issues actually closed.
- [x] Enable branch protection on master requiring CI status check (one-time setup, see #57) — N/A this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Policy ledger no-ops no longer dirty the tree (#3528).** Typed-policy writers and the welcome path now share one convention: real transitions append `changed=true`; a no-op `product-signal:enable` (and sibling policy verbs) leave `meta/policy-changes.log` unmodified. Existing rows stay. Refs #1250, #746.
+
 ### Removed
 
 ## [0.105.0] - 2026-08-19
--- a/content/UPGRADING.md
+++ b/content/UPGRADING.md
@@ -874,7 +874,7 @@ Six phases, each detection-bound so a partial re-run resumes cleanly:
 
 ### Audit trail
 
-Every policy write (Phase 2 / Phase 4) appends an entry to `meta/policy-changes.log` with `actor=triage-welcome`, the field name, the new value, the previous value, and a `changed=true|false` token. This mirrors the existing `scripts/policy.py::set_policy` audit format so `git grep triage-welcome meta/policy-changes.log` surfaces every ritual run.
+Every **real** policy write (Phase 2 / Phase 4, and every other `meta/policy-changes.log` writer) appends an entry with the field name, the new value, the previous value, and a machine-readable `changed=true` token. Welcome rows still use `actor=triage-welcome`. Typed-policy verbs (`product-signal:enable`, `policy:enable-value-feedback`, `policy:set-ceremony-dial`, `policy:allow-direct-commits`, …) stamp the same `changed=` token. A no-op (value already matched) does **not** append, so `task check` cannot dirty the tracked ledger. Historical rows may omit the token or record `changed=false`; do not rewrite or prune them (#3528 / #746). `git grep 'changed=true' meta/policy-changes.log` is the transition history.
 
 ### References
 
--- a/packages/cli/src/dispatch.test.ts
+++ b/packages/cli/src/dispatch.test.ts
@@ -1159,6 +1159,23 @@ describe("native policy-set handler (#2022)", () => {
     expect(readFileSync(auditLogPath(), "utf8")).toBe(before);
   });
 
+  it("equal-value wip-cap still persists a legacy plan.policy key migration", async () => {
+    writeFileSync(
+      projectDefPath(),
+      JSON.stringify({
+        xBRIEFInfo: { version: "0.8" },
+        plan: { title: "T", status: "running", items: [], policy: { wipCap: 5 } },
+      }),
+      "utf8",
+    );
+    const result = await runPolicy(["wip-cap", "--set", "5", "--confirm", "--project-root", root]);
+    expect(result.code).toBe(0);
+    expect(result.out).toContain("audit:");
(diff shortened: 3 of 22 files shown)


MScottAdams:
ended_at: 2026-08-20T04:59:20.118Z

PR #3588, opened 2026-08-21
fix(hooks): gate Shell dest-forms through active-xBRIEF (#3438)

### LINKED ISSUE #3438: Scope gate: file mutations via Bash bypass the active-xBRIEF check that blocks Edit/Write
## Summary

The active-xBRIEF scope gate refuses file mutations made through the host's `Edit`/`Write` tools, but the same mutations made through `Bash` (`git checkout --`, `rm`) go through unchallenged. Nothing downstream catches it either — the pre-commit gates don't require an active xBRIEF — so a shell-only change can be made and committed entirely outside a scope.

## Evidence

Single session, `deftai/agentsentri`, `xbrief/active/` empty, the two calls seconds apart:

**Allowed** — four tracked/untracked files mutated:

```
git checkout -- apps/web/tsconfig.json apps/web/next-env.d.ts && rm apps/web/AGENTS.md apps/web/CLAUDE.md
```

**Refused** — the very next call, a one-line edit to `.gitignore`:

```
Directive denied Edit: No active xBRIEF artifact was found under xbrief/active/
(or the legacy vbrief/active/ compatibility path). Recovery: run `deft scope:activate -- <path>` ...
```

Same repo, same empty `active/`, same class of mutation. One surface enforces, the other doesn't.

The commit path doesn't close the gap. After scoping, my commits ran:

```
verify_encoding / verify_forward_coverage / verify_vbrief_conformance / verify_xbrief_drift / branch-protection
```

None of those requires an active xBRIEF, so the shell-only version of this change would have committed clean.

## What I can't determine from outside

`.claude/settings.json` (Directive-managed) does register a `PreToolUse` hook on Bash:

```json
{
  "matcher": "Shell|Bash|BashTool|shell|bash",
  "hooks": [{ "type": "command", "command": "deft-hook --host claude --event tool.before", "timeout": 5 }]
}
```

So Bash is wired. Whether the hook fired and declined to scope-gate, or didn't fire at all, I can't distinguish from the consumer side. The gates I *observed* firing around commit and push (`deft branch-protection`, `deft destructive-gh-verb gate (pre-push)`) appear to be git-level hooks rather than the tool hook, so they aren't evidence either way.

## Why it matters

The refusal message is emphatic and reads as authoritative, which makes the gate easy to trust as complete. An agent that reaches for `sed -i`, `rm`, or `git checkout --` — including one following host guidance to prefer shell tools — silently operates unscoped. The gap is widest exactly where scope matters most: cleanup and revert work, which is naturally shell-shaped.

In my case the unscoped mutation was the change I actually wanted, and I scoped it retroactively. The concern is that nothing required me to.

## Expected

Either the Bash `tool.before` hook applies the same active-xBRIEF check to recognized file-mutating commands, or the refusal message stops implying coverage it doesn't have. A commit-time active-scope check would also close it, at the cost of catching things later.

## Environment

- `@deftai/directive` 0.104.0 (engine `@deftai/directive-core@0.104.0`)
- Deposit generation 4, contentVersion 0.104.0, stamped 2026-08-15
- Host: Claude Code, `deft-hook --host claude`
- Platform: darwin
- Consumer repo: `deftai/agentsentri` (xbrief layout, not legacy vbrief)

## Related

- #2885 (closed) touched the same write gate from the containment angle — absolute paths outside the checkout. This is the complementary direction: paths inside the checkout, reached through a different tool surface.

MScottAdams:
## Summary

Shell dest-forms (`git checkout -- <paths>`, `git restore`, `rm`/`rmdir`) now go through `inspectMutationGates` with Edit/Write exemptions (assist/scratch, proposed lifecycle, story `file_scope`). Edit/Write `scope-not-ready` copy names direct-write and spawn. `git status` stays fail-open. `python -c`, `cmd /c copy`, and obfuscated bash stay known-open.

## Test plan

- [x] `pnpm exec vitest run packages/core/src/hooks`
- [x] Reporter compound `git checkout --` plus `rm` denies with empty active
- [x] `git status` / `python -c` / `cmd /c copy` remain fail-open

Closes #3438


MScottAdams:
ended_at: 2026-08-23T02:33:12.905Z

MScottAdams:
Human merge override (operator, this session).

- HEAD: `28f4b790c4be160d882fbb81985bb6e27ab8405e`
- Greptile: SHA match, P0=0, P1=0, confidence **4/5** vs dogfood floor **5** (`clean_gate_holdout=confidence`). Advisory should-not-merge in review body.
- Residual: dest-form tokenizer / double-quoted backslash (`rm "foo\bar"`) and shell dialect. **Parked on #3624**, which was split out of #3438 / #3588 so this PR can close. Not another dest-form redesign on this branch.
- Review-cycle babysit hit dual-stop cap 3/3. Operator authorized override rather than a fourth batch or lowering `minGreptileConfidence`.
- `Closes #3438`. Does **not** close #3624.

Refs #3624, #3438, #2881, #3225, #3095.

Issue #3616, opened 2026-08-21
test(vitest,windows): CLI-subprocess tests time out under post-#3480 parallelism — capping workers regresses #3480 instead of fixing it


MScottAdams:
## Symptom

On win32 with many cores, CLI-subprocess tests fail as **timeouts, not assertions**, and the
failing set changes run to run. Observed durations 22–32s against a 20s cap.

Config (`vitest.config.ts`):

- `testTimeout: isWin32 ? 20_000 : 5_000` (line 208)
- `maxWorkers: winMaxWorkers` where `winMaxWorkers = max(1, min(12, floor(cpus * 0.25)))` (line 217)
- On a 32-core box that resolves to **8 fork workers with a 20s per-test cap**

## Evidence

Six full-suite runs of the same branch, same machine, within a few hours — failure counts and
*identities* both unstable:

| Run | Failures | Files |
| --- | --- | --- |
| 1 | 54 | 20 |
| 2 | 59 | 24 |
| 3 | 86 | 23 |
| 4 | 46 | 10 |
| 5 | 8 | 7 |
| 6 | 21 | 13 |

Runs 5 and 6 were **functionally identical** in the code path under test, yet differed in both
count and file set — which is the signature of contention, not logic.

Representative failures, all CLI-subprocess spawns, all ~20s+:

```
× inspectCoverageDebt / inspectCheckResume + CLI aliases      32215ms
× policy:show registers CLI alias and canonical path          32339ms
× registers policy:show field and alias                       32536ms
× inspects via policy:show alias and canonical name            32722ms
× refreshes .deft/core and rewrites a stale managed section    22255ms
```

Counter-evidence that this is not the code under test:

- The same files pass in isolation (`session/` alone: 53 files / 471 tests, 0 failures).
- The CLI itself is fast standalone: `policy:show --field=runtimeAuthority` and `--version`
  both return in well under a second.
- The failing files do not import the modules changed on the branch under test.

## Root cause hypothesis

`maxWorkers` bounds **vitest workers**, not the **subprocesses those workers spawn.** Each
CLI-alias test launches `node packages/cli/dist/bin.js`, so real concurrency is 8 workers × their
children, and Windows process creation is expensive. A 20s cap that was comfortable when files ran
serially is not comfortable at that multiple.

The history explains the calibration gap:

- #2546 added the win32 worker cap for coordinator/RPC saturation.
- #3480 **dropped `fileParallelism: false`** to stop the suite running serial on 32 cores, and the
[...]

Issue #3480, opened 2026-08-18
perf(test,windows): upgrade vitest 4 and drop fileParallelism:false — local suite runs serial on 32 cores


MScottAdams:
## Summary

`vitest.config.ts` disables file parallelism on win32 whenever coverage is on. `task check` runs coverage. So the local suite executes its ~1008 test files **one at a time** regardless of core count, while Linux CI runs them in parallel. This is the dominant term in the 40-minute local `task check` reported in #3479.

## Evidence (2026-08-18, 32-logical-core Windows 11 host, repo at `9e826084`)

| Run | Result |
|---|---|
| `pnpm exec vitest run` (no coverage, 8 workers) | **486s / 8m06s** wall; 14,023 passed / 149 skipped |
| ↳ internal breakdown | collect 980s + tests 1222s + prepare 377s of **CPU** |
| Same work serialized (`fileParallelism: false`) | ~2580s ≈ **43 min** — that CPU cannot overlap |
| CI `task check:merge`, identical monolith, 4-vCPU Blacksmith | **4m01s** job total incl. checkout + `pnpm install` + Go/uv setup |

The projection reproduces the field-reported ~40 minutes, and the CI number confirms the gate itself is small — this is purely a local-execution artifact.

## Cause

Three win32 workarounds in `vitest.config.ts`, all added for vitest 3.2.x coverage races (#2546 / #2580 / #2634):

- `fileParallelism: false` when `isWin32 && coverageEnabled` — the expensive one
- `maxWorkers` / `poolOptions.forks.maxForks` capped to `min(4, floor(cpus * 0.25))` under coverage
- `coverage.processingConcurrency: 1`

The config comment already names the exit: upstream fix `vitest-dev/vitest#10117`, shipped in **vitest 4**. The repo is on **vitest 3.2.6**. #2634 was closed as the ENOENT bug, not as the parallelism restoration, and no open issue tracks the upgrade.

## Proposed change

1. Upgrade vitest 3.2.6 → 4.x.
2. Remove the `fileParallelism: false` branch and restore normal worker counts under coverage on win32.
3. Confirm the #2580 `coverage/.tmp` ENOENT race and the #2546 `onTaskUpdate` RPC timeout do not recur — a green full-suite-with-coverage run on Windows is the acceptance signal. (One `onTaskUpdate` timeout was still observed on 3.2.6 during the measurement above, so this is also the fix for that flake class.)
[...]

Issue #2546, opened 2026-07-14
test(vitest): full suite passes then process fails on unhandled onTaskUpdate worker timeouts (Windows)


MScottAdams:
## Problem

On native Windows, `pnpm run test` / `task check` can report **all tests green** and still exit non-zero because Vitest records unhandled worker RPC errors:

```
Error: [vitest-worker]: Timeout calling "onTaskUpdate"
```

Observed during v0.77.0 release Phase 1 preflight on `master` (2026-07-14):

- Test Files: **629 passed** | 2 skipped
- Tests: **9129 passed** | 51 skipped
- Errors: **2** (`onTaskUpdate` timeouts)
- Duration ~99s; then `[ELIFECYCLE] Test failed` / `ts:check-lane` fails
- Same pattern on immediate retry — not a single-test assertion failure

Stack points at Vitest 3.2.6 RPC (`vitest/dist/chunks/rpc.*.js` `onTimeoutError`), not application code under test.

## Why it matters

- Blocks honest `task check` / release Step 5 preflight on Windows maintainers even when the suite is logically green
- Easy to misread as a real regression and burn release time
- May be load/RPC related (large parallel suite + coverage) rather than a flaky unit test

## Expected

- Either the worker RPC timeouts do not fail the process when all tests passed, or
- Pool/timeout/worker config is tuned so `task check` is deterministic on Windows, or
- A documented, gated workaround for release preflight that does not hide real failures

## Non-goals

- Skipping the suite
- Treating this as coverage-threshold failure (coverage still printed after the unhandled errors)

## Acceptance

- [ ] Repro notes (OS, Node, Vitest, `pnpm run test` vs `task check`) captured
- [ ] Root cause class identified (Vitest RPC timeout vs hung worker vs coverage reporter)
- [ ] Fix or durable mitigation so Windows `task check` does not fail solely on `onTaskUpdate` after a green suite
- [ ] Regression guard or known-flake doc pointer if mitigation is config-only

## Environment (repro)

- OS: Windows 10/11 native (not WSL)
- Repo: `deftai/directive` @ master during v0.77.0 cut prep
- Vitest: 3.2.6 (from lockfile / stack)

## Related

- Release skill Phase 1 uses `task check` when `task ci:local` is absent

Issue #2467, opened 2026-07-13
Windows-native task check fails: path.resolve and basename assumptions are POSIX-only


MScottAdams:
## Summary

On a Windows-native checkout (not WSL), `task check` / `pnpm run test` fails with a large wave of assertion errors. Release cut for v0.75.0 was blocked after Phase 1 preflight on native Windows; the same suite passes under WSL/Linux.

## Symptoms

Representative failures:

1. **`path.resolve` rewrites POSIX absolute fixtures**
   - Expected: "/custom/slices.jsonl"`
   - Received: "C:\custom\slices.jsonl"`
   - Example: `packages/core/src/triage/queue/cache-branches.test.ts` → `resolveSlicesLogPath({ slicesLogPath: "/custom/slices.jsonl" })`
   - Cause: production helpers correctly call `path.resolve(...)`; on win32, Node treats `/custom/...` as absolute from the current drive.

2. **Basename via `split("/")` on Windows paths**
   - Expected: `["a.xbrief.json", "b.xbrief.json"]`
   - Received: full `C:\Users\...\active\a.xbrief.json` strings
   - Example: `packages/core/src/triage/refresh/extract.test.ts` → `paths.map((p) => p.split("/").pop())`
   - Cause: Windows paths use `\`; splitting on `/` never splits.

Order-of-magnitude from one native run: ~145 failing assertions across many packages/core (and related) suites — mostly the same two patterns, not unrelated product bugs.

## Non-causes (ruled out during release preflight)

- npm publish / `NPM_TOKEN` — publish uses OIDC trusted publishing in Actions; not machine-local.
- Missing pnpm alone — installable via user-prefix `npm i -g pnpm`; toolchain then proceeds into tests.
- Pack-drift — separate Windows env issue (`core.autocrlf=true` checks out CRLF; pack-drift compares LF). Workaround: `git -c core.autocrlf=false checkout -- <projections>`. Track as hardening below.

## Recommended fix

1. **Tests (primary):** Make path assertions platform-aware.
   - Prefer `join(tmpdir(), ...)` / relative fixtures over hardcoded "/tmp/..."` / "/custom/..."`.
   - When an absolute POSIX string must remain as input to a `resolve()` helper, assert against `resolve(input)` (or `normalize`) instead of the literal Unix string.
   - Replace `p.split("/").pop()` with `basename(p)` from `node:path`.
[...]

PR #3623, opened 2026-08-21
fix(test,windows): raise win32 testTimeout to 120s (#3616)

MScottAdams:
One-line config change with a measured justification.

## Why

Windows process creation drains through a **fixed-rate chokepoint** — AV filter drivers scanning on execute — measured at ~2.4–3.5 spawns/s **at any concurrency**:

| Command | conc 1 | conc 8 | conc 16 |
| --- | --- | --- | --- |
| `git --version` | ~409ms | 2300ms | 4604ms |

Throughput stays flat; only latency multiplies. A git-fixture test issuing 10–30 sequential spawns therefore runs 25–80s.

A full suite run on such a machine produced **164 failures across 38 of 1028 files, every one timeout-class** — min 20.04s, p50 25.6s, p90 50.4s, max 81.5s, **zero assertion failures**. A *passing* test was clocked at 16.9s, so the suite was grazing the old 20s cap rather than sitting inside it.

The 20s value was calibrated when #3480 still ran files serially, and was never revisited after that change parallelised them. 120s clears the measured worst case with margin.

## Deliberately not doing

**Lowering `maxWorkers` or re-serialising files.** Spawn throughput is concurrency-independent, so capping parallelism costs wall-clock and buys no headroom. That is why the "cap workers locally, run the task, revert" workaround kept getting rediscovered without ever fixing anything — it never reduced throughput, it only stopped tests straddling the timeout. It also regresses #3480, whose directive ("do not serialize files or coverage processing") is still in the config.

If timeouts return, the next step is partitioning spawn-heavy suites into their own vitest project — **71 of 1028 files carry 64.6% of aggregate test time and 79% of the failures** — keeping the unit lane at full parallelism. Details and a smoke-tested config sketch are on #3616.

## Scope

- `packages/core/src/vitest-runner/vitest-config-contract.test.ts` does not pin `testTimeout`; it needs no change and still passes (16 tests).
- Linux/CI stays on the 5s default — this is win32-only, and CI was never affected (its runners lack the AV, which is why the merge gate has been green while local `task check` failed).

## Complementary, out of repo

AV exclusions for the repo, `%TEMP%`, and `git.exe`/`node.exe` roughly halved the serial spawn cost on the reporting machine (409ms → ~200ms). That makes the suite *faster*; this change makes it *pass*. Both are wanted.

Refs #3616, #3480, #2546, #2467.

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -205,7 +205,23 @@ export default defineConfig({
     include: ["packages/*/src/**/*.test.ts"],
     // Windows git fixture suites (session:start) exceed the 5s default under
     // full-suite parallelism; Linux CI stays on the default. Refs #2467.
-    testTimeout: isWin32 ? 20_000 : 5_000,
+    //
+    // 120s, not 20s (#3616). Windows process creation drains through a
+    // fixed-rate chokepoint (AV filter drivers scan on execute), measured at
+    // ~2.4-3.5 spawns/s at ANY concurrency: a serial `git --version` costs
+    // ~400ms and 16 concurrent cost ~4.6s each, while throughput stays flat.
+    // A git-fixture test issuing 10-30 sequential spawns therefore takes
+    // 25-80s, and the whole suite grazes the old cap -- a full run produced
+    // 164 failures, every one a timeout (min 20.04s, max 81.5s), zero
+    // assertion failures. The 20s value was calibrated when #3480 still ran
+    // files serially and was never revisited after it parallelised.
+    //
+    // ⊗ Do not "fix" this by lowering maxWorkers or re-serialising files.
+    // Spawn throughput is concurrency-independent, so capping parallelism buys
+    // nothing but wall-clock and regresses #3480. If timeouts return, the next
+    // step is partitioning spawn-heavy suites into their own vitest project
+    // (71 of 1028 files hold 79% of the failures) -- not trading cores away.
+    testTimeout: isWin32 ? 120_000 : 5_000,
     ...(coverageEnabled
       ? {
           teardownTimeout: 120_000,


MScottAdams:
## Validated: 154 failures to 2

Full `packages/core` run on `master` (`f2ca40fd`) with `--testTimeout=120000` (behaviourally identical to this PR's config change), on a machine where AV exclusions had also been applied:

| | Baseline (20s cap, no exclusions) | This change + AV exclusions |
| --- | --- | --- |
| Failed tests | **154** | **2** |
| Failed files | 32 of 1028 | 2 of 904 |
| Passed | 13,055 | 13,210 |
| Wall clock | 1348.6s | **994.8s** |

The two survivors are not timeout-class and not new: `session/branches.test.ts` (asserts on branch state, sensitive to the checked-out branch) and `session/package-manager-network.test.ts` (network). Both were failing before this change and both fail independently of the timeout.

Contributions are separable and both worth having:

- **AV exclusions** (out of repo, operator action) took the serial `git --version` spawn cost from ~409ms to ~200ms and account for the ~26% wall-clock drop.
- **This change** is what converts the remaining timeout tail into passes: the measured worst failure was 81.5s against a 20s cap.

Note the exclusions alone would not have been sufficient — at ~200ms/spawn a 10-30 spawn fixture test still lands in the 2-6s range per spawn burst under 8-way concurrency, and the old 20s cap left no margin. Nor is this change dependent on them: 120s clears the pre-exclusion worst case of 81.5s on its own.

Acceptance bar from #3616 was three consecutive green runs plus a margin guard. This is run 1 of 3 at 2 pre-existing failures; the remaining two runs and the margin check are worth doing before calling #3616 closed.

Issue #3693, opened 2026-08-24
fix(intake): a GitHub App installation token cannot prove which App it belongs to


MScottAdams:
A GitHub App installation token cannot prove which App it belongs to by any **side-effect-free read**. That is the recorded limit. Installation credentials are **not accepted as worker principals** pending a GitHub-side observation path.

Split from #3665 during its review cycle. Bound by design-critique successor lean 5455183890 (round 1, 3 of 3, 17 headings all-accept). This body is recut to that map. The 2026-08-28 synthesis comment 5455210374 named the same remedy but missed the completed-arc sentence shape.

## Decision

**Direction 4.** Keep failing closed. Installation credentials are not accepted as worker principals until GitHub ships an observation path (an endpoint or introspection that discloses App identity to the installation token without reading the token value, #3664).

This records acceptance item 1. Fail-closed is already the shipped *function* (`FAILURE_INSTALLATION_IDENTITY_UNVERIFIABLE` in `packages/core/src/intake/github-auth-modes.ts`). It is **not** the enforcing SCM gate. That gap is #3858 and is out of this issue's subject.

## The limit (restated)

The original premise "nothing callable with that token discloses the App's identity" is false: a write is callable, and installation-authored comments / check runs can carry App identity after the fact. The claim that survives is:

**No side-effect-free read reachable by an installation token discloses which App it belongs to.**

Illustration, not the argument:

| Endpoint | Returns App identity? | Callable with an installation token? |
|---|---|---|
| `GET /user` | n/a | **No** — 403, an installation has no authenticated user |
| `GET /app` | yes | **No** — requires a JWT signed with the App's private key |
| `GET /repos/{owner}/{repo}/installation` | yes | **No** — same JWT requirement |
| `GET /installation/repositories` | **no** | yes — returns repositories, not the App |

Reading the token itself is separately forbidden (#3664). Nobody on the critique arc held a live App installation credential; inside-out behaviour remains inferred from published schemas.

## Struck

- **Declared App slug as a success path (old direction 3).** Not salvageable. `GitHubAuthValidationResult` is a boolean `ok`. Three production consumers key on it. A `proof: verified | declared` field does not close it. Keep "declared vs verified" as the *reason* the validator fails closed, never as a success kind.
- **Attributed write then readback.** Listed so it is not invisible. Struck: it requires write capability before authorization, and the probe target is attacker-influenceable through `deriveValidationRepo`.
[...]

Issue #1351, opened 2026-05-25
grok-build / hybrid: orchestrator should pre-approve and inject GH credentials for sub-agents (follow-up from #1349 review-cycle handoff)


MScottAdams:
> **Recut 2026-08-24** after a two-round design-critique arc ([write-back](https://github.com/deftai/directive/issues/1351#issuecomment-5389716939) · [round 1](https://github.com/deftai/directive/issues/1351#issuecomment-5389751712) · [successor lean](https://github.com/deftai/directive/issues/1351#issuecomment-5389839201) · [round 2](https://github.com/deftai/directive/issues/1351#issuecomment-5389863402)). The May 2026 text is preserved at the bottom. Mechanism unchanged; scope, sequencing and obsolete references corrected.

## Problem

A dispatcher has no supported way to hand a worker the GitHub credential it approved for that worker. `content/patterns/multi-agent.md:208` pins env-var injection as the contract, and `:204-206` records that v1 ships it as operator-facing runbook prose only, per #983 non-goals. #1351 is the deferred automation of that non-goal.

This is a **declared deferral, not a compliance gap.** `content/templates/agent-prompt-preamble.md:451` carries two MUSTs — inject the credential, and record the selected `github_auth_mode`. The same sentence says v1 keeps injection operator-implemented on purpose.

The record MUST is **half implemented**, not implemented. Both `:451` and `:127` are conjunctive — the mode must be recorded in the launch manifest **and** the worker dispatch envelope, and `:129-134` documents them as separate surfaces. `packages/core/src/swarm/launch.ts:632-633` stamps the manifest entry, and that is the only production `github_auth_mode` write in the dispatch path: `git grep -n github_auth_mode 8bb1e528 -- packages/core/src/swarm packages/core/src/orchestration` returns `launch.ts:633` and nothing else. There is no dispatch-envelope writer, so manual and non-manifest dispatches carry no mode label at all. `:432` lets a worker read either surface, which makes the manifest operationally sufficient for manifest-launched workers — it does not satisfy the conjunction.

[...]

Issue #3703, opened 2026-08-24
bug(swarm): verify:routing is a fail-closed pre-dispatch gate that nothing invokes


MScottAdams:
`verify:routing` is described as a **pre-dispatch gate** — "fails when a dispatched worker role has no decision in `.deft/routing.local.json`." It is fail-closed, it is implemented, it has tests, and **nothing ever invokes it.**

AGENTS.md requires the decision it guards:

> **Deliberate model routing:** Before ANY sub-agent dispatch (cohort OR single), make a deliberate per-`worker_role` routing decision via `task verify:routing` / `task swarm:routing-set` — never silently inherit the parent model.

So the rule is stated, the mechanism is built, and the two are not connected.

## Measured at `8bb1e528`

`verify:routing` appears in **no** `Taskfile.yml` target, **no** hook under `.githooks/`, and **no** workflow under `.github/workflows/`. It runs only when an agent remembers to type it.

`DEFAULT_GATED_ROLES = ["leaf-implementation"]` (`packages/core/src/swarm/routing-verify.ts:30`), widened only by an explicit `--roles a,b` (`:99`).

`.deft/routing.local.json` exists in this clone, so a decision was recorded at some point — but nothing checked it against any dispatch.

## How it failed in practice

On 2026-08-24 a single session dispatched roughly forty sub-agents: 35 critic posts across four models, plus builders and pollers. **`verify:routing` was invoked zero times.** No gate fired, no warning appeared, and the omission surfaced only because a human asked which models were being used.

The routing decisions themselves were deliberate and are recoverable from the `model:` comment lead that #3628 and #3646 landed. That is the saving grace and it is *post-hoc* — the rule asks for a pre-dispatch decision, and nothing verified one existed.

## Two distinct gaps

**1. No invocation point.** A fail-closed gate that is never called is documentation. This is the same structural class as #3661 (a mechanized consumer with no producer obligation) and #3672 (a MUST with no completion criterion), except sharper: here a *working* fail-closed implementation exists and is simply not wired to anything.

**2. The gated role set is narrower than the rule.** AGENTS.md says "ANY sub-agent dispatch." The gate defaults to `leaf-implementation` alone. Critic dispatches — 35 posts in one day, across four different models, each a deliberate routing choice — are not in the default set and would not have been caught even if the gate had run.

## Why an invocation point is hard, and worth saying up front

[...]

Issue #3702, opened 2026-08-24
docs(triage): the triage skill never mentions the design-critique gate [NOT IMPLEMENTED - superseded]


MScottAdams:
The `deft-directive-triage` skill contains **no reference to design-critique at all**. An agent completing a triage lean is never told the gate exists.

Verified at `9a4dff79`: `git grep -l design-critique -- content/skills/deft-directive-triage/` returns nothing.

## Why this is separate from #3579

#3579 is about **auto**-triage stamping the ADR-005 marker on write-backs it emits. It is blocked on #1423, which is an unbuilt RFC with no comment emitter in the tree.

This is the **manual** path, and it is not blocked on anything. An operator or agent running the triage skill today finishes with a lean and no prompt to ask whether it is mechanism-shaped.

## What is missing

`content/commands.md:566` already documents the pairing: when an author stamps a lean as mechanism-shaped, write both the `mechanism-shaped: true` body field and the mirrored `design-critique:mechanism-shaped` label, then a separate clearance line `design-critique: warranted | not warranted, because …`.

The design-critique skill is in the Skills Index with triggers including `mechanism-shaped triage`. So both ends exist. **Nothing connects them at the point of use.**

The gate is advisory by design — `verify:judgment-gates --enforce` is deliberately unused, and no marker means the gate never fires. That makes discoverability the entire mechanism. A rule that only fires when someone remembers it, in a skill that never mentions it, fires when someone remembers it.

## Evidence it is not firing on its own

Sixteen design-critique arcs ran on 2026-08-24. Every one was opened because an operator decided to, not because a triage prompted it. **Not one triage that day recorded `not warranted`** — the negative case the convention explicitly permits was never exercised, which is what you would expect if the question is never asked.

## Fix

A pointer, not an essay. At the point the triage skill produces a lean, ask whether it is mechanism-shaped; if yes, stamp the field and mirrored label and record the clearance line; then point at the design-critique skill by name for the arc itself.

Copy no rules. `content/commands.md` and `content/contracts/design-critique.md` are the normative homes and the skill should reference them, consistent with how the thin router already works.

## Acceptance

- [ ] The triage skill names the mechanism-shaped question at the point a lean is produced
- [ ] It states the write-back pairing (body field plus mirrored label) and the separate clearance line, by pointer rather than by copying rule bodies
- [ ] It names the design-critique skill as the next step when the clearance is `warranted`
[...]

PR #3705, opened 2026-08-25
Inject identity-bound worker credentials at spawn

MScottAdams:
## Summary

Dispatchers that already hold a user-bearing worker credential now validate it with the existing principal check, inject it into the worker process environment, and stamp the expected GitHub login so identity comparison is not optional.

- `prepareWorkerCredentialInjection` binds an identity, not merely a token (`DEFT_EXPECTED_GITHUB_LOGIN` / envelope `expected_github_login`).
- App-installation credentials still fail closed (`installation_identity_unverifiable`). Installation identity remains #3693.
- Token values stay in process env only — never in prompts, transcripts, or launch manifests.
- `task swarm:launch` auto-validates only when the probed mode is already `injected-token` and a token is held. Ambient `GH_TOKEN` on `host-gh` is not auto-promoted (that path is the maintainer workaround).

## Related Issues

Tracking: #1351

Product change only. The active xBRIEF is parked for a leftover `scope:complete` land after merge so provenance can pass on this PR.

## Checklist

- [x] `/deft:change <name>` — N/A (scoped story #1351 with existing xBRIEF; one surface)
- [x] `ROADMAP.md` — N/A (#1351 is not a ROADMAP-tracked item)
- [x] Tests pass locally (`packages/core/src/swarm/launch.test.ts` 15/15)

## Test plan

- [x] Helper injects token + expected login on grok-build and local-hybrid when a user-bearing credential validates
- [x] Missing token on an injected-token write path is BLOCKED with the named credential and dispatcher-side remedy
- [x] Installation identity fails closed; no host-gh fallback and no continue-under-host-identity
- [x] Envelope and launch manifest never contain a token value
- [x] Launch does not auto-promote ambient `GH_TOKEN` when the probe is `host-gh`

## Post-Merge

- [ ] `task verify:orphan-active -- --issue 1351` exits 0
- [ ] `task scope:complete` (leftover land PR if needed — never hand-write `xbrief/completed/`)
- [ ] `task verify:completed-tracked -- --issue 1351` exits 0 on `origin/master`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Identity-bound worker credential injection at spawn (#1351).** A dispatcher that holds a user-bearing worker credential validates it with the existing principal check, injects it into the worker process environment, and stamps the expected GitHub login so comparison is not optional. Installation credentials still fail closed. Closes #1351.
+
 - **Land leftover completed-tracked artifact for #3650 (#3264 / #1358).** The #3650 xBRIEF stayed untracked after squash of PR 3706. Moved to `xbrief/completed/` via `scope:complete`. Does not reopen or recut that issue. Refs #2321, #3476.
 
 - **Land leftover completed-tracked artifact for #3674 (#3264 / #1358).** The #3674 xBRIEF stayed untracked after squash of PR 3698. Moved to `xbrief/completed/` via `scope:complete`. Does not reopen or recut that issue. Refs #2321, #3476.
--- a/content/templates/agent-prompt-preamble.md
+++ b/content/templates/agent-prompt-preamble.md
@@ -130,6 +130,7 @@ When present, document the fields in a separate `## Runtime and GitHub auth mode
 
 - `runtime_mode`: one of `local-unsandboxed`, `cursor-native-sandbox`, or `cloud-headless` -- the execution envelope the worker runs in (from the read-only runtime probe, #1557a).
 - `github_auth_mode`: one of `host-gh` or `injected-token` -- which GitHub credential rule applies to this worker (#1557b).
+- `expected_github_login`: the bound user login when the dispatcher injected a user-bearing credential (#1351 / #3665). Login only -- never a token value. Omit when `github_auth_mode` is `host-gh` and no injection occurred.
 
 Launch-manifest entries (#1387 C2 contract) carry the same two fields at the top level alongside `allocation_context`. Workers MUST read the dispatch envelope (or launch manifest) and apply the identity-separation rules in §8 according to `github_auth_mode`, not the historical one-size-fits-all injected-token default.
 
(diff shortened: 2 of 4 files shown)


MScottAdams:
## Review residual (confidence 4/5, floor 5)

Greptile’s remaining P1 on `swarmLaunch` discarding `spawnEnv` is a contract mismatch, not an incomplete bind.

`task swarm:launch` is PREP only (#1387 C2). It validates a held user-bearing credential, stamps `expected_github_login` on a **token-free** manifest, and stops. It does not spawn workers. The spawn-time helper is `prepareWorkerCredentialInjection`: orchestrators apply the returned `spawnEnv` to the worker **process environment** only.

Persisting `spawnEnv` on the launch manifest (or in launch stdout) would put a credential value in an artifact — the #1351 prohibition. Making `swarmLaunch` spawn would change the launch contract.

The earlier named holes are closed on this HEAD:
- missing-token injected-token launch fail-closes
- validation and spawn use the same selected token (all gh token slots stamped; `GH_CONFIG_DIR` dropped)

I will not lower `minGreptileConfidence`. Residual is **class B / conf_floor**. Further passes cannot raise 4 → 5 without violating a prohibition.

MScottAdams:
## Operator acceptance — Greptile residual accepted, floor overridden for this PR

Greptile is 4/5 against a dogfood floor of 5, with 0 P0/P1 and CI ready. The operator has accepted the residual and authorised merge. Recording the reasoning so it is checkable.

### The finding asks for something two contracts forbid

Greptile wants `swarmLaunch` to deliver `spawnEnv` to a worker. There are only two ways to do that, and both are prohibited:

**Launch spawns.** Forbidden by #1387 C2 — `task swarm:launch` emits the manifest and stops; the spawn itself is agent-driven via the platform adapter.

**The credential rides the manifest or stdout.** Forbidden by #1351 acceptance clause 5, and now by this PR's own `⊗` in preamble §2.75: *"Place a credential value in the dispatch prompt, a transcript, or a launch-manifest entry. `spawnEnv` is process-env only."*

The worker declined both and declined to lower the floor. That was correct on all three counts.

### The loop is closed, just not inside launch

Preamble §2.75, landed in this PR, names the consumer: the dispatcher calls `prepareWorkerCredentialInjection` before spawn and applies the returned `spawnEnv` to the worker **process environment** — `GH_TOKEN` plus `DEFT_EXPECTED_GITHUB_LOGIN`. Process env is the correct home for a credential precisely because it is not an artifact.

### The steelman, which is stronger than "the reviewer is wrong"

The only thing consuming `spawnEnv` is a prose instruction to an agent. Prose instructions demonstrably do not always fire — that is the finding behind #3702, where a skill never mentions a gate, and #3703, where `verify:routing` is a fail-closed gate nothing invokes.

So Greptile has identified a **real weakness**. It is not a defect in this change; it is an architectural consequence of launch being PREP-only and credentials being barred from artifacts. It cannot be closed inside #1351 without violating one of the two contracts above.

**Recorded as a residual and routed to #3663**, which already owns re-checking the credential proof inside the worker envelope. That is the natural home for verifying an injection actually happened.

### What this override is and is not

An operator accepting a specific documented residual on a specific PR, with reasoning recorded. It is **not** a change to the floor, the gate, or any evaluator. `plan.policy.review.minGreptileConfidence` stays at 5 and was not touched.

For calibration: two review batches moved this from 2/5 to 4/5 by fixing real defects — injected-token launch with no token now fail-closes, and the same token is stamped into every `gh` token slot with `GH_CONFIG_DIR` dropped so a store cannot bind another identity. The remaining point is the contract fight above, not unfixed review debt.

Issue #3713, opened 2026-08-25
feat(feedback,scm): allow BLOCKER in the title for consumer hard-stops -- the one sanctioned title classification


MScottAdams:
> **Recut after a refutation-variant critique and an operator override.** The original filing argued for a title marker on the grounds that consumers cannot apply labels. The critic ([5405053415](https://github.com/deftai/directive/issues/3713#issuecomment-5405053415)) broke that reasoning — see the withdrawn citations below — but the operator overrode the disposition on separate grounds. Trail: lean [5405190016](https://github.com/deftai/directive/issues/3713#issuecomment-5405190016), verified claims [5405190379](https://github.com/deftai/directive/issues/3713#issuecomment-5405190379), revised lean with override [5405233019](https://github.com/deftai/directive/issues/3713#issuecomment-5405233019).

## Problem

A consumer hitting a Directive adoption blocker has no way to say so where a maintainer will see it. They cannot apply `adoption-blocker` — GitHub requires push access to set labels at issue creation, and labels are silently dropped otherwise. The label is the ranking mechanism and it is reachable only by people who already have commit rights.

The result is that the reports that matter most arrive indistinguishable from everything else.

## This convention already exists

It is not being invented here:

- **27 issues** in this repo already carry a BLOCKER-ish title token (`search/issues?q=repo:deftai/directive+in:title+BLOCKER`).
- **#3699** is open because ten of them are unlabeled and invisible to `triage:queue` ranking.

People reached for the title because it was the only surface they controlled. #3650 then prohibited the practice without asking whether the population it served could use the alternative.

## Operator decision

`BLOCKER` in the title is permitted for the consumer hard-stop class, and is **the only classification allowed in an issue title**. Every other classification stays label-only.

Rationale recorded by the maintainer: finding hot issues that consumers drop is essential, and no other surface delivers it. A maintainer scanning an issue list sees titles without opening anything; ranking only helps once someone with push access has already classified the issue.

Token is `BLOCKER`, matching the 27 existing issues, so nothing needs migrating.

## Title and label are two stages, not two options

#3650 treated them as competing. They are sequential:

| Stage | Surface | Who | Purpose |
|---|---|---|---|
| Inbound flare | `BLOCKER` in title | The filer, no push access needed | Make it visible in a list |
| Ranked state | `adoption-blocker` label | Privileged actor, after the #3706 body-evidence test | Enter `triage:queue` ranking |

The label remains the authority. The token is how a report reaches someone who can apply it.

[...]

Issue #1309, opened 2026-05-22
fix(templates,agents-md,tests): propagate welcome / WIP cap / session-start ritual / skill routing from maintainer AGENTS.md into templates/agents-entry.md + propagation-discipline gate


MScottAdams:
## Problem

The maintainer's `directive/AGENTS.md` has grown with new rules over recent releases (#1149 session-start ritual, #1152 umbrella current-shape convention, #954 multi-agent orchestration discipline, #1242 CHANGELOG entry style, plus the welcome/WIP/triage discoverability surface from #1119 / #1143 / #1122 / #1124). `templates/agents-entry.md` -- which is the source-of-truth for every consumer-side AGENTS.md managed-section block -- has NOT been kept in lockstep.

Net effect: a consumer who installs deft (or upgrades to the latest) gets a much narrower AGENTS.md than the maintainer has. A consumer agent reading their project's AGENTS.md has no idea that `task triage:welcome`, the WIP cap surface, the session-start ritual, or `task doctor` exist as canonical commands. The mechanism to propagate is sound (`cmd_agents_refresh` byte-replaces the managed-section from the template); the bug is that the template hasn't been updated.

## What's in the maintainer file but NOT in `templates/agents-entry.md`

- `## Session-start ritual (#1149)` -- the canonical ritual (alignment confirmation -> `task doctor` -> branch-policy disclosure -> `task triage:welcome` -> `task verify:cache-fresh`).
- `task triage:welcome` -- the consolidated onboarding/status command (#1143; default mode subsumes `task triage:summary` plus first-time/incomplete nudge -- see Resolution below).
- WIP cap surface (`plan.policy.wipCap` default 10; cap-reached relief commands).
- `## Cache-as-authoritative work selection (#1149)` -- the `task triage:queue` rule for "what's next?".
- `task doctor` -- the consolidated install-integrity + toolchain + AGENTS.md-freshness probe (#1308).
- `## Skill Routing` -- the keyword-to-skill routing table (welcome / refinement / swarm / triage / build / cost / pre-pr / etc.).

## Why this matters

The consumer has access to the TOOLS via the Taskfile (assuming the consumer's project Taskfile includes the deft Taskfile per the canonical pattern), but the AGENTS.md doesn't TELL them to use any of it. A consumer agent never emits `task triage:welcome` at session start because no rule in their AGENTS.md prescribes it. The maintainer-side discipline is rich; the consumer-side is bootstrap-only.

[...]

PR #3724, opened 2026-08-25
docs(agents): propagate the BLOCKER title classification to consumers (#3713)

MScottAdams:
Lands the #1309 consumer propagation for #3713, which missed the squash of PR #3720.

PR #3720 merged as `bfa66731` with all three maintainer-side prohibition sites amended, but commit `e3f21e2e` -- the `content/templates/agents-entry.md` mirror -- was not in the squashed range. Greptile held #3720 at 4/5 and its P2 asked for exactly this propagation, so this also clears that finding.

Delta is the propagation only: two lines in `agents-entry.md` and the `agents:refresh` managed-section update in `AGENTS.md`.

The propagated rule carries the same constraints the maintainer-side text does: `BLOCKER` is the only classification permitted in a title, `feedback:file --blocker` writes it beside `[framework-gap]`, no auto-mirror from consumer-authored title text to `adoption-blocker`, and absence of the token does not mean "not a blocker."

Refs #3713, #1309, #3706.


MScottAdams:
Superseded by a clean cut off current master -- this PR was opened from a stale branch whose commits were already squashed in, so it showed 10 changed files and failed its merge gate. Replacement carries the same 2-line propagation.

Issue #3422, opened 2026-08-17
feat(swarm,review-cycle): forge-outage drop-back + human report + 30m default retry (USER.md / project override)


MScottAdams:
## Summary

Autonomous / YOLO / `drive-to: merge-ready` workers still **thrash the forge and burn tokens** when GitHub (or another SCM API) is in a **partial or major outage**. They keep polling, re-pushing, empty-committing, close/reopening, and retrying `gh`/`ghx` against a dying API.

Landed work covers **CI weather classification** and **status-page attribution**. It does **not** encode a **worker drop-back**: stop GitHub I/O, **report the outage to the human once**, then **probe again on a timer**.

**Default retry: 30 minutes.** Override from **USER.md Personal** (always wins) or **`plan.policy` in PROJECT-DEFINITION**.

## Why this is not #3167 / #3180

| Issue | What it shipped | Gap that remains |
|-------|-----------------|------------------|
| #3167 | `pr:watch` / `pr:merge-ready` weather reason codes + babysit thrash caps (1–2 empty-commit / close-reopen then BLOCKED) | Caps a **CI-holdout babysit loop**. Does not stop **general API death** (Issues, PRs, REST 20% errors, archive/raw 50%, webhooks). No human report + scheduled re-probe. |
| #3180 | Status-page probe + `attribution: platform \| capacity \| repo_config \| unknown` for CI weather handoffs | Attribution tells the agent *why* CI is empty. Remediation is still “wait + re-check HEAD.” No **interval**, no **USER.md / project override**, no **must-report-to-human**. |
| Dual-stop #2442 / #3273 | Iteration / budget / no-progress stop | Stops *product* loops, not *platform* I/O. |
| #694 (open) | Prefer reactive retry over proactive sleep for **rate limits** | Rate-limit pattern. Outage drop-back is a different trigger and a much longer default. |

`content/scm/github.md` § Platform status probe (#3180) already says: after thrash caps, stop empty-commit loops and wait. Agents still invent their own cadence (or none) and often keep beating the API.

## Observed (2026-08-17)

GitHub **Partial System Outage** (incident `zkxwbgr0cnmx`):

- ~20% errors on web + API
- ~50% errors on archive/raw
- PRs, Issues, API, Actions, Webhooks degraded
- Copilot `major_outage`

Operator standing order in that session: do not send the human to the web UI; if the API dies, **do not thrash** — wait and retry. Live cohort used **15 minutes** as a session override. Product default requested here is **30 minutes**.

[...]

PR #3744, opened 2026-08-25
File writes no longer wait on triage-cache freshness (Closes #3738)

### LINKED ISSUE #3738: bug(hooks): cache_fresh is in the write gate's required set and does live forge I/O to satisfy it (80.4s measured on a clean cache)
## Summary

`cache_fresh` is in the write gate's required set, and it performs live forge I/O to satisfy itself. Triage-cache drift is a **work-selection** precondition, not a **write-authorization** one. The cheapest correct fix is not to make that step fast — it is to remove it from the set of things a write must prove.

Measured at **80.4 seconds on a clean cache with no drift**, against a 30s Cursor deposit budget whose green baseline is 7.9s.

Split from #3736. Recut after round-1 open critique, which found all three of my original directions wrong as stated; see History.

## Measurements

| Condition | Cost | Notes |
|---|---|---|
| Green write path, `cache_fresh` satisfied | **7.9s** | ~22s headroom under the 30s budget |
| `verify:cache-fresh --work-selection`, probe live, **no drift** | **80.4s** | exit 0 — nothing was wrong |
| Injected non-green `cache_fresh` | 35.7s | deny, `ritual-not-ready` |
| Unplanned production hit | ~115s | after closing four issues; recovery `cache:fetch-all` took 142.1s |

The 80.4s figure is the important one and it was not in the original filing. **The expensive path is the clean one.** A fix aimed only at the drifted branch would leave the common case broken.

## What actually arms it

`shouldRearmDriftProbe` re-arms `cache_fresh` when the recorded `drift_probe` is `skipped-no-work-selection` and `detectWorkSelection` returns `inPlay`; `hasActiveStoryArtifact` is true as soon as `xbrief/active/*.xbrief.json` exists.

**`scope:activate` arms it.** Story start, not merging. A `drive-to: merge-ready` worker meets it at the *beginning* of its unit of work, and every subsequent write is exposed.

## Correction: the gate already fails open

The original filing said fail-open was "explicitly not proposed" on the grounds that it would convert resource exhaustion into an authorization bypass. **That bypass already exists on this path.**

```581:587:packages/core/src/preflight-cache/evaluate.ts
    const drift = runDriftProbe(resolvedRepo, cacheRoot, source, options);
    if (
      drift !== null &&
      (drift.stateDriftNumbers.length > 0 || drift.contentDriftNumbers.length > 0)
    ) {
      return formatDriftFailure(resolvedRepo, drift);
    }
```

`runDriftProbe` catches every exception and returns `null` (`evaluate.ts:358-372`). A `null` result fails the `drift !== null` test, so the gate **passes**. A swarm that exhausts the REST bucket, a 429, or a forge outage silently satisfies `cache_fresh`.

So the design question is not *whether to introduce* fail-open. It is *where the existing one should live and whether it should be visible*. Note the collision: bounding the probe (original direction 3) would **remove** a fail-open that #3422's forge-outage posture arguably wants.

## Remedy

**Remove `cache_fresh` from the write gate's required set, and strip the write gate's execution authority for every gated step except `agent_hooks`.**

`GATED_STEPS = ["agent_hooks", "doctor", "cache_fresh"]` is one flat list with no per-surface required set. That flatness is the defect: it makes a write prove a work-selection precondition. Splitting the set answers the question the triage posed — *what must be true for this write versus for this session* — and subsumes the three directions below rather than choosing among them.

The write gate MAY execute `agent_hooks`, which is deliberately non-cacheable and is what the 30s budget was sized around. It MUST NOT execute any other gated entrypoint.

## History: why the original directions were wrong

Recorded because each failure is instructive, not to pad the issue.

1. **"Move the drift probe to lifecycle boundaries"** — already built. `session:start` orientation runs the probe and records the verdict into `gated_steps` precisely so the gated verify can skip it. The write path executes it only as a *repair attempt* after the recorded verdict goes red. Direction 1 binds as **already satisfied**, which relocates the question rather than answering it.
2. **"Use `inspectSessionRitual` on the write path"** — over-corrects. It executes nothing, silently dropping the deliberate live `agent_hooks` re-probe. `session:ready` already composes the correct shape.
3. **"Bound the synchronous I/O"** — cannot stand alone. Against 22s of headroom and an 80.4s observed probe, any bound that fits would fire on essentially every probe, converting an intermittent timeout into a permanent deny. It binds as a **backstop composed after** the probe leaves the path, never as the fix.

Also sharpened: the unbounded element is a server-chosen `sleep` plus subprocess fan-out, not generic latency, so a bound placed at the wrong layer will not hold.

## Acceptance

Replaces the original criterion, which conditioned on a drifted cache — the rarer branch of the trigger it named.

- **Zero forge I/O on the write path**, asserted through a seam rather than a stopwatch.
- A **rendered decision** inside the deposit budget on the **clean-cache re-arm** path — the common case after `scope:activate`, not the drifted one.
- `agent_hooks` still executes on the write path; no readiness regression. A test that passes because nothing runs is a failure.
- The existing forge-error fail-open is either preserved deliberately with a recorded rationale, or replaced with a **visible** outcome. It must not remain silent by accident.
- No gate weakened to reach the number (#3156).
- CHANGELOG `[Unreleased]` entry.

## Sequencing

**This should land before #3736.** With `cache_fresh` off the write path the green baseline is 7.9s inside a 30s budget, so #3736's redundant git spawns sit inside a budget that already fits. #3736 stays split — it is real and independently correct — but it is the smaller term once this lands.

Round-1 critique: [5417129709](https://github.com/deftai/directive/issues/3738#issuecomment-5417129709). Split from #3736. Related: #3739, #3732, #3422 (forge outage posture), #3246.

MScottAdams:
## Summary

File writes no longer wait on triage-cache freshness. The write gate still re-checks hook readiness and still reads the recorded doctor step. It does not require or execute cache_fresh.

Triage-cache drift stays a session and work-selection precondition. scope:activate no longer arms an 80s forge probe on every Write.

## Remedy

- Split the flat GATED_STEPS list by surface.
- Write required set: gent_hooks, doctor (record only).
- Write execute set: gent_hooks only.
- Session/full gated verify still uses the full set and still re-arms cache_fresh when work selection appears.

Rejected directions were not reintroduced: the lifecycle probe already exists; inspectSessionRitual on write would drop live gent_hooks; bounding I/O alone would turn an intermittent timeout into a permanent deny.

## Fail-open decision (#3738 / #3422)

Preserved, with a recorded rationale. 
unDriftProbe still returns null on throw; evaluate still passes when drift === null. Converting this to fail-closed would deny session/work-selection on 429/502/DNS and collide with #3422 (local work MAY continue when it does not need the forge). A visible probe-errored verdict is left as a follow-up (decision revisit trigger). Record: xbrief/decisions/2026-08-25-3738-preserve-cache-fresh-forge-error-fail-open.decision.json.

## Related Issues

Closes #3738

## Test plan

- [x] Write-gate tests exercise the clean-cache re-arm path (drift_probe === skipped-no-work-selection + work selection in play) and assert erify:cache-fresh is not executed
- [x] Dispatcher test renders write-ready without executing erify:cache-fresh or doctor
- [x] Red recorded cache_fresh does not deny a write
- [x] Red recorded doctor still denies without executing doctor
- [x] gent_hooks still executes on the write path
- [x] Evaluate still fails open when the drift probe throws
- [ ] CI 	ask check green
- [ ] Greptile Confidence Score 5/5 on the review comment (not only the check-run conclusion)

## Checklist

- [x] /deft:change <name> — N/A: one issue, one surface (write-gate required/execute split). Not a new cross-cutting proposal.
- [x] ROADMAP.md — N/A (bug, not a tracked roadmap item)
- [x] Tests pass locally (targeted write-gate + evaluate: 51 passed). Full local 	ask check previously failed on unrelated DEFT_ROOT leakage and load timeouts; CI is the merge chokepoint.


MScottAdams:
ended_at: 2026-08-26T00:17:45.696Z

Issue #3739, opened 2026-08-25
bug(hooks): a post-ritual deny on a ready consumer path renders 3-4.6s after the nested 5s budget has already elapsed


MScottAdams:
## Summary

On a **ready consumer mutation path**, a nested host's deny envelope is produced 3 to 4.6 seconds *after* its 5-second deposit budget has already elapsed. The host kills the hook before Directive's verdict exists on stdout, so the denial is computed and never delivered.

Nested hosts (Claude, Grok, Codex) get `timeout: 5`. The framework's own source says 5s is below the work's ceiling — and says it while raising *Cursor* to 30s for exactly that reason.

Split from #3736. Claim scoped after round-1 critique; see Scope below for what this issue does **not** claim.

## The asymmetry, in the framework's own words

```ts
// packages/core/src/init-deposit/agent-hooks.ts
/**
 * ... Live readiness alone has a multi-host fixture ceiling of ~24s
 * (`content/contracts/agent-hook-readiness.md`); the historical deposit
 * default of 5s was below that budget. ...
 */
export const CURSOR_TOOL_BEFORE_TIMEOUT_SECONDS = 30;

/** Nested Claude/Grok/Codex command-hook default timeout (seconds). */
export const NESTED_HOOK_TIMEOUT_SECONDS = 5;
```

The diagnosis was correct and the remedy reached one host of four. `failClosed: true` likewise appears only in the Cursor block.

## Measured: post-ritual consumer denials miss the budget

Method: force the consumer branch of `evaluateAgentHookReadiness` by pointing `DEFT_ROOT` at the installed framework package for the child process only, keeping this checkout's green ritual and deposits; then require a denial *downstream* of the live gated ritual via `DEFT_SESSION_SLASH_VERB=/github-issue` (intent ceiling). Same Write payload throughout.

| Host | Critic run | Independent re-run | Verdict rendered | Budget |
|---|---|---|---|---|
| Claude | 8.385s | **8.2s** | structured `permissionDecision: "deny"` | 5s — killed first |
| Grok | 8.087s | **7.8s** | structured `decision: "deny"` | 5s — killed first |
| Codex | 9.633s | **8.3s** | structured `permissionDecision: "deny"` | 5s — killed first |

Two observers, independently. A separate consumer-root live-readiness invocation passed all eight allow/deny fixtures across the four hosts in 5.516s, confirming the probe genuinely ran rather than short-circuiting.

[...]

Issue #3729, opened 2026-08-25
bug(session): the occupancy lease is exclusion-only -- an agent that never claims is never gated


MScottAdams:
## Problem (restated after the round-1 panel; see successor lean 5420853425)

The occupancy write gate fails open when no live lease exists: `if (record === null || liveness !== "live") return { allow: true, ... }`. The original framing of that as "an agent that never claims is never blocked" was refuted -- an actor that never claimed also never ran `session:start`, and the ritual gate already fails closed on a missing `.deft/ritual-state.json` wherever the hook is installed.

What survives the panel is narrower and still real: **the lease and the ritual bind have mismatched lifetimes**, so an actor can hold a valid ritual bind while its lease has lapsed, and the gate then permits every same-user process in that tree.

Measured at `eacd7435900a6583f119b0e055c1c007058d2e1a`, `packages/core/src/session/occupancy.ts`:

| Constant | Value |
|---|---|
| `OCCUPANCY_TTL_MS` | 20 minutes |
| `OCCUPANCY_REFRESH_AFTER_MS` | 5 minutes (TTL / 4) |
| `OCCUPANCY_STALE_WARN_MS` | 15 minutes (TTL * 3/4) |
| `OCCUPANCY_MAX_LEASE_MS` | 12 hours (TTL * 36), per decision record `2026-08-26-bound-the-occupancy-lease-with-an-absolute-12-hour-age-cap-on-cl` |
| `DEFAULT_SESSION_RITUAL_STALENESS_HOURS` (`packages/core/src/policy/index.ts`) | 4 hours |

**12:1.** A ritual bind outlives its lease by an order of magnitude, so the residual window is an owner that stops writing for 20 minutes while remaining ritually bound for up to 4 hours.

## What this recut changed, and why

Every item below was accepted `accept-into-contract` on the round-1 disposition map (3 critics, empty disagreement set). The author of the original text is the parent of that arc and cleared none of these himself.

- **Struck: "an agent that never claims is never blocked" as the headline.** The unclaimed case is covered by the ritual gate at `dispatcher.ts`, not by occupancy. Replaced with the lifetime-mismatch statement above.
[...]

Issue #3718, opened 2026-08-25
bug(swarm): swarm:launch is unreachable for a freshly-triaged story -- readiness needs 15 hand-authored fields and file_scope collides with the provenance gate


MScottAdams:
## Problem

`swarm:launch` cannot accept a freshly-triaged story. The artifact `triage:accept` produces is rejected by `swarm:readiness` on fifteen fields, no verb or skill bridges the gap, and the one field that would bridge it — `file_scope` — triggers a separate enforced gate that demands a prior merged PR and a human TTY.

The result is that the sanctioned batch-launch path is unreachable for exactly the work most likely to need it: a cohort of newly-accepted issues.

## Reproduction, tonight, six stories

Issues #3666, #3673, #3679, #3690, #3712, #3713. All accepted via `triage:accept`, promoted, and activated.

**First failure** — `swarm:launch --stories …` against `pending/`:

```
Error: could not resolve every cohort member:
  - #3666: no active story references this issue.
```

`swarm:launch` requires `active/`, not `pending/`. Nothing says so before you run it.

**Second failure** — after `scope:activate` on all six:

```
Error: story '…3666…' is not launch-ready -- swarm:readiness gate failed:
Missing fields: plan.metadata.kind=story, plan.id, plan.narratives.ImplementationPlan,
plan.narratives.UserStory, plan.items[].narrative.Acceptance,
plan.metadata.swarm.readiness=ready for concurrent allocation,
plan.metadata.swarm.parallel_safe, plan.metadata.swarm.file_scope,
plan.metadata.swarm.verify_commands, plan.metadata.swarm.expected_outputs,
plan.metadata.swarm.depends_on, plan.metadata.swarm.conflict_group,
plan.metadata.swarm.size, plan.metadata.swarm.file_scope_confidence,
plan.metadata.swarm.model_tier, Traces or missing_traces_justification
```

Every story, all fifteen fields. Meanwhile `task xbrief:preflight` exits **0** on the same artifacts — "ready for implementation."

So two gates disagree about whether the same brief is buildable, and the one that says yes is the one the intent gate uses.

## Nothing produces the metadata

The shape is documented — `content/vbrief/vbrief.md` § Swarm-Ready Story Contract gives the full `plan.metadata.swarm` block. What is missing is anything that writes it for a story-sized issue:

- `triage:accept` emits narratives, items, and acceptance clauses, and no `swarm` block at all.
[...]

Issue #3613, opened 2026-08-21
tracker: multi-agent coordination defects -- occupancy model gaps


MScottAdams:
Tracker for multi-agent coordination defects surfaced on 2026-08-21 while landing #3588, plus
#3730 observed on 2026-08-25 during an overnight swarm cohort. All were observed live, in this
repo, by agents working concurrently -- not reasoned about in the abstract.

Amendment 2026-08-25: restore **identity** (#3611, grouped here on 21 Aug then dropped from the
children list when #3730 was added) and add **obligation** (#3729). #3649 stays related, not a
child -- it is launch sequencing, not a missing dimension of the lease model.

Amendment 2026-08-26 (pass-2): **#3730 moves to Related.** Its round-1 panel established
unanimously that child mortality is a parent observing *another process*, not a session signalling
about itself, so it is not a dimension of this model -- it is an upstream *producer* of #3604.
Current shape comment carries the full pass-2 state, wave order, and the resolved
coordination-vs-authorization fork.

## The common shape

Directive's coordination primitive is the **worktree occupancy lease** (#3433). It is the right
idea with gaps in different dimensions of the same model:

| Dimension | Gap | Issue |
| --- | --- | --- |
| **Liveness** | nothing marks a *live* session alive -- the lease never heartbeats during work, so a session mutating for longer than the 20-minute TTL reads as abandoned | #3599 |
| **Lifecycle** | nothing marks a *dead* session dead -- `releaseOccupancy` is called only on swarm cohort completion, so leases outlive their sessions and accumulate (12 orphans found, oldest ~3.7 days) | #3604 |
| **Keyspace** | the lease is keyed on a worktree path, so it cannot protect the *shared* coordination surface -- GitHub issue threads. Two agents holding valid leases on different worktrees have no interlock on issue #N, and a concurrent comment corrupts an open triage/critique ceiling (#3434 / #3452) | #3607 |
| **Identity** | the owner cannot bind to their own lease -- `evaluateOccupancyWriteGate` only accepts `DEFT_SESSION_ID` from the hook process env, which host-managed direct-write tools do not inherit from `session:start`. The session that just claimed is denied as a stranger. Blocks #3599 (same function). Not Codex-only (also reproduced on Claude Code). | #3611 |
[...]

PR #3751, opened 2026-08-26
chore(xbrief): track queued proposed briefs and one decision record

MScottAdams:
Tracks six proposed briefs generated by `triage:accept` today, plus a policy decision record from 2026-08-24. The repo already tracks 26 briefs in `xbrief/proposed/`, so these were the outliers rather than the convention.

**Queued, not started:** #3598, #3718, #3736, #3739.
**Occupancy cluster, owned separately under #3613:** #3599, #3729.
**Decision record:** `minGreptileConfidence` stays at 5, do not lower.

No lifecycle transition -- all six remain in `proposed/`. Committing them so the v0.107.0 release preflight sees a clean tree rather than needing `--allow-dirty`.

### DIFF
--- a/xbrief/decisions/2026-08-24-plan-policy-review-mingreptileconfidence-stays-at-5-do-not-lower.decision.json
+++ b/xbrief/decisions/2026-08-24-plan-policy-review-mingreptileconfidence-stays-at-5-do-not-lower.decision.json
@@ -1,0 +1,23 @@
+{
+  "schemaVersion": "deft.decision.v1",
+  "id": "plan-policy-review-mingreptileconfidence-stays-at-5-do-not-lower",
+  "decision": "plan.policy.review.minGreptileConfidence stays at 5; do not lower it to reduce residual passes",
+  "governingRule": {
+    "description": "review-surface precedence (#2308)",
+    "path": null,
+    "rfc2119": null
+  },
+  "alternativesConsidered": [
+    {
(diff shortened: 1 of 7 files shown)


MScottAdams:
Merged at Greptile **2/5**, below this repo's 5/5 bar. Disclosing rather than leaving it in the log unexplained.

**Process failure:** I issued the merge and read the score in the same command block, so I saw 2/5 after the merge rather than before it. The score comment is the authoritative surface -- the check-run conclusion reads `success` regardless -- and I know that; I just sequenced the commands wrong.

**On the findings themselves, Greptile is right.** All three are the ingest defect class tracked by #3717:

- #3598's brief derives Markdown punctuation as a seventh acceptance clause -- the same `--` separator promotion that blocked `scope:complete` on #3721 and #3712 today.
- #3599's brief has item titles and clauses truncated mid-sentence.
- #3739's brief has an empty work-item surface.

**Why I am not reverting.** These are `proposed/` planning artifacts. Nothing is active, nothing builds from them, and none of them ships in the release payload. The concrete risk Greptile names -- blocking completion on a meaningless criterion -- is real but deferred to build time, and today established the mitigation: any brief generated at `triage:accept` is stale or malformed by the time its issue is built, and gets regenerated then. I did exactly that for #3738 this afternoon, where the pre-recut brief still carried a refuted remedy.

**Carried forward:** these three are flagged for regeneration before their issues are built, and the independent confirmation is recorded on #3717.

Issue #3742, opened 2026-08-25
tracker: agent-hook enforcement defects -- reachability, authority scope, delivery, cost, reporting


MScottAdams:
Tracker for agent-hook enforcement defects surfaced on 2026-08-25 by turning agent hooks on in this repo and working normally for an afternoon. Every one was observed live — none was reasoned about in the abstract, and several contradicted what reading the code suggested.

All four children have been through a design-critique arc. A fifth issue was filed and **withdrawn** when its critic refuted it.

## The common shape

Directive's pre-execution gate stack (`PreToolUse` → `inspectMutationGates` → gated `verifySessionRitual`) enforces occupancy, scope, intent ceiling and runtime authority. The defects are not in what it decides. They are in **whether the decision is reached, delivered, and reported honestly**.

| Dimension | Gap | Issue |
| --- | --- | --- |
| **Reachability** | Should a maintainer checkout run the gates it ships? `doctor` reports `skip -- maintainer source checkout`; the gated ritual reports `INCOMPLETE` on the same condition. Two surfaces disagree, so the gate sat unreachable here without anyone noticing. | #3732 |
| **Authority scope** | `cache_fresh` is in the write gate's required set and does live forge I/O to satisfy it — 80.4s measured on a *clean* cache. Triage-cache drift is a work-selection precondition, not a write-authorization one. | #3738 |
| **Delivery** | On a ready consumer path a post-ritual deny renders 3–4.6s *after* the nested 5s budget elapsed. Directive computes a verdict it never gets to print. | #3739 |
| **Cost** | Five git subprocesses per write, four redundant, because `evaluateLoadedState` runs twice and nothing memoizes. | #3736 |
| **Reporting** | `verify:hooks-installed` reports `registered and structurally valid` — a claim about a file, presented as a claim about enforcement. | #3732 |

## Why they compound rather than merely coexist

Observed causal chain, all on 2026-08-25:

1. Agent hooks were never installed in this checkout, so the whole stack was unreachable. `verify:hooks-installed` printed a green tick throughout (#3732).
2. On installing them, an ordinary file write was killed at the 30s Cursor budget. The gate's `failClosed: true` renders a timeout indistinguishable from a policy refusal.
3. The cost turned out to be two independent terms that stack: redundant git spawns (#3736) and a live forge read (#3738).
4. Deleting the deposits did not disable anything — the host had already loaded them — and it *broke* the ritual, denying every write. Recovery required re-depositing through the shell, which is ungated only by accident.
[...]

PR #3776, opened 2026-08-26
Keep a working session's worktree lease alive, and bound it (#3599)

### LINKED ISSUE #3599: fix(session): occupancy lease never heartbeats during active work, so a working session gets stolen (#3433)
## What happened

A mutation session claimed the occupancy lease on a worktree, worked for ~3.5 hours, and
produced three pushed commits. Its `heartbeat_at` never advanced past the original claim. A
concurrent agent correctly read that lease as abandoned and stole it.

```json
{
  "session_id": "285342a2-…",
  "intent": "mutation",
  "claimed_at":   "2026-08-21T13:24:13Z",
  "heartbeat_at": "2026-08-21T13:24:13Z"   // never advanced, across 3 commits
}
```

Nothing was lost here because the work was already pushed, but the steal is a data-loss
mechanism in general: two sessions mutating one worktree is exactly what #3433 exists to
prevent.

The stealing agent did nothing wrong. The lease genuinely looked dead.

## Defect 1 — the write gate discards a liveness signal it already has

`evaluateOccupancyWriteGate` (`packages/core/src/session/occupancy.ts:394`) runs on **every**
product-path write. It resolves the live occupant, compares session ids, and returns `allow`:

```ts
const live = liveOccupant(projectRoot, now);
if (live === null) return { allow: true, message: null, occupant: null };
const incoming = input.sessionId?.trim() || (input.env ?? process.env).DEFT_SESSION_ID?.trim() || "";
if (incoming.length > 0 && incoming === live.sessionId) {
  return { allow: true, message: null, occupant: live };   // <-- no heartbeat
}
```

So a session that is *actively writing files* — the strongest available evidence of liveness —
never refreshes its own lease. Liveness is judged on a wall-clock timestamp that nothing
advances while work happens.

The same-session branch is the natural place to re-stamp: it already knows the identity
matches, it already runs on the cadence of real work, and it costs one write on a path that is
already writing.

## Defect 2 — no way to refresh through the documented surface

`task --list` exposes exactly one occupancy verb:

```
* occupancy:steal:   Supersede a live worktree occupancy lease (#3433). Flags: --confirm --occupant <id>
```

There is no `occupancy:heartbeat` / `occupancy:renew`. Refreshing requires reverse-engineering:

1. `resolveOccupancySessionId` (`occupancy.ts:106`) consults `DEFT_SESSION_ID`
2. `applyWorktreeOccupancy` (`occupancy.ts:140`) treats an incoming id equal to the live
   occupant as a heartbeat rather than a denial
3. therefore `DEFT_SESSION_ID=<existing-id> task session:start` re-stamps

No agent or operator will discover that from the exposed surface. In practice **every**
long-running mutation session goes stale and becomes a steal candidate, which makes the
mechanism actively misleading: `occupancy:steal` exists and works, while staying alive does not.

## Why this matters more than a single lost lease

The lease is the only thing standing between two agents and concurrent mutation of one
worktree. If the holder cannot signal liveness, the gate degrades into "whoever ran most
recently wins," and the failure is silent on both sides — the holder gets no warning that its
lease expired, and the stealer gets no signal that the occupant is mid-work.

## Proposed fix

- [ ] Re-stamp `heartbeat_at` in `evaluateOccupancyWriteGate`'s same-session branch (guard
      against write amplification — e.g. only when the beat is older than some fraction of the
      staleness window)
- [ ] Expose `occupancy:heartbeat` (or `occupancy:renew`) as a task/CLI verb so refresh is
      discoverable without reading source
- [ ] Document the refresh path in `.deft/core/commands.md` § Session routing alongside
      `occupancy:request` / `occupancy:steal`
- [ ] Warn the holder when its own lease is within the staleness window rather than letting it
      expire silently
- [ ] Consider requiring `occupancy:steal --confirm` to report the occupant's last *write*
      time, not just heartbeat age, so a mid-work occupant is visible to the stealer

## Repro

1. In a worktree, `task session:start` (mutation intent) — note `heartbeat_at`
2. Make and commit file changes for longer than the staleness window
3. Re-read `.deft/occupancy.json` — `heartbeat_at` is unchanged
4. From another session, the lease reads as expired and `occupancy:steal` succeeds against an
   actively-working holder

Found while landing #3588. Refs #3433, #3110, #3597.

## Added requirement -- absolute lease-age cap (operator-approved 2026-08-26)

Arrived after this arc bound, from the #3755 round-2 panel. It is a direct consequence of this issue's own fix and belongs in the same change.

**Requirement: an absolute lease-age cap keyed on `claimed_at`, independent of refresh.** No amount of writing may extend a lease past a hard bound.

### Why

Refreshing `heartbeat_at` on owner-allow writes is correct in isolation. The problem is the admission model it lands on: occupancy admits **anyone presenting the occupant's session id**. The #3755 round-2 panel established unanimously that the lease is a bearer-string boundary, not a lineage. So "an owner that writes at least once per TTL never expires" really means *any process holding the string* holds the lease indefinitely -- and TTL crash recovery, which is what rescues a worktree from a dead session, stops working.

Measured before this change lands: a round-2 critic held a lease open across **twelve heartbeats over two hours with `claimed_at` frozen**. This fix makes that shape structural rather than incidental.

Verified against master: `liveOccupant` expires on heartbeat age alone. `claimedAt` is stored and correctly preserved across refreshes (`occupancy.ts:209`), but **nothing bounds it**. The module header at `occupancy.ts:13-14` mentions a "hard age cap"; none is implemented on this path.

### Design notes, not prescriptions

- The cap belongs where liveness is decided, so a lease past the bound reads as expired regardless of how recently it was refreshed -- `isOccupancyExpired` / `liveOccupant` is the natural site.
- Choose the bound deliberately and record the reasoning. It must be comfortably longer than a legitimate long-running session -- agent sessions here routinely run hours -- while short enough that a dead session's tree is reclaimable the same working day.
- `claimedAt` must keep surviving refreshes, or the cap is trivially defeated by the heartbeat it exists to bound. **Pin that with a test**: it is the single line whose regression silently disables the whole mechanism.
- Consider whether deny and warning text should distinguish "stale heartbeat" from "past the absolute cap" -- they call for different operator responses.

### Additional acceptance

- [ ] A lease refreshed continuously **past** the cap reads as expired
- [ ] A lease refreshed continuously **under** the cap stays live -- no regression to the refresh fix
- [ ] `claimed_at` survives a refresh, pinned by test
- [ ] The cap appears in the CHANGELOG line -- user-visible behaviour, not an internal detail

MScottAdams:
## What this fixes

`evaluateOccupancyWriteGate` was read-only on the owner-allow path, so the one event that proves a session is alive — its own gated write — never extended its lease. The 20-minute occupancy window ran from the claim, once, however long the session worked. A session that worked for hours looked abandoned, and a peer that stole its worktree was doing exactly the right thing.

## What changed

**The owner's writes keep the lease alive.** A gated write by the owner re-stamps `heartbeat_at`, floored at a quarter of the TTL so the lease file is not rewritten per event. Refresh is opt-in at the API (`refresh: true`) and the dispatcher turns it on only for the final evaluation before an allowed write, so the stamp records a write that really happened rather than a probe.

**`last_write_at` is recorded separately from the heartbeat.** `heartbeat_at` moves on any lease touch; `last_write_at` only on a gated product write. That distinction is what tells a would-be stealer whether the occupant is mid-edit or merely claimed and walked away, so `occupancy:steal` reports it *before* the steal rather than leaving heartbeat age to stand in for activity.

**The holder hears about its own staleness.** Inside the staleness window (three quarters of the TTL) the write gate says so on the allow, instead of the owner discovering it went stale only when a peer takes the lease.

**A discoverable refresh verb.** `occupancy:heartbeat` (Task: `task occupancy:heartbeat`) refreshes the caller's own lease across long quiet stretches — reading, building, waiting — that produce no gated write to ride on. It extends a lease the caller already holds: it never claims a free worktree, never takes a foreign lease, and never mints an owner. It is registered on the #3611 exact-lifecycle rewrite bridge, so under a supported host the command receives the matching `--session-id` automatically.

**An absolute lease-age cap, keyed on `claimed_at` and independent of refresh.** Occupancy admits whoever presents the occupant's session id, so "the owner is still writing" only proves that *some* process holds that string. With refresh and no absolute bound, any such process keeps a worktree indefinitely and the heartbeat TTL — the sole mechanism that reclaims a tree from a dead session — stops working. `OCCUPANCY_MAX_LEASE_MS` is decided in `occupancyLiveness`, where liveness is decided, so a lease past the bound reads as expired however recently it beat. `claimed_at` survives every re-stamp, and the re-stamp itself refuses an already-expired lease, so no refresh path can reach around the cap.

## Why twelve hours

The bound is sized by the **stalled** owner, not the busy one. Refresh keys on writes, and an agent dispatched overnight that finishes its work and then waits for operator input is alive, correct, and silent: it performs no writes, so it stops refreshing while remaining entirely legitimate. Overnight cohorts with a morning handoff are the normal operating shape here, so a bound that only cleared the longest actively-working session (eight hours was the first candidate) would expire a live owner between a 23:00 dispatch and a 09:00 return.

Twelve hours is 36 TTLs, following the existing TTL-relative idiom (`OCCUPANCY_REFRESH_AFTER_MS`, `OCCUPANCY_STALE_WARN_MS`):

- **Covers overnight-to-morning.** The dispatch-at-night, hand-off-in-the-morning window fits inside the bound.
- **Still reclaims inside a day.** A four-hour cap would interrupt real sessions; a 24-hour cap would satisfy "bounded" while failing the operational goal of getting a dead session's tree back.
- **Cheap when it does fire.** Passing the cap costs the owner one re-claim, never its work: the tree reads as free, the holder's next gated write is still allowed, and `session:start` re-claims and resets the age clock.

**Known limitation, recorded rather than solved here.** A pure time cap cannot distinguish a stalled-but-live owner from a dead one, because the only liveness signal on this path is a write. Twelve hours buys headroom; it does not resolve the ambiguity. If that becomes a real problem, the answer is a liveness signal that does not require a write — an explicit parked state, or refresh on non-write activity — not a larger number. That is the revisit trigger on the decision record, deliberately not built in this PR.

**The two dead states read differently, because they call for different responses.** A stale heartbeat means beat sooner, and points at `occupancy:heartbeat`. Past the cap means the tree is unheld while you are still writing, and points at `session:start`. The holder gets the cap message on its own next gated write and from `occupancy:heartbeat`, rather than having its protection disappear silently mid-session.

## Finding on the "is a refresh verb still useful post-#3611?" item

Yes, and #3611 is what makes it usable. Two separate points:

1. **The automatic refresh does not cover every live session.** It rides on gated writes, so a session that is reading, building, polling a review, or waiting on an operator produces nothing to re-stamp from and still ages out mid-work. That is the gap the verb closes — and the same gap that sizes the age cap above.
2. **Before #3611 the verb would have been close to unusable by the party that needs it.** Refresh must present the owner id, and an agent shell had no reliable way to do that: environment-inherited `DEFT_SESSION_ID` does not survive across fresh hook processes. #3611's exact-lifecycle rewrite bridge supplies the host's `--session-id` on the exact command, so registering `occupancy:heartbeat` there (`EXACT_LIFECYCLE_VERBS` / `DIRECT_LIFECYCLE_VERBS` plus its argument policy) means the caller who most needs to refresh can actually invoke it. Fixture coverage for that rewrite is in this PR.

## Testing

- Owner refresh keeps a lease alive across writes spaced under the TTL, past the point where it previously expired; refresh stays off by default and is floored against amplification.
- Refresh never resurrects a lease taken by another session, and never claims a free or expired one.
- A lease refreshed continuously past the cap reads as expired *with a fresh heartbeat*, so only claim age can explain it; refreshed continuously under the cap it stays live, with no regression to the refresh fix.
- `claimed_at` survives all three refresh paths (same-session re-arm, the verb, the write gate), pinned specifically — it is the single line whose regression silently disables the cap.
- The cap's deny and warning text is asserted distinct from the stale-heartbeat text in both directions.
- Dispatcher coverage for renew-on-allowed-write and the surfaced staleness warning; CLI coverage for `occupancy:heartbeat`; a `cases.ts` fixture for the #3611 rewrite.

Out of scope and deliberately untouched: the ritual-state fallback for identity, join negotiation (`occupancy:request`), spawn-tool occupancy (#3755), and hook reachability (#3742).

Rationale and the revisit trigger are recorded in `xbrief/decisions/2026-08-26-bound-the-occupancy-lease-with-an-absolute-12-hour-age-cap-on-cl.decision.json`.

Closes #3599


MScottAdams:
model: claude-opus-5-thinking-high-fast
role: parent

Guidance on the two Greptile P1s, so the age-cap one is not resolved by weakening the cap. Both findings are valid — I read the PR head at `occupancy.ts:591-640` and confirmed them independently. Not a re-dispatch; you own this PR.

## P1 at 600-608, "age cap breaks exclusion" — valid, and the remedy is a design call

Greptile is right, but the framing understates it. Returning `allow: true, occupant: null` on a non-live lease is pre-existing behaviour and correct for TTL expiry: a lease that stopped heartbeating means the holder is probably gone, so the tree is unheld and the next writer may proceed.

The age cap is the first mechanism that voids a lease whose holder is **demonstrably alive and writing right now** — the warning at 604-606 exists precisely because we know they are there. So the cap introduces a state that TTL never produced: two actively working sessions both allowed on one tree. That is a real loss of exclusion, not a cosmetic one.

**Resolve it by denying the capped holder, not by softening the cap.**

On `liveness === "age-capped"` with `incoming === record.sessionId`, return `allow: false` with a remediation that names the cap and tells the holder to release and reclaim. Peers hitting an age-capped lease keep `allow: true`, exactly as with any expired lease — that path is unchanged.

The reasoning: the cap was requested so a session that starts overnight and stalls waiting on an operator stops holding the tree indefinitely. Letting the capped holder keep writing achieves neither half of that — the tree is not released, and the holder now works unprotected. Denial achieves both: the tree becomes claimable by others, and the stalled holder is told to re-arm, which is the behaviour the cap was asked for.

Turning the warning into a denial is a strengthening, so it is in scope for this PR. Do not resolve this by extending the cap, making it advisory, or documenting the gap.

## P1 at 633-639, "failed refresh stays allowed" — valid, needs a discriminated result

`restampOccupancyHeartbeat` returns `null` for two different situations, and the docstring at 642-649 shows this is deliberate: the lease is now someone else's, or the refresh hit lock contention or IO error. Best-effort-on-error is right; best-effort-on-replacement is not. The fallback at 636 then writes under `occupant: live`, a record we already know is stale.

The window is narrow — a peer has to claim between the unlocked read at 598 and the locked refresh — but `stealOccupancy` makes it reachable without waiting on a TTL boundary.

Have the refresh return a discriminated result so the caller can tell the cases apart. Replacement detected: deny, with the new occupant in the message. Contention or IO error: keep the current allow, which is correct.

## P2 at `occupancy-heartbeat.ts:32` — valid, trivial

Confirmed at 621-624 and 638: `warning` is computed from the **pre-refresh** age and then returned alongside `refreshed: true`. The owner is told their lease is stale and to refresh it, on the very call that refreshed it. Suppress the warning when the refresh succeeded.

## Bar

Greptile 5/5 read from the score comment, not the check-run conclusion. Rebase onto master before merge. CHANGELOG conflicts resolve by union via `task changelog:resolve-unreleased`. Add a regression test for each of the three.

MScottAdams:
model: claude-opus-5-thinking-high-fast
role: parent

## STOP — the hard deny deadlocks its own remediation unless you reorder something first

Both fixes are the right calls. But I traced the deny path after posting and the capped-holder deny, as stated, locks the holder out of the command it tells them to run. This is #3611's joint unsatisfiability about to recur in a new place, so please read before you finish it.

### The trace

The occupancy gate covers **every non-spawn tool, shell included**. The only exemption is spawn:

```996:1003:packages/core/src/hooks/dispatcher.ts
  const actor = isSpawnTool(toolName) ? null : resolveMutationActor(input, environ);
  const occupancyGate = isSpawnTool(toolName)
    ? { allow: true, message: null as string | null, occupant: null }
    : evaluateOccupancyWriteGate(projectRoot, {
```

I checked for an earlier bail-out that would spare shell tools. There is none — every guard from 493 to 1245 keys on `isSpawnTool`, never on `isShellTool`, so a `Bash` / `Shell` payload reaches the deny at 1028-1045 and returns `occupancy-occupied`.

The lifecycle carve-out that would rescue this runs **downstream of that return**:

```1419:1421:packages/core/src/hooks/dispatcher.ts
  const lifecycle = inspectExactLifecycleCommand(input.payload);
  if (lifecycle === null) return decision;
  if (!lifecycle.requiresOwner) return decision;
```

Line 1419 is never reached when line 1039 has already returned a denial.

### Why that breaks your fix specifically

Today the owner is never denied — denial only ever hits a non-owner, and the code says so at line 141, "product-path write while **another** live session occupies this worktree". Your change introduces the first state in which the **owner** is denied.

At that moment the holder is denied on every tool, including the shell tool carrying `deft occupancy:release`. And the remediation string already promises the opposite:

```110:110:packages/core/src/session/occupancy.ts
    "The occupant may release (`occupancy:release` / `session:end`)."
```

So the capped holder would be told to release, and denied when they try. No write, no release, no reclaim — the tree stays held by a session that cannot do anything about it until a human intervenes from an unhooked terminal. That is strictly worse than the exclusion gap you are closing.

### What to do

**Move the exact-lifecycle-command check ahead of the occupancy deny**, so `occupancy:release`, `session:end`, and `occupancy:steal` are evaluated before 1028 rather than at 1419. That is the general fix, it is independently correct — the remediation at occupancy.ts:110 has been writing a cheque this ordering cannot cash — and it makes your hard deny safe.

If you would rather keep the blast radius small, the narrow alternative is to permit those verbs inside the capped-holder branch only. I prefer the reorder, but either satisfies me; say which you took and why.

Do **not** resolve this by dropping back to `allow: true` on the cap. The exclusion gap Greptile found is real and the deny is still the right call — it just needs an exit.

### Test gap you will need to close

The nearest fixtures assert identity classification only, not the gate verdict:

```695:711:packages/core/src/hooks/fixtures/cases.ts
    id: "claude-posix-shell-identity-missing",
    ...
      tool_input: { command: "deft occupancy:release" },
    expected: {
      ...
      hostIdentity: { status: "missing", sessionId: null },
```

Nothing anywhere asserts what the gate *decides* for a lifecycle command against an occupied or capped tree. Add that case: capped holder runs `occupancy:release` and is allowed through.

### Out of scope for you, I am filing it separately

The same ordering means a denied **non-owner** cannot run the `occupancy:steal` that `formatOccupancyRemediation` prints for them either — the deny at 1028 fires before the lifecycle rewrite at 1419. It works from a human terminal, which is unhooked, and not from an agent, which is the population the message is written for. That is pre-existing and not yours to fix in this PR. Code-path trace only; I have not run it.

MScottAdams:
role: worker (resumed after the workspace disconnect)

Rebased onto master (`63d29d5b`) and pushed `b0f85cc5`. Below: the WIP snapshot review, the deadlock question, and what changed.

## The rescued WIP snapshot was sound

`0e019c41` compiled and its tests passed, and re-reading it against your guidance it had genuinely landed both P1s: the capped-holder deny at `occupancy.ts:600`, and the discriminated `RestampOutcome` with `lost` re-deciding against disk. No half-written edit. Two things it did not carry: the P2, and any `dispatcher.ts` change.

## The deadlock does not fire, so I took neither remedy

You asked me to say which I took. Neither, because I could not reproduce the deadlock, and I think the trace has one wrong step.

The occupancy gate at `dispatcher.ts:996-1045` lives inside `inspectMutationGates`. That function has three call sites: `1760` (spawn), `1791` (direct-write tools), and `1340` — the only shell one. The shell site is inside `decideShellDestFormsThenRuntimeAuthority`, and it is reached only for targets that `classifyProductDestForms` resolves:

```1320:1321:packages/core/src/hooks/dispatcher.ts
  if (command !== null && destFormsEnforced) {
    for (const dest of classifyProductDestForms(command)) {
```

That classifier recognizes four verbs — `rm`, `rmdir`, `git checkout`, `git restore` (`dest-form.ts:24`, `:54`) — and the whole branch is gated on `shellDestForms === "enforce"`, which is off by default (`dispatcher.ts:1319`, and `dispatcher.test.ts:81` records the default). A `deft occupancy:release` payload is not a dest form under either setting, so it never enters `inspectMutationGates` and never reaches the deny at 1028. Everything you traced downstream of that line is correct; the step that does not hold is "the occupancy gate covers every non-spawn tool, shell included." It covers Edit/Write and, under an opt-in policy, four destructive shell verbs.

Rather than argue it, I pinned it. `dispatcher.test.ts` now asserts, on a real age-capped lease, that the holder's `Edit` is denied with the cap remediation and that both `deft occupancy:release --session-id=owner` and `deft session:start --session-id=owner` are allowed — under `enforcingSeams()`, so the assertion holds even with `shellDestForms: enforce`. If shell gating ever widens to cover lifecycle commands, that test fails and the reorder becomes required work with a live reproduction attached to it.

I did not do the general reorder, for the reason you gave yourself: it also resolves the non-owner `occupancy:steal` case you are filing separately, and moving lifecycle commands ahead of a deny for non-owners is an authorization change that deserves its own review rather than riding a review-cycle fix batch.

On the corroboration: I think that one is a different mechanism. `releaseOccupancy` resolves its caller from `--session-id` then `DEFT_SESSION_ID`, and denies when the result is neither the owner nor an expired lease (`occupancy.ts:509-510`). A bare invocation with neither set gets an empty caller, so it fails the ownership test in the CLI — no hook involved. That also means a capped holder's bare release does work, because an age-capped lease reads as expired.

## Fixed on this push

**Greptile's remaining P1 (`occupancy.ts:642-648`) — valid.** `unavailable` was still returning the pre-lock record's allow. Your guidance said contention keeps the allow, and Greptile's case is why that is not quite enough: a peer takeover is one of the things that holds the lock, so a re-stamp blocked to timeout can hide the very handover the `lost` branch exists to catch. Both non-refreshed outcomes now re-decide against the file on disk. Contention over a lease still ours re-allows exactly as before — the #3736 direction — and a lease in another session's name denies. The discriminated union stays; `heartbeat` still needs it to tell the owner "changed owner" from "still yours, retry."

**P2 — fixed.** A successful re-stamp returns no warning. One existing dispatcher test asserted the contradiction directly ("renews the lease *and* warns"); it now asserts the absence. Note the consequence: since the write gate always refreshes and the staleness floor (15m) is above the refresh floor (5m), that warning now reaches an owner only when the automatic renewal could not run. That is the only case where it was ever actionable, but it is narrower than the CHANGELOG claimed, so I tightened the CHANGELOG and `commands.md` rather than leave them overselling it. I also removed a sentence the snapshot had left contradicting its own new behaviour ("past the cap the tree reads as free" beside "the holder's writes are refused").

Three regression tests added: takeover during a lock timeout, warning suppression on a refreshing write, and the capped-holder release path. I checked the takeover test against the pre-fix condition to confirm it is not vacuous.

`task check`: 14956 passed, 151 skipped, 0 failed. `verify:forward-coverage` exit 0.

Issue #3791, opened 2026-08-27
feat(lifecycle): automate the post-merge brief sweep and stamp the evidence a merge can honestly produce


MScottAdams:
## Summary

The lifecycle move from `active/` to `completed/` cannot happen before a merge, but it is still performed **manually** afterwards. That manual step is what leaves orphaned briefs on master, and it is the one part of the sequence that does not need a human.

`task swarm:finalize-cohort` already exists and already does most of this. Automating its invocation on merge, and teaching it to stamp the evidence it can honestly produce, is a smaller change than any of the gates filed around this failure -- and closes more of the class.

## Why automation is the right layer

The window is structural and cannot be designed away. `scope:complete` refuses without delivery evidence: a merge commit that is an ancestor of the refreshed delivery ref, which by construction cannot exist before the merge. All three round-1 critics on the set critique verified this independently.

So the sequence is forced: merge, *then* complete. What is **not** forced is that a human performs the second step, or that an arbitrary amount of time passes between the two. Today the gap is bounded only by whether someone remembers.

Three occurrences in two days -- #3611, #3598, #3609 -- each an instance of nobody performing a step that a machine could have performed immediately.

## The incumbent is closer than it looks

Despite the name, `finalize-cohort` is **not** cohort-scoped:

```11:16:packages/core/src/swarm/finalize-cohort-cli.ts
export const FINALIZE_COHORT_USAGE = `Usage: task swarm:finalize-cohort -- [--pr <n>[,<n>...]] [options]

After cohort PRs merge, sweep story briefs active/ -> completed/ and open the lifecycle PR.

  --pr <n>[,<n>...]            Merged PR numbers (closing issues map to active stories)
```

It accepts arbitrary merged PR numbers and derives the stories from each PR's closing issues, so `--pr 3786` was always a legal invocation against the solo external PR that caused the most recent occurrence. Sibling `complete-cohort` (#1487) handles the sweep; #2225 shipped the post-merge flow.

Two things stop it closing the class today:

1. **It is opt-in.** Nothing invokes it on merge; someone has to remember, which is the same failure being automated away.
2. **It supplies delivery evidence but never acceptance evidence.** `completeCohort` routes through `runTransition`, which runs the acceptance gate unconditionally, so a brief with unattested items is refused no matter how it is invoked. This is why running it manually on #3609 would not have helped.

## Remedy sketch

[...]

Issue #3790, opened 2026-08-27
enh(verify): verify:orphan-active reports unrelated master orphans on every PR, misattributing them


MScottAdams:
## Summary

`verify:orphan-active` scans the whole working tree with no diff scoping. As a member of `task check` it runs on every pull request, so a pre-existing orphan on master surfaces as a failure on **every open PR**, attributed to changes that did not cause it.

This is a **noise and attribution** problem, not a merge blocker. The original filing claimed it blocked merges; that was wrong and is corrected below.

## Correction to the original filing

Two claims in the first version were false. Recording them because the escalation comment repeats them.

**"One orphan on master makes every open PR unmergeable."** False. Branch protection on master requires exactly one context, `TypeScript (build + lint + test)`. `Merge gate (task check)` is **not** required. PR #3784 was observed `mergeable: true`, state `unstable`, while the orphan existed. Nothing was blocked. The effect is a red advisory check on every PR.

**"This is not gate weakening, because master-tip enforcement stays fail-closed."** That defence does not hold, because the fallback it relies on is itself a non-required check. Scoping the PR-time check without also hardening the master-tip surface would leave the property enforced nowhere that blocks anything.

## Why diff scoping -- the original remedy -- does not work

The first version proposed failing on orphans "introduced by this branch" and exempting inherited ones. That cannot work, for two measured reasons:

- **The introducing PR is invisible to the gate.** CI runs before the merge; the issue closes on the merge. At check time the issue is still open, so no orphan exists. PR #3786's own `Merge gate (task check)` returned **success** while creating the orphan. No branch ever introduces an orphan at a moment the gate can see, so a diff trigger would exempt everyone and catch nobody.
- **The brief need not be in the diff at all.** For #3598 the brief landed on master seventeen hours before its closing PR.

## What is actually left, and it is small

Two changes, both caller-side:

**1. Scope the PR-time invocation to the PR's own closing issues.** The evaluator **already implements** `--issue N` -- the scoped form used as the DONE precondition. Passing the PR's closing issues turns the check from "is the repository clean" into "did this PR leave its own lifecycle in order". No evaluator rewrite, which also dissolves the collision with #3767 and #3774 that the first version flagged.

**2. Harden the master-tip surface so the property is enforced somewhere actionable.** A scheduled or on-push master run that reports to a place a maintainer reads, rather than an optional check on unrelated PRs. Without this, step one removes the only signal anyone currently sees.

[...]

Issue #3767, opened 2026-08-26
bug(lifecycle): verify:orphan-active prefers a stale cached issue state over the live read that would correct it


MScottAdams:
## Summary

`verify:orphan-active` prefers a **stale cached issue state over the live read that would correct it**. A cache hit is returned unconditionally, with no age bound and no basis reporting, so the gate can be confidently wrong — and being wrong is exactly what stops it from checking.

Recut after an N=3 design-critique panel (successor lean 5427870380). The panel broke the original framing unanimously. Two claims in the first version were false and are withdrawn below.

## Withdrawn: "there is no live fallback"

The original body quoted `readCachedIssueState` in isolation and concluded a cache miss returns `null` and passes vacuously. **That is wrong.** The decision point is `resolveIssueState`:

```ts
const cached = readCachedIssueState(projectRoot, ref);
if (cached !== null) return cached;
if (skipGh) return null;
return fetchIssueStateLive(ref, runGh);
```

A miss already falls through to a live REST read, locked by a test named *"uses live gh issue state when cache is absent."* One critic reran that file 24/24 green; another measured a run with no `.deft-cache` at all and got a correct exit 1. **Vacuous green requires `--skip-gh`,** which is a deliberate flag rather than an accident of cache state.

Also withdrawn: the fresh-clone argument built on that claim.

## The actual defect

A **cache hit wins unconditionally**. No freshness bound, no age check, no comparison. A stale `open` beats reality and suppresses the live path.

Measured on 2026-08-26 immediately after #3611 closed:

| | |
|---|---|
| live state of #3611 | `closed` |
| `.deft-cache` entry | `open` |
| cache written | ~12 h earlier |
| `verify:orphan-active` | `scanned 1 running brief`, **exit 0** |

Reproduced again the same day during a session ceremony: `cache-fresh: stale-by-drift — 2 cached-open issue(s) absent from live open set`. Same cache, different consumer, and that one noticed.

## Remedy: split by query shape

The panel converged here. **Not one mechanism — two.**

**Scoped `--issue N`** — a direct authoritative read. It is a proof obligation before DONE, N is 1, and it should stay fail-closed on unknown. Largely already correct today.

[...]

Issue #3737, opened 2026-08-25
BLOCKER: bug(scm): resolveBinary prefers ghx on presence alone, so a broken ghx silently kills triage, ingest, cache and pr:watch


MScottAdams:
## Summary

`resolveBinary` prefers `ghx` over `gh` on **mere PATH presence**, with no health probe and no fallback. When the `ghx` binary is present but cannot run, every SCM read path in the framework fails with an opaque error: no stderr, no diagnostic, and an exit code that means nothing to the reader.

I hit this today. `ghx.exe` on this machine cannot start at all -- not even `--version`:

```
PS> ghx --version
exit=-1073741502          # 0xC0000142  STATUS_DLL_INIT_FAILED

PS> gh --version
gh version 2.x            # fine
```

Reproduced through the exact call shape the framework uses (`packages/core/src/scm/gh-rest.ts`):

```
node -e "spawnSync('ghx',['api','repos/deftai/directive/issues/3736'],{env:process.env})"
  status = 3221225794     # 0xC0000142
  error  = none
  stderr = ""
  stdout = 0 bytes

node -e "spawnSync('gh', ['api','repos/deftai/directive/issues/3736'],{env:process.env})"
  status = 0
  stdout = 7622 bytes
```

## Verbs this breaks

Everything that reads GitHub through the shared resolver. Observed today, all with the same empty-stderr failure:

| Verb | Symptom |
|---|---|
| `deft triage:accept --issue N` | `issue:ingest delegation failed; audit entry rolled back` |
| `deft issue:ingest --issue N` | `gh CLI failed fetching #N:` (blank reason) |
| `deft cache:fetch-all -- --source github-issue` | `rest_issue_list_paginated failed ... exit=3221225794 stderr=''` |
| `deft pr:watch <N>` | `could not resolve PR HEAD sha (gh pr view + REST both failed)` |

Cached reads still work, which makes this worse rather than better: `issue:ingest` on an already-cached issue succeeds, so the framework looks healthy until you touch anything new. A freshly filed issue cannot be triaged at all.

## Cause

`packages/core/src/scm/binary.ts`:

```ts
export function resolveBinary(whichFn: WhichFn = defaultWhich): string {
  for (const candidate of BINARY_PREFERENCE) {   // ["ghx", "gh"]
    if (whichFn(candidate) !== null) {
      return candidate;                          // presence only
    }
  }
  ...
}
```

[...]

PR #3802, opened 2026-08-27
Refuse a merge that closes an issue whose brief is left running and unattested (#3781)

### LINKED ISSUE #3781: bug(lifecycle): a PR can merge while closing an issue whose brief is left running and unattested
## Summary

A pull request may merge while closing issue `N`, leaving `N`'s brief in `xbrief/active/` with `plan.status == "running"` and acceptance items carrying no evidence and no disposition. Nothing checks this before the merge, so the unattested state lands on master and `verify:orphan-active` then reports an orphan that no in-flight party can clear.

Three occurrences in under two days: #3611, #3598, #3609. The third was caused by a **maintainer merging**, not a worker exiting -- so worker envelope discipline does not cover it.

## The trigger is the closing reference, not the diff

This is the load-bearing design decision and the round-1 panel settled it against the obvious alternative.

**A diff-based trigger does not work.** Two independent reasons, both measured:

- CI runs *before* the merge, and the issue closes *on* the merge. At check time the issue is still open and no orphan exists. PR #3786's own `Merge gate (task check)` returned **success** on both runners while creating the orphan. A gate keyed on the branch diff can never fail the PR that introduces the problem.
- The brief need not even be in the diff. For #3598 the brief landed on master **seventeen hours before** its closing PR, so no diff trigger fires on the culprit at all.

**The correct trigger is available and unambiguous.** At merge time the forge knows which issues the PR closes. Gate on: *this PR closes issue N, and issue N's brief is in `active/` with `running` status and unattested items.* Both facts are known, cheap to compute, and specific to the PR being merged.

Triggered this way, this gate alone closes all three observed occurrences.

## Fail closed, not warn-first

The original filing proposed warn-first, reasoning that recording evidence at completion time is legitimate and blocking a PR on completion-time metadata would be wrong.

That reasoning was wrong, for a reason the panel made explicit: **warn-first shipped alongside a blast-radius fix (#3790) is worse than today.** The orphan reaches master anyway, and the only check other branches still hit has been downgraded to advisory. Nobody is then forced to notice.

The narrow rule keeps ordinary work unblocked: a PR may leave a brief unattested; it may not **merge** one whose issue is closing in the same act.

## Why this is not a #3779-shaped trap

The obvious objection is that a fail-closed merge gate might block someone who cannot perform the remedy -- the pattern where a gate prints a recovery command the blocked party cannot execute.

It does not, and correcting the reason matters because it was wrong in the original filing.

```226:227:packages/core/src/scope/acceptance-evidence.ts
  if (!isNonEmptyString(obj.recorded_by)) {
    return { ok: false, message: "evidence.recorded_by is required" };
```

Evidence requires only a non-empty `recorded_by`. The human-origin gate applies to **dispositions**, and its own comment states the intent: *"Reuse authz human-origin gate (#2944) so agent-self stamps cannot waive criteria."* An agent may not **waive** a criterion. Nothing prevents it **evidencing** one.

So the "only the original author can attest, therefore the blocked party has no remedy" premise in the first version of this issue was **false**. The PR author can attest their own items. The blocker is self-serviceable, which is exactly what makes fail-closed acceptable here.

## Risk this creates, and the mitigation

Because evidence is an unauthenticated string, a fail-closed gate creates **fabrication pressure**: the cheapest way past it is to type a plausible pointer.

That is a real cost and should be designed against rather than ignored:

- Prefer machine-derivable evidence where the kind allows it. A `merge` kind can be stamped truthfully and automatically -- see the companion issue on post-merge automation.
- Respect the #3240 kind taxonomy. A merge cannot evidence a `smoke`, `uat`, `deploy`, or `observed_behavior` criterion, and automation must not stamp `test` across the board simply because the field accepts a string.
- Consider surfacing evidence pointers in review, so a fabricated one is visible to a reviewer rather than only to the gate.

## Acceptance

- [ ] A PR that closes issue `N` and leaves `N`'s brief `running` in `active/` with unattested items fails a merge-time check
- [ ] The trigger is the PR's closing reference, **not** the branch diff
- [ ] The failure names each unattested item and the exact shape it needs (`{kind, pointer, recorded_at, recorded_by}` or `{disposition, reason, provenance, recorded_at}`)
- [ ] A PR that leaves an unattested brief **without** closing its issue is unaffected
- [ ] The check reuses `evaluateAcceptanceEvidenceGate` rather than reimplementing the rule, so the two cannot drift
- [ ] Regression fixture reproducing #3609: PR closes the issue, brief has five unattested items, gate fails
- [ ] Regression fixture for the #3598 shape: brief already on master well before the closing PR, gate still fires on the closing reference

## Corrections to the original filing

Recorded because the first version asserted them and they are wrong:

- **"Only the author can attest."** Half false. Human-origin provenance is required for dispositions only, not evidence.
- **"One orphan on master makes every open PR unmergeable."** False. Branch protection on master requires exactly one context, `TypeScript (build + lint + test)`; `Merge gate (task check)` is **not** required. PR #3784 was observed `mergeable: true`, state `unstable`. The effect is a loud red advisory on every PR, not a repository-wide merge stop. The escalation comment on this issue overstates the harm and should be read with this correction.
- **Warn-first posture.** Superseded, as above.

## Out of scope

- Blast radius when a bad merge happens anyway (#3790, demoted to a caller-side change).
- Post-merge automation of the lifecycle move (companion issue) -- complementary, and the two together make the manual path rare.
- Closeout ownership (#3783, closed as dominated).

## Why this cannot be solved by doing it in one step

`scope:complete` refuses without delivery evidence: a merge commit that is an ancestor of the refreshed delivery ref, which by construction cannot exist before the merge. Verified by all three round-1 critics. The two-step is genuinely structural, so the window always exists -- this issue narrows what may enter it, and the companion issue shortens how long it stays open.

## Provenance

Round-1 set critique of #3781 / #3783 / #3790 (synthesis 5433848104). Trigger design, fail-closed posture, and every correction above were established by that panel and independently re-verified by the parent at `e2b62217`.

Refs #3609, #3598, #3611, #3786, #3240, #3041, #3790, #3783.

MScottAdams:
## What this does

A pull request may leave a brief unattested. It may not **merge** one whose issue it closes in the same act.

`task verify:pr-closeout-attestable -- --pr N` fails closed when PR `N` closes issue `M` and `M`'s brief is still `running` in `xbrief/active/` with acceptance criteria carrying neither `x-directive/evidence` nor `x-directive/disposition`. `pr:wait-mergeable-and-merge` runs it as the last gate before the `gh pr merge` call.

Closes #3781.

## The trigger is the closing reference, not the diff

This is the load-bearing choice and it was settled by the round-1 panel on measured grounds, both of which the tests encode:

- **CI runs before the merge; the issue closes on the merge.** At check time the issue is still open, so no orphan exists yet. PR #3786's own `Merge gate (task check)` returned **success** while creating the orphan that later red-flagged the repo. A gate keyed on the diff can never fail the PR that introduces the problem.
- **The brief need not be in the diff.** For #3598 it landed on master seventeen hours before its closing PR.

At merge time the forge knows which issues the PR closes. The gate reads those structured closing references (`closingIssuesReferences`, via the incumbent `fetchClosingIssuesReferences`), maps them to `active/running` briefs, and asks the acceptance-evidence gate.

## Where it hooks in, and why there

Immediately before `mergeFn` in `waitMergeableAndMerge`, after the head-bound approval gate.

The #701 protected-issue check runs *before* the wait loop, and that was the obvious template. It was not the right one here: the state this gate reads is the brief in the working tree, which can still change during a 60-minute wait. Binding the verdict to the tree that actually merges is the point. The standalone `task verify:pr-closeout-attestable` verb covers the early-warning case (pre-PR, review-cycle, local) through the *same* code path, so there is no second call site to drift.

Design record: `xbrief/decisions/2026-08-27-the-merge-time-closeout-gate-triggers-on-the-pr-s-structured-clo.decision.json`.

## Reuse, not reimplementation

`evaluateAcceptanceEvidenceGate` had exactly one caller, `scope/transition.ts` on `complete`. It now has two. The rule is not restated anywhere in the new module -- the gate calls it and maps its blockers into the message. Axis inference reuses the exported `inferRequiredStrictAxes`. The two cannot drift.

## Designing against the fabrication pressure this creates

A fail-closed gate on an unauthenticated evidence string makes typing a plausible pointer the cheapest way past. That is not hypothetical: an agent under this pressure stamped five items with `kind: test` where three pointers were markdown docs (#3791 comment 5433984468). The pointers were real; the *kinds* were wrong.

So the refusal names each unattested criterion **and the exact shape it needs**, with the kind taxonomy narrowed to what that criterion's axis actually accepts:

- no strict axis inferred -> the full `test|review|merge|deploy|smoke|uat|observed_behavior` taxonomy
- one strict axis inferred -> only that kind, plus an explicit line that `merge` and `review` cannot satisfy it
- **two** axes inferred -> no single `evidence.kind` covers two axes, so offering a kind list would be a lie. The message says that and names the three real routes: pin one axis with `requires`, split the criterion, or record a human-origin disposition.

It also states plainly that `recorded_by` accepts any non-empty string, and that an agent may *evidence* a criterion but only human-origin provenance may *waive* one.

Kind-versus-pointer validation is **not** built here -- that is #3791's scope.

## Fail-closed boundaries

Three-state exit: `0` attestable or closes nothing / `1` unattested closeout / `2` config or lookup error.

- A closing-reference lookup that cannot be resolved is `2`, never `0`. The gate does not certify a merge it could not check.
- The read pins plain `gh` via the existing `makeGateRunner()`, because `ghx` is a cached GET proxy and a stale closing-reference read would fail this gate **open**. When `gh` is absent the verdict carries the proxy caveat rather than implying freshness (#3767 / #3737).
- A PR that leaves an unattested brief **without** closing its issue is untouched: no closing reference, no check.

## Regression fixtures

Both shapes from the issue, and they exercise genuinely different mechanics:

- **#3609 shape** -- brief with five bare `title` + `status: proposed` items, PR closes its issue, gate exits 1 with five named criteria.
- **#3598 shape** -- a real temp git repo. The brief is committed to `master` first, then a branch commits only unrelated source. The test asserts `git diff master...HEAD --name-only` is exactly `["src.ts"]` (so a diff-keyed gate would see nothing) and that the gate still fires on the closing reference.

## Notes for the reviewer

- Not added to `task check`: the gate needs a PR number, and `task check` has no PR context. That absence is the whole reason a diff-keyed variant was rejected.
- No AGENTS.md rule. Per Rule Authority and Thin Fail-Closed Design (#3265) this ships as one deterministic check with one remediation; the cascade chains it automatically, so prose would add nothing an agent must remember.
- `skipCloseoutAttestableGate` defaults to `skipHumanMergeGate`, mirroring the existing `skipMergeApprovalHeadGate` convention in the same function. No CLI flag sets `skipHumanMergeGate`, so production always runs the gate; the dedicated tests opt in explicitly.
- #3790 is demoted and must land after this. Not touched here.

## Checklist

- [x] `/deft:change` -- N/A. Single-scope story under `xbrief/active/2026-08-27-3781-...xbrief.json`; not cross-cutting or architectural beyond the one gate it adds.
- [x] `CHANGELOG.md` -- entry added under `[Unreleased]` / Added.
- [x] `ROADMAP.md` -- N/A. Not a tracked roadmap item.
- [x] Tests pass locally -- `task check:merge` green: 1059 test files passed, 6 skipped; 15050 tests passed, 152 skipped; statements 88.13%, branches 81.33%.

## Post-Merge

- [ ] Verify #3781 auto-closed after squash merge (#167).
- [ ] `task scope:complete`, then `task verify:orphan-active -- --issue 3781` and `task verify:completed-tracked -- --issue 3781` against `origin/master`.


MScottAdams:
## The gate blocks this PR

This PR closes #3781, and #3781's brief is `running` in `active/` with eight bare `title` + `status: proposed` items. So the gate it adds refuses its own merge. Run against this PR at `b94b1a98`:

```
$ task verify:pr-closeout-attestable -- --pr 3802
verify:pr-closeout-attestable: PR #3802 closes #3781, leaving 1 active/running brief with 8 unattested acceptance criteria (project_root=C:\Repos\deft\directive).
  Merging now strands the brief on master: the issue closes, the brief stays running in
  active/, and scope:complete then refuses it. A PR may leave a brief unattested; it may
  not merge one whose issue is closing in the same act (#3781).
  Unattested criteria:
    xbrief/active/2026-08-27-3781-...xbrief.json (closes #3781)
      - items[0] "A PR that closes issue `N` and leaves `N`'s brief `running` in `active/` with unattested items fails a merge-time check"
          why:   no x-directive/evidence or x-directive/disposition; scope:complete refuses to auto-complete (#3240)
          needs: x-directive/evidence {kind: test|review|merge|deploy|smoke|uat|observed_behavior, pointer, recorded_at, recorded_by}
                 or x-directive/disposition {disposition: waived|deferred|not_applicable, reason, provenance {kind: operator-cli|operator-session|human-event, actor: <non-agent>}, recorded_at}
      ... 7 more, each named ...
  Evidence is not authenticated — recorded_by accepts any non-empty string. Record what you
  actually did: a pointer must be the artifact its kind names (a test run for test, this PR's
  merge for merge, a deployment for deploy). An agent may evidence a criterion; only
  human-origin provenance may waive one (#3240 / #2944).
  Remediation (performable by this PR's author): stamp the criteria above on the brief in this
  branch, commit, push, then re-run:
    task verify:pr-closeout-attestable -- --pr 3802
  Trigger is the PR's closing references, not the branch diff. A PR that leaves an unattested
  brief without closing its issue is unaffected.

exit 1
```

Three things this confirms that the unit tests state but do not prove end to end:

- The trigger works against the live forge. Nothing about the diff was consulted; the gate read this PR's structured closing reference, found #3781, and mapped it to the brief.
- The blocked party can self-remediate, which is what makes fail-closed acceptable rather than a #3779-shaped trap. The author of this PR is the party who has to stamp, and they can.
- The refusal is actionable without reading source: each criterion is named with the exact record shape it needs.

The next commit stamps those eight items and this run flips to exit 0.

MScottAdams:
ended_at: 2026-08-27T16:02:07.337Z

Issue #3801, opened 2026-08-27
feat(orchestration): no form exists for a standing unattended mandate \u2014 operator hands a target list and leaves


MScottAdams:
## Summary

An operator can hand a parent a **standing mandate over a target list and leave**. Run on 2026-08-25/26 across the #3613 occupancy cluster: roughly nine hours unattended, 14 design-critique arcs, 40 critic deposits, nine issues shipped with lifecycle artifacts tracked on master.

It worked. Nothing in the framework describes it, bounds it, or says what the parent may decide while the operator is asleep.

## What the mandate actually looked like

Reconstructed from the session, because this is the artifact worth generalizing:

> "I am going to go AFK for the night but I want you to work the 3613 issues for occupancy hang-ups. Look it over now and ask questions as needed. I want you to run critics against each and then re-run difference critics as needed until agreement is reached. You may go past the original 3 limit but no more than 5. Let's see how it works. You will be the only orchestrator working."

Six elements, each load-bearing:

1. **A bounded target list** -- the children of one tracker, not "make progress"
2. **A motion to apply** -- run arcs, re-run on disagreement
3. **A budget with headroom** -- past the default, capped at 5
4. **An exclusivity clause** -- sole orchestrator, so contention was the operator's decision rather than an accident
5. **A question window before departure** -- "ask questions as needed" while the operator was still present
6. **An experimental posture** -- "let's see how it works" set the tolerance for imperfect outcomes

## What this is not

- **`swarm:launch --autonomous`** is cohort-dispatch consent: a pre-approved set of stories collapsing per-phase approval into one #1378 allocation token. It authorizes *dispatch*, not a standing session mandate over decisions.
- **The AGENTS.md headless bypass** is about skipping onboarding for a dispatched task.
- **`vbrief.md:440`** lets headless runs apply provisional estimates and stage transitions without confirmation. Narrow, and about estimates.

None of these is "here is a list, here is the motion, I am asleep, use judgement."

## The hard problem: what substitutes for operator confirm

This is the crux and the reason to file rather than just document.

`contracts/design-critique.md:318` makes auto-stamp conditional on "**the operator has confirmed or amended that map**". Overnight there is no operator. `contracts/human-origin-authz.md` has no standing or deferred consent form -- consent is present-tense by construction.

[...]

Issue #3785, opened 2026-08-26
bug(hooks): fail-closed hook registration travels without a hostable runtime (exit 127)


nheroux-bit:
## Problem

A Cursor cloud agent VM fail-closes every mutation when `deft-hook` is not on PATH. That is not a laptop that forgot `deft update`. The image has no hook binary and no Node, so there is no recovery from inside the session.

Observed live: Cursor cloud agent `bc-e6bdd4c0` on deftai/Market-Aperture, 2026-08-26.

**Live consumer hazard (read this first).** The reporter unblocked by setting `failClosed: false` in the tracked `.cursor/hooks.json`. That works once, and it leaves the repo drifted: the next `deft update` rewrites that file and silently re-arms the lockout. Until this ships, the durable escape is `deft policy:disable-host-hooks --host cursor --confirm` (`plan.policy.hostHooks.cursor = false`). Do not hand-edit the deposited hook file.

## What happened

- Fail-closed project hooks call `deft-hook`.
- `deft-hook` is missing. Every Write, Shell, and `gh` dies with exit 127.
- GitHub MCP 404s on the consumer repo.
- Skip-hook resume did not unblock writes.
- Agent asked for a no-op `deft-hook` on PATH that exits 0.

Nothing was committed. The session could diagnose and could not land the isolated PR.

## What this is not

- Not a duplicate of #3732 (framework repo has no maintainer path to deposit agent hooks).
- Not a duplicate of #3742 (tracker for hook enforcement defects in a maintainer checkout).
- Not a developer laptop that skipped `deft update`. The cloud VM never had the binary, Node, or a documented no-Node recovery.
- Not a case for fail-open-on-absence. Round-1 panel (synthesis 5434430152) rejected that option 3/3: when the binary is absent, no Directive code runs; "absent AND never deposited" selects nothing because deposit is the precondition for the hook firing; deleting the `deft-hook` binary would become a bypass of the write fence, intent ceiling, ritual, and occupancy lease (#3156).

## The actual defect

Directive git-tracks the fail-closed registration and gitignores the runtime it names. `.cursor/hooks.json` is kept trackable; `.deft/core/` is born-ignored. Registration travels via git; implementation does not. The deposited command is an npm bin pointing at a Node ESM script, so even a host-shipped binary would not resolve it without a hostable runtime.

This is not Cursor-cloud-specific. Every CI runner, container, and fresh clone without a global npm install has it. The reporter's VM is the instance that got reported, not the boundary.

[...]

Issue #3774, opened 2026-08-26
bug(lifecycle): verify:orphan-active reaches its config exit code through a catch-all instead of handling ScmStubError


MScottAdams:
## Summary

`verify:orphan-active` reaches its configuration exit code by accident. When neither `gh` nor `ghx` is on PATH, `ScmStubError` escapes the evaluator and the verb handler, and is absorbed by a generic catch-all in the CLI dispatcher:

```3257:3262:packages/cli/src/dispatch.ts
    return await invokeHandler(handler, handlerArgv);
  } catch (err: unknown) {
    const message = err instanceof Error ? err.message : String(err);
    io.writeErr(`directive: ${message}\n`);
    return 2;
  }
```

Exit 2 is the correct code, so the operator-facing behaviour is not broken. It is unowned. Five sibling call sites handle this explicitly and this one does not.

## What this issue is not

The original filing claimed a missing toolchain **exits 1 and is reported as an orphan**, with a `scope:complete` remediation attached. **That is false and has been removed.** The inherited measurement behind it came from the #3767 round-1 panel and was taken against a direct `evaluate()` import, not against `task verify:orphan-active`. The production path has always returned 2.

The body's structural diagnosis was right — "an exception escaping around the verdict logic" — but the consequence it drew from that was wrong.

## The actual defect

`ScmStubError` is handled explicitly at five call sites, all mapping it to exit 2:

```
packages/core/src/intake/reconcile-issues-cli.ts:36
packages/core/src/intake/issue-ingest-cli.ts:37
packages/core/src/scm/main.ts:34
packages/core/src/scm/main.ts:121
packages/core/src/triage/actions/index.ts:97
```

`orphan-active` is absent from that list. Three consequences:

1. **The message is wrong.** The operator sees `directive: <raw error text>` instead of the gate's own diagnostic. It does not name the gate, and the #2275 toolchain guidance is not presented in the gate's voice.
2. **The inner API contradicts the published contract.** `tasks/verify.yml:506` documents "Three-state exit (0 clean / 1 orphan / 2 config)" and the evaluator types it as `readonly code: 0 | 1 | 2` (`packages/core/src/orphan-active/evaluate.ts:21`), already emitting `code: 2` at lines 363 and 389. A programmatic caller importing `evaluate()` gets a throw where the contract promises a code.
[...]

PR #3807, opened 2026-08-27
fix(hooks): gate linked-worktree writes against the target tree

MScottAdams:
## Summary

Commit 1 of #3794. Direct writes are gated against the Git worktree that contains the write target.

- Occupancy and ritual move together to effectiveRoot
- effectiveRoot is admitted only when --git-common-dir matches payloadRoot
- Authz grant scoping, the authz audit trail, and the kill-switch stay on payloadRoot
- Foreign-repository targets are refused
- Deny messages name both roots

Commit 2 (active scope / file_scope / assist-scratch reclassification) is out of this PR.

## Test plan

- [x] Linked-worktree occupancy allow/deny matrix
- [x] Deposited hook CLI without --project-root
- [x] task check:merge green

Tracking #3794


MScottAdams:
ended_at: 2026-08-27T16:43:37.852Z

MScottAdams:
Residual pass on 313896c0 (commit 1 of #3794 only).

Class A leftover closed: 11 `dispatcher.test.ts` cases no longer get `foreign-repository-deny` for relative writes under a non-git payloadRoot. Unproven `--git-common-dir` lookup falls back to payloadRoot (AC); proven-different common-dir still denies. TypeScript + merge gates are green on this SHA.

ApplyPatch P1 3873543502 closed in-tree: declared path + patch-body paths are admitted together; a two-worktree span is refused.

Parked (not flipping back):
- Greptile 4/5 "unproven common-dir should fail closed" fights commit-1 AC and the class A test matrix. Fail-closed on unproven was 2880c809 and caused the 11 CI failures.
- SLizard file_scope/scope-relativization is commit 2.

Not merging at conf 4 with should-not-merge prose (#3225 / #3095 dogfood floor 5). Issue #3794 stays open.

MCP unavailable this session -- used `gh api` REST for review comments.

MScottAdams:
model: claude-opus-5-thinking-high-fast
role: parent

## Rebased onto master, and a way out of the Greptile fork

Rebased `fix/3794-worktree-effective-root` onto current master at `2140da28`. Master had moved six commits (#3781, #3774, #3806, #3811 and their lifecycle lands); the only conflicts were CHANGELOG, union-resolved via `task changelog:resolve-unreleased` across three commits. No code conflicts. The `dirty` state is cleared.

That leaves the real blocker: **Greptile at 4/5 on "unproven common-dir should fail closed"**, against the #3225 / #3095 dogfood floor of 5.

## The fork is narrower than it looks

The commit history shows this already swung once -- `a6d67a39` fails closed on unproven, `48027ddb` reverts to fall-back -- and fail-closed cost 11 CI failures.

Worth stating plainly, because I think it resolves the disagreement: **falling back to `payloadRoot` on unproven is not a weakening. It is exactly today's behaviour.** Every gate uses `payloadRoot` right now. This PR's contribution is the *new* protection -- proven-different common-dir denies -- layered on top. Fail-closed on unproven would be a brand-new denial class that did not exist before this PR, which is precisely why eleven tests broke.

Greptile is applying a correct instinct in a place where the fallback is status quo rather than a hole.

## A sharper rule that should satisfy both

The eleven failures were all **relative writes under a non-git `payloadRoot`**. That is the tell. "Unproven" is currently one bucket holding two very different situations:

| Case | Meaning | Right answer |
|---|---|---|
| `payloadRoot` is **not a git repo** | there is no common-dir to compare; containment is meaningless | fall back -- nothing to protect |
| `payloadRoot` **is** a git repo, target lookup fails | a genuine unknown where containment *does* matter | fail closed |

Splitting on that gives Greptile the fail-closed behaviour it is asking for in the only case where it means anything, keeps the acceptance criterion intact, and does not reintroduce the eleven failures -- because those all sit in the first row.

It is also a better rule on its own terms. "We could not determine containment in a context where containment applies" is a real non-decision. "There is no containment question here" is not.

## Alignment with the #3785 finding

This is the same taxonomy that arc landed on (synthesis 5434430152): three diagnostic states collapse to **two authorization classes** -- a verdict was rendered, or none was. Applied here, a failed lookup inside a git repo is a non-decision and should fail closed; a non-git root never posed the question, so there is no verdict to miss.

Adopting that split makes this PR consistent with the answer #3785 reached rather than looking like an exception to it.

## Suggested order

1. Land the split above as one commit, with the two cases named in a test each.
2. Re-read Greptile from the score comment body on the new head, confirming it cites the current SHA.
3. If it still holds at 4/5 on this point, that is a genuine standing disagreement rather than an unaddressed finding -- and the disposition should be recorded on the PR the way the SLizard false positives were on #3775 and #3784, rather than merged silently or stalled indefinitely.

SLizard's `file_scope` / scope-relativization finding stays parked for commit 2, per the recut. That is correct and should not be reopened here.

Standing-authority note: rebase performed under the overnight mandate at #3801 comment 5434371637. The suggestion above is a recommendation, not a change -- no code was modified.

PR #3793, opened 2026-08-27
fix(policy): harden project-definition mutation

dbcall2:
## Summary

This follow-up hardens the PROJECT-DEFINITION mutation and diagnostic paths delivered by #3786. PR #3786 was externally merged before the final independent adversarial review completed, so these later findings were not part of that merge.

The canonical completed #3609 lifecycle record is already on `master` with verified acceptance evidence for #3786. This PR carries only the undelivered hardening and a separate active lifecycle scope; it does not claim the follow-up is already delivered. Scott's unmerged `6bf4e67` was lifecycle-only and contained no production fixes.

## Final adversarial findings addressed

1. **P1 stale-reap ABA:** publish a prepared non-empty directory lock atomically, and require a stale reaper to remove the unique owner entry before removing the directory. A delayed reaper cannot remove a replacement lock.
2. **P1 legacy/malformed sidecars:** recover legacy NUL, empty, partial, malformed, and short owner metadata after a safety grace period without stealing an identifiable live owner. Failed owner creation cleans its own prepared inode.
3. **P2 path binding:** resolve the configured PROJECT-DEFINITION target once, then pass that exact artifact identity through lock, read, and atomic write. The legacy-layout fallback now catches only its intended missing-layout error.
4. **P2 hostile conformance keys:** retain raw semantic finding keys while rendering bounded, redacted, control-safe diagnostics.
5. **P3 shadow semantics:** preserve the true sorted `shadowedSubKeys` inventory and expose a separate bounded/redacted display summary.

## Verification

- `task verify:ac -- xbrief/active/2026-08-27-3609-post-merge-hardening.xbrief.json` — 6/6 clauses and the literal six-file oracle pass
- focused Vitest oracle — 6 files, 145 tests pass
- `pnpm exec tsc -b`
- `task verify:vbrief-conformance`
- `task packs:render -- --pack skills --check`
- `task verify:forward-coverage`
- `task check`
- `task verify:orphan-active -- --issue 3609`
- `DEFT_TRIAGE_REPO=deftai/directive task verify:completed-tracked -- --issue 3609 --tip upstream/master`

The local full suite completed with 15,106 passing and 6 skipped tests. Its only four failures are host-specific `session-coda` footer assertions caused by this machine's missing `~/.openclaw` pin warning; the same unrelated environment-only failures existed before this diff.

Tracking: #3609

Issue #3609, opened 2026-08-21
BLOCKER: bug(setup): Phase 2 emits bare plan.policy that vBRIEF conformance rejects


dbcall2:
## Summary

The Deft 0.105.0 setup skill requires a greenfield project definition to store the branch-policy decision at bare `plan.policy.allowDirectCommitsToMaster`. The same release's conformance engine rejects bare `plan.policy` and defines `plan["x-directive/policy"]` as the canonical storage key.

This leaves a newly onboarded consumer unable to pass the standard conformance gate without running a migration immediately after setup.

## Reproduction

1. Initialize a greenfield Deft 0.105.0 consumer.
2. Run `deft-directive-setup` Phase 2.
3. Choose branch-based workflow, which records `allowDirectCommitsToMaster: false`.
4. Inspect `xbrief/PROJECT-DEFINITION.xbrief.json`:

```json
{
  "plan": {
    "policy": {
      "allowDirectCommitsToMaster": false
    }
  }
}
```

5. Run `task deft:verify:vbrief-conformance`.
6. The gate fails with:

```text
xbrief/PROJECT-DEFINITION.xbrief.json [plan] bare key 'policy' at plan
```

## Conflicting contracts in the shipped release

- `.deft/core/skills/deft-directive-setup/SKILL.md` instructs the agent to write `plan.policy.allowDirectCommitsToMaster`.
- `@deftai/directive-core/dist/policy/plan-extensions.js` defines `PLAN_POLICY_KEY = "x-directive/policy"` and describes bare `policy` as legacy input only.
- `@deftai/directive-core/dist/vbrief-validate/conformance.js` has no bare-key exception and rejects `policy`.

The writer contract and verifier contract disagree within the same version.

## Actual behavior

A setup-generated project definition remains readable through a legacy fallback but is invalid under the required conformance gate.

## Expected behavior

Setup should write the canonical namespaced policy key through the shared policy writer. A newly generated project should pass conformance without a follow-up migration.

## Acceptance criteria

- Setup writes `plan["x-directive/policy"].allowDirectCommitsToMaster`.
- Setup re-entry handles namespaced-only, legacy-only, and conflicting dual-key inputs safely.
- A greenfield setup integration fixture runs `verify:vbrief-conformance` successfully.
- Documentation distinguishes the operator-facing `plan.policy.*` name from the stored namespaced extension key where necessary.
- Existing legacy consumers retain an explicit migration path.

## Environment

- `@deftai/directive` 0.105.0
[...]

PR #3832, opened 2026-08-28
Harden PROJECT-DEFINITION mutation identity and mixed-version locking (#3796)

MScottAdams:
## Summary

Lock recovery for `PROJECT-DEFINITION` used to delete whatever occupied the lock pathname. When an older deft version held that pathname with a plain file, deleting it released the only name that version recognises while its critical section was still live — two processes could then enter mutation at once. `rename` is atomic name movement, not compare-and-remove, so moving the file aside has the same defect: rename-to-quarantine was refuted during the design arc, and the bound contract is fail-closed.

This implements that contract.

**Legacy file sidecars are preserved, never moved.** A file at the public lock pathname is never unlinked or quarantined. Acquisition exhausts one bounded budget and then reports constant-label manual recovery steps. On Windows the guard has to be explicit: POSIX `rename` refuses to move a directory onto a file (`ENOTDIR`), but `MoveFileEx` is called with `MOVEFILE_REPLACE_EXISTING` and replaces it. Absence of a non-directory at the pathname is now a precondition of publication rather than an accident of POSIX semantics.

**Current-version locks publish a whole directory.** A fully-materialised, non-empty owner directory is renamed onto the pathname, so no contender observes a partial lock and a delayed `rmdir` from an older generation cannot remove a published replacement. The direct barrier regression asserts exactly that.

**Reap requires an unambiguously dead owner.** A live pid, a possibly-reused pid, an unreadable owner record, and a malformed lock directory all fail closed to manual recovery. Elapsed time is no longer read as evidence of safety, so the age-based grace is gone.

**One monotonic acquisition budget** is read once and shared by every retry and recovery branch; critical-section execution stays outside it.

**One mutation capability owns artifact identity.** `withProjectDefinitionMutation` binds load, parse, and persist to the path captured when the lock was taken. All 18 core lock call expressions, the 2 `packages/cli` policy writers that previously wrote with no lock at all, and three module-local duplicates of the resolver / loader / write sink now route through it. A source inventory (`verify-source/project-definition-mutation-boundary.ts`) fails closed on any new raw resolver, lock, or write bypass and pins the per-file census.

**Diagnostics are labelled.** Exact semantic keys stay whole for machine inventories (the doctor finding's `shadowed_sub_keys`, conformance finding keys); bounding and redaction happen only at the display boundary. Lock and loader messages name a configured artifact by a constant label instead of echoing its path.

### Notable findings while implementing

- The Windows `rename(dir, file)` clobber above was found by running the new regression on win32; it is not reachable from the Darwin probes that shaped the design. This is why the verified-claims table named a Windows CI regression as a delivery precondition — the CI job now runs the protocol test on `windows-latest`.
- The mechanized inventory immediately surfaced four raw write bypasses and three module-local duplicate loaders/write sinks that a name-based migration missed, including one in `render/project-render.ts` that locked on the project root while resolving and writing an independently-derived path.

### Residual, disclosed

On Windows a legacy client that creates the sidecar in the window between the pre-publish check and the rename can still have its file replaced. Nothing in portable Node makes the publish itself conditional on the destination; the pre-check closes every observable case and the code says so at the call site. POSIX has no such window.

## Related Issues

Tracking: #3796
Refs #3609, #3786, #3793

Supersedes the closed, unmerged PR #3793 — this is a fresh branch and a fresh PR, per the bound contract. #3793 is reference only.

## Checklist

- [x] `/deft:change <name>` — N/A: single scope xBRIEF (`xbrief/active/2026-08-28-3796-...`) covers every change; the 8 acceptance clauses are bound to their shipped artifacts.
- [x] `CHANGELOG.md` — entry added under `[Unreleased]`.
- [x] `ROADMAP.md` — N/A.
- [x] Tests pass locally — `task check` exit 0: 1055 test files passed / 6 skipped (1061), 14992 tests passed / 152 skipped (15144). `task verify:ac` passed with 3 acceptance commands run verbatim, each exit 0.

## Post-Merge

- [ ] **Verify issue auto-close**: `gh issue view 3796 --json state --jq .state`.
- [ ] `task scope:complete` on the active brief, then `task verify:orphan-active -- --issue 3796` and `task verify:completed-tracked -- --issue 3796`.


MScottAdams:
## Merge-readiness audit (#3796)

Recording why this merges while four P1-labelled inline comments are still visible on the head commit.

**Greptile (review of record)** — `Last reviewed commit: 7f4436d1d50837c11486ecd950343e1b8995e44b` (= head), Confidence 5/5 against the dogfood floor of 5, 0 P0 / 0 P1, `Greptile Review` check-run `completed` / `success`. `task pr:watch -- 3832 --one-shot` returns `CLEAN` with `sha_match=true`, `ci_failures=0`, `clean_gate_holdout=null`. `task pr:merge-ready -- 3832` exits 0 (`MERGE-READY`, `Advisory should-not-merge: False`).

**The four SLizard P1s are from the previous pass, not this head.** Each was created at `02:34:54Z`–`02:34:57Z` with `original_commit_id = d56e14d4`; the fix commit `7f4436d1` is dated `02:48:38Z`. GitHub re-anchored them onto the new head because the lines still exist. SLizard's pass that actually reviewed this head ran `02:49:19Z` → `03:01:02Z`, completed `neutral`, and raised exactly one new comment — a P2, not a P1.

Their premise ("exported from a production module but only referenced from test files") is also false on this head: `evaluateProjectDefinitionMutationBoundary` reads all four census constants in the same production module (`project-definition-mutation-boundary.ts:366-378`). That was the substance of fix batch 1 — the module now owns its own fail-closed verdict instead of exporting raw numbers for a test to re-derive.

**The remaining P2** (50% confidence, path traversal) asks to "verify `mutation.persist()` enforces directory containment internally". It does: `persist` calls `assertWriteTargetSafe` before `atomicWriteProjectDefinition`, rooted at the project root for an in-tree artifact and at the artifact's own directory for a `DEFT_PROJECT_PATH` override that legitimately resolves outside it. Both paths are covered in `project-definition-mutation.test.ts` (symlink refusal, out-of-tree configured artifact). No change needed.

**CodeQL** raised three dead-store findings on `lastReason`; all three were real and are fixed in fix batch 1 by the `yieldOrFail(reason)` restructure. `Analyze (javascript-typescript)` is `success` on this head.

**Required checks on `7f4436d1`**: TypeScript (build + lint + test) `success`, Merge gate (task check) `success`, Go (test + build) `success`, Windows (task dispatch regression) `success` — the last one includes the new `#3796` lock-protocol step, which is the Windows regression the verified-claims table named as a delivery precondition.

Review-cycle iterations used: 1 of 3.

Issue #2442, opened 2026-07-11
feat(build,swarm,main): dual stop rule — success gate plus failure/budget stop on multi-iteration work


visionik:
feat(build,swarm,main): dual stop rule — success gate plus failure/budget stop on multi-iteration agent work

## Summary

Loop engineering practice (and the article "Loop Engineering In 5 Minutes") requires every long-running autonomous loop to have two stop conditions: success (goal met) and failure (retries exhausted, no progress, budget exceeded). Directive has strong success-shaped gates (task check, AC, STOP on plan mismatch) but does not systematically require a failure/budget stop. Agents can thrash for many iterations on unsolvable problems and burn tokens without escalating.

## The gap

- #1613 covers STOP conditions when plan preconditions fail (reality mismatch)
- Dual stop is different: even when the plan is still "valid," the agent must quit after N failed attempts, N identical no-progress outcomes, or a token/time budget
- Swarm and build multi-iteration repair paths can retry without a mandatory failure envelope

## Proposed rule

For any multi-iteration autonomous work (build quality loop, swarm repair, research fan-out, review fix cycles):

1. Success stop: goal/AC/checker met
2. Failure stop: at least one of:
   - max iterations (e.g. 3 or 15 depending on task class)
   - no-progress detector (same outcome N times in a row)
   - explicit budget (time or token) if available

On failure stop: halt, report what was tried, what is missing, and what human decision is needed. Do not continue silently.

## Proposed changes

- main.md -- dual stop principle under agent behavior / decision making
- skills/deft-directive-build/SKILL.md -- multi-iteration implementation and pre-pr-style loops carry failure stop
- skills/deft-directive-swarm/SKILL.md -- repair and monitor loops carry failure stop (composes with #2439 minimal repair)
- Optional: vBRIEF plan item fields for maxIterations / failureStop text

## Acceptance criteria

- Documented dual stop (success + failure) for multi-iteration agent work
- Build and swarm skills name a default failure stop (iteration and/or no-progress)
- Failure stop produces an operator-visible report, not silent continuation
- Single-turn tasks remain exempt (not every task is a loop)

## Related

- Extends #1613 (STOP conditions — precondition halt; this is retry/budget halt)
- Related to #1499 (verification independence — checker should be corrective)
- Related to #2439 (minimal subgraph repair — repair loops need failure stops)

## Source

[...]

PR #3834, opened 2026-08-28
fix(design-critique): publish a closed completed-arc citation grammar with a position predicate

### LINKED ISSUE #3831: bug(design-critique): the completed-arc citation grammar is unpublished, too narrow and too wide at once, and split across two parsers
> **Recut 2026-08-28** to the bound contract in successor lean `5447023339`, synthesis accepted `5447750579`. The original filing diagnosed this as "the parser is too strict" and proposed widening `CITE_RE`. A three-critic panel refuted that remedy 3/3 and measured two further defects the filing missed. History is in the thread.

## Summary

The completed-arc citation grammar is **unpublished**, **simultaneously too narrow and too wide**, and **split across two divergent parsers**. `evaluateCompletedArcRecord` is the ingest-clearance predicate for every design-critique arc, so all three properties decide what counts as a completed arc repo-wide.

## The four measured defects

**1. Too narrow — a well-formed citation is rejected.** `CITE_RE` cannot parse an id inside a code span. The live #3796 synthesis cites its lean and table correctly and extracts zero ids, blocking as `lone-shape` with a message asserting it "does not cite an accepted successor lean." Backticks are the sole cause *in that body*, but bold, `#`-prefix, italic, paren, quoted, and non-canonical display links all fail too — including a bare id whose **keyword** is bolded, though `**Lean:**` is the mandated house style.

**2. Too wide — non-affirmative occurrences already clear.** Fenced blocks, blockquotes, and explicitly **negated prose** (`do not use successor lean N`) all parse as affirmative citations today. There is no author predicate, so an arbitrary forged comment clears ingest. Accepting an id is not the same as accepting a citation, and the current parser does not distinguish them.

**3. Fail-open — the table requirement is silently waived.** `extractCitedCommentIds` and the separate `claimedTable` regex at `completed-arc-record.ts:158` are two parsers answering one question. A bare lean id plus a decorated table id returns `{"status":"complete","citedTableId":null}` — the verified-claims-table citation is read as *no table claimed*. This is the inverse of defect 1 and strictly worse.

**4. Document-order latching contradicts the contract.** `.find()` at `:144` latches the first lean-shaped id. `Supersedes lean OLD. Bound contract is lean NEW.` blocks as `missing-record`, asserting "a later successor lean recut the arc" — a recut that never happened. The same two sentences reordered complete. The Successor lean section of the contract **mandates** citing "the prior lean it supersedes," so obeying the contract in natural order blocks ingest.

## Why strictness is not the alternative

The contrary position — that a strict grammar is a stronger oracle — is empirically dead. A whole phrase inside a code span already parses while a backticked id does not, so today's accept set is an arbitrary partition rather than a machine contract. With no author predicate and no position check, strictness buys none of the safety it appears to.

Equally, widening is not the remedy. Given defects 2 and 3, a broader accept set is a net safety regression.

## Bound remedy

1. **Publish a closed citation grammar in the contract.** It currently requires "citing the accepted successor lean" and specifies no lexical form at all; the grammar exists only in `CITE_RE` and the fixtures. Bare decimal, balanced single-backtick decimal, and canonical permalinks are the defensible set.
2. **Add a position predicate.** Blockquoted, fenced, struck-through, and negated occurrences must not clear. Prior art is `classifyHit` in the PR closing-keywords detector, which classifies a match's position rather than deleting spans. The obvious normalisers cannot be reused — they delete code spans *with their contents*, destroying the digits.
3. **One parser.** `extractCitedCommentIds` and the `:158` `claimedTable` check must share it. Two regexes answering one question is not a safe oracle.
4. **Fix document-order latching** so citing the superseded lean cannot block, and correct the `missing-record` detail that asserts a recut.
5. **Diagnostics after the grammar.** A message naming accepted forms is only possible once they exist. Adding "formatting may be the cause" to today's message preserves the ambiguity and closes nothing.
6. **State which convention governs.** The intake layer encodes the opposite polarity by deliberate design.

## Acceptance

- [ ] The contract publishes the closed set of accepted citation forms
- [ ] A synthesis citing `` successor lean `NNNNNNNN` `` parses; #3796's original synthesis parses without being rewritten
- [ ] Blockquoted, fenced, struck-through, and negated occurrences do **not** clear
- [ ] Citation extraction and table-claim detection share one parser; mixed-form matrices assert the typed `citedTableId`
- [ ] Citing the superseded lean before the bound contract does not block
- [ ] The `missing-record` detail does not assert a recut that did not occur
- [ ] Fixtures cover decorated ids, bolded keywords, and each negative position class
- [ ] No regression: a genuinely uncited synthesis still blocks

## Out of scope

- #3796's arc content, unblocked separately via an already-accepted permalink spelling with no code change.
- #3829, the acceptor-alias legibility gap.
- #3824, self-reported model slugs.
- Whether catalog chips gate ingest — #3806 settled that they do not.

## Panel

A `claude-opus-5-thinking-high` `5447002763` · B `cursor-grok-4.6-high-fast` `5446977097` · C `gpt-5.6-sol-medium` `5446965147`. Dispatch SHA `c99f6159`. All three self-reported leads matched the routing recorded on the panel-deposit.

Refs #3806, #3796, #3156.

MScottAdams:
## Summary

The completed-arc citation grammar was unpublished, simultaneously too narrow and too wide, and split across two divergent parsers. `evaluateCompletedArcRecord` is the ingest-clearance predicate for every design-critique arc, so all three properties decided what counted as a completed arc repo-wide.

This ships the remedy bound by the #3831 three-critic panel: a published closed grammar, a position predicate, one parser, membership-based lean selection, and diagnostics written after the grammar exists. It is **not** the original filing's remedy (widen `CITE_RE`), which the panel refuted 3/3.

## What changed

- **Published grammar.** `content/contracts/design-critique.md` gains `## Citation grammar`: the closed accepted set (bare decimal, colon form with optional space, balanced single-backtick decimal, emphasised keyword, and the two canonical comment permalinks), the refused positions, and a paragraph recording which code-span convention governs here versus the intake scanners.
- **Position predicate.** New `packages/core/src/design-critique/citation-grammar.ts` classifies where a match landed rather than stripping spans, and refuses fenced, inline-code-keyword, blockquoted, struck-through, and negated occurrences. `isInsideCodeFence` is now exported from `pr-closing-keywords/detect.ts` and reused; that detector is the prior art the panel identified. The obvious normalisers cannot be reused — they delete a code span with its contents and destroy the digits.
- **One parser.** The second literal regex at `completed-arc-record.ts:158` is gone. `extractCitedCommentIds` and the verified-claims-table claim now read the same typed parse.
- **Set membership, not document order.** `.find()` over extraction order is replaced by membership against the latest successor lean, so citing the superseded lean first — which the contract's `## Successor lean` section requires — no longer blocks.
- **Observed diagnostics.** `lone-shape` reports how many id-shaped runs were present, how many keyword-anchored occurrences were refused and by which class, and names the accepted forms. `missing-record` names the latest lean and what the existing records cite instead of asserting a recut.

## Measured before and after

Against the live #3796 thread and the shipped module:

| Case | Before | After |
| --- | --- | --- |
| Live #3796 synthesis body `5439970683` | `cited=[]`, `blocked/lone-shape` | `cited=[5439956893 table, 5439547122 lean]`, `complete` |
| Negated prose `do not use successor lean N` | `cited=[N]` | `blocked` |
| Blockquoted, fenced, struck-through, whole-phrase code span | `cited=[N]` each | `blocked` each |
| Bare lean plus a backticked table id | `complete` with `citedTableId: null` | `blocked/missing-table-cite` naming the id |
| `Supersedes lean OLD. Bound contract is lean NEW.` | `blocked/missing-record` | `complete` (both clause orders) |

The whole live #3831 thread also evaluates `complete` on `synthesisCommentId 5447750579`, `citedLeanId 5447023339`, `citedTableId 5447745771` — the arc's own record, written in permalink form.

## Gate integrity (#3156)

This edits the ingest-clearance predicate, so the net direction matters. The accept set **narrows**: it adds a backticked id token and keyword-side emphasis, and removes every fenced, blockquoted, struck-through, negated, and whole-phrase-code-span occurrence that cleared before. No gate definition, verifier, coverage floor, or policy flag was weakened to go green; the two `task check` failures seen locally were an ambient `DEFT_SESSION_ID` leaking into session fixtures, not this change.

## Related Issues

Closes #3831

Refs #3806, #3796, #3156.

## Checklist

- [x] `/deft:change <name>` — N/A: the change set is the bound remedy recorded on the #3831 arc (successor lean `5447023339`, verified-claims table `5447745771`, synthesis accepted `5447750579`), which is the operator-confirmed proposal surface for this work.
- [x] `CHANGELOG.md` — entry added under `[Unreleased]`.
- [x] `ROADMAP.md` — N/A: generated projection, refreshed on release.
- [x] Tests pass locally — `task check` exit 0. Full suite: 1064 test files passed, 6 skipped; 15231 tests passed, 152 skipped, 0 failed. Coverage: statements 88.21%, branches 81.41%, functions 93.58%, lines 89.12%. Changed modules: `citation-grammar.ts` 100% lines / 94.54% branches, `completed-arc-record.ts` 100% lines / 92.53% branches.

## Post-Merge

- [ ] Verify issue auto-close on #3831.
- [ ] `task scope:complete` on the active brief, then `verify:orphan-active` and `verify:completed-tracked`.


MScottAdams:
## Review-cycle halt: #2442 fix-batch cap reached

Three fix batches landed against this PR. Two new P1 findings arrived on the third, so the loop stops here rather than continuing without consent.

**HEAD:** `0a23378d3dd14b8feda6e4fe8136ef1f93f15e0e`
**Greptile confidence:** 3/5 against a resolved floor of 5 (`minGreptileConfidence`, source `dogfood`).
**CI:** TypeScript (build + lint + test) and Merge gate (task check) pass. Local `task check` exit 0 — 1067 test files passed, 6 skipped; 15280 tests passed, 152 skipped, 0 failed.

### Addressed across the three batches (5 P1)

| Finding | Batch | Fix |
| --- | --- | --- |
| Indented fences bypass refusal | 1 | Local indent-aware fence stack; the borrowed opener was anchored at column zero |
| Markers misclassify affirmative citations | 1 | Explicit negation closing within three plain words of the keyword |
| Invalid fence clears ingest | 2 | A closer now requires an empty info string |
| Multiline code clears ingest | 2 | Inline-code and strikethrough scans read the enclosing block, not one line |
| Lazy blockquotes clear ingest | 3 | The blockquote marker carries to the blank line that ends the quote |

The earlier threads for these still show open because Greptile does not resolve its own threads; their comment ids are unchanged from the round that raised them.

### Outstanding, not addressed (2 P1)

Both are **fail-closed over-refusals** — they block a valid record rather than clearing an invalid one — and both live in `packages/core/src/design-critique/citation-grammar.ts`, which this PR already owns, so both are class A in-AC residual.

1. **Negation of an intervening verb** (`3877895228`). `we cannot deny that successor lean N binds` reads as a negated citation. Remedy: disqualify a filler that introduces a complement clause (a trailing `that`), and shrink the filler bound.
2. **Blockquote state survives a fence** (`3877895230`). A `>` line inside a fenced example sets the quoted-block flag, and a citation after the closing fence with no blank line is refused. Remedy: do not set the flag from lines inside an open fence, and reset it when a fence opens or closes.

### SLizard advisory P2 — verified false positive

`3877798005` reports that `evaluateCompletedArcRecord` still uses a legacy `CITE_RE`. It does not: `CITE_RE` was deleted in this PR, and both call sites read `scanCitations` (`completed-arc-record.ts` lines 90 and 193). SLizard is advisory only and this verdict was checked against the source rather than accepted on the label.

### Resume

```
RESUME: residual=greptile_p0_p1 leftover=A conf=3/5 floor=5 standing=one-shot PR=https://github.com/deftai/directive/pull/3834
```

One more fix batch closes both remedies above. Fresh operator consent is required to reopen the loop past the `#2442` cap — see `deft-directive-review-cycle` § Operator follow-up after dual-stop / hard stop.

MScottAdams:
## Fix batch 4 — both remaining P1s closed, rebased onto master

**HEAD:** `d251862e038477a1df2a1a380d96032ea32c3504` (rebased onto `f1886c8d`; the CHANGELOG conflict was resolved with `task changelog:resolve-unreleased`, not by hand).

### Addressed (2 of 2 outstanding P1, 0 deferred)

Both were fail-closed over-refusals in `packages/core/src/design-critique/citation-grammar.ts` — class A, in-AC, on a file this PR already owns. Both reproduced against the code before the fix, and the new tests fail on the prior head.

| Thread | Finding | Fix |
| --- | --- | --- |
| `3877895228` | Negation of an intervening verb. `we cannot deny that successor lean N binds` discarded the only lean cite and blocked as `lone-shape`. | A negated **verb of denial** affirms the citation. The disqualifying set is closed and published — `deny`, `doubt`, `dispute`, `contest`, `question` — anchored to the end of the sentence segment. |
| `3877895230` | Blockquote state survived a fence. A `>` line inside a fenced example set the quoted-block flag, so a citation after the closing fence with no blank line was refused. | A fence delimiter ends the quote block, and a line inside an open fence never opens one. |

**A trailing `that` is not the signal.** `that` is also a determiner, so keying on it alone would have re-admitted a genuinely negated occurrence. `do not use that successor lean N`, the cleft `the record is not that successor lean N`, and a second negation inside the complement clause (`we do not doubt that this does not bind successor lean N`) all stay refused, each with a regression test.

The predecessor's second suggested remedy — shrinking the filler bound — was **not** applied: it would break the existing `this does not cite the bound successor lean N` refusal, which is a three-word filler the panel requires refused.

### The accept set does not widen on net

Every occurrence class the panel required refused was replayed through `scanCitations` on the prior head and on this one. All 27 return the **same** rejection class:

| Class | Shapes | Before | After |
| --- | --- | --- | --- |
| fenced | 7 (plain, indented, tilde, mismatched closer, closer with info string, permalink, fence containing a `>` line) | `code-fence` | `code-fence` |
| blockquoted | 3 (marked, lazy continuation, blockquoted fence) | `blockquote` | `blockquote` |
| struck-through | 2 (same line, across a newline) | `strikethrough` | `strikethrough` |
| whole-phrase code span | 4 (single, double-backtick, across a newline, unbalanced run) | `inline-code` | `inline-code` |
| negated | 11, including the two `that` shapes above | `negation` | `negation` |

The only verdict changes in the whole matrix are the two over-refusals themselves.

### Gate integrity (#3156)

The narrowing is a correctness fix to an over-refusal that blocks valid records, raised by review — not a gate weakened to go green. Both conditions are **published** in `content/contracts/design-critique.md` under `## Citation grammar` rather than living only in the regex, and `design_critique_contract.test.ts` locks the new prose.

### Gates

`task check`: exit 0 — 1067 test files passed, 6 skipped (1073); **15304 tests passed, 152 skipped, 0 failed**. `task verify:forward-coverage`, `task verify:encoding` clean.

### Already-verified, not re-investigated

SLizard P2 `3877798005` (`evaluateCompletedArcRecord` still uses a legacy `CITE_RE`) remains a confirmed false positive: `CITE_RE` is deleted and both call sites read `scanCitations` (`completed-arc-record.ts` lines 90 and 193). SLizard is advisory.

The five P1s addressed in batches 1–3 still show as open threads because Greptile does not resolve its own threads; their `original_commit_id` values predate this head.

MScottAdams:
## Fix batch 5 — closes the widening batch 4 opened

**HEAD:** `d4414f9c06e4e1f7684359eeba72758126703077`.

The 4/5 finding is valid and was my regression, not a pre-existing one. Batch 4's denial-complement carve-out read only the text *before* the citation keyword, so `we do not doubt that successor lean N does not bind` cleared ingest from a synthesis that says the lean does not bind.

The carve-out is now a **suspension, not an acceptance**:

- It runs only after the prefix negation has already fired, so it can never refuse on its own and cannot widen the *refused* set either.
- When it fires, the complement clause decides — a negation anywhere in the rest of that sentence keeps the refusal.
- The forward scan runs to the sentence break rather than to a clause boundary. That refuses slightly more than it strictly must inside this one carve-out, which is the fail-closed direction, and every such body was already refused before batch 4.
- The negation marker set is built once and shared by the anchored prefix rule and the unanchored complement scan, so the two cannot drift.

Four new regression tests pin the shape Greptile named: `does not bind`, `is not the bound contract`, `no longer binds`, and `doesn't bind` after the citation. A fifth pins that an unrelated negation in the *next* sentence does not refuse.

### Net effect of batches 4 and 5, measured

Replayed a 35-shape matrix through `scanCitations` against the pre-batch-4 head and this head. **4 verdicts change**, and all four are the named over-refusals — two denial complements and two fenced-quote-marker cases. The 31 required refusals return the same rejection class, including the four new denial-after guards.

`task check`: exit 0 — 1067 test files passed, 6 skipped; **15306 tests passed, 152 skipped, 0 failed**.

The contract carries the new condition under `## Citation grammar`, and the contract test locks it.

Issue #3780, opened 2026-08-26
bug(triage): a comment-id ceiling pins position but not content, so an edited or deleted input silently breaks a pass


MScottAdams:
## Summary

A design-critique pass declares a comment-id **ceiling** so its verdict is reproducible. A ceiling pins a *position* in the thread. It does not pin *content*.

Comments are editable and deletable after a pass reads them, including the ceiling comment itself, and a REST body fetch shows no trace that an edit happened. So a pass can satisfy every reproducibility requirement it has and still be unreproducible.

This is not hypothetical. **Two declared ceilings from the last twenty-four hours are already gone:**

| Comment | Role | State now |
| --- | --- | --- |
| `5420630035` | declared ceiling, #3729 critique pass | **HTTP 404** |
| `5420630232` | declared ceiling, #3611 critique pass | **HTTP 404** |

Both passes completed, produced accepted findings, and fed recuts. Neither can now be re-derived from its stated inputs.

## Who broke it, and why that matters

The parent orchestrating those passes deleted them, during a repair of comments that had been posted with a malformed `gh api -f body=@file` invocation. The deletion was legitimate cleanup of its own bad output.

That is the important part. The corruption did **not** come from a second agent arriving mid-pass — it came from the pass owner mutating its own inputs. Any design that gates *arriving* writers leaves this untouched, which is why it is filed separately from #3607 rather than inside it.

## Why position-based pinning cannot fix it

- Deletion removes the anchor entirely; a ceiling id that 404s cannot bound anything.
- Edits are worse than deletes because they are silent: `updated_at` moves, but a consumer reading `body` has no baseline to compare against and no reason to look.
- The failure is invisible on both sides. The synthesizer does not learn its inputs changed, and a later reader cannot tell a reproducible pass from a corrupted one.

## Remedy sketch

Record content, not just position. At pass start, capture a manifest of the comment ids in scope with a hash of each body, plus the ceiling id. A synthesizer, an auto-stamp, or a later auditor can then re-fetch and detect three distinguishable states: unchanged, edited, deleted.

Deliberately modest — this is a **detection** mechanism, not prevention. Nothing can stop a maintainer editing a comment, and nothing should. The defect is that corruption is currently silent, not that it is possible.

Open questions for the design pass:

- Where does the manifest live? An artifact under `xbrief/`, a fenced block in the pass-open marker, or local pass state.
- Does the ceiling comment's own hash get special handling, since its disappearance is unrecoverable rather than merely detectable?
[...]

Issue #3476, opened 2026-08-18
drive-to:merge-ready DONE leaves completed xBRIEFs untracked on master (#3264 / #1358 recurrence)


MScottAdams:
## Summary

`drive-to: merge-ready` workers run `scope:complete` **after** the product PR merges. That move is filesystem-only. The completed xBRIEF stays **untracked** in the feature worktree. `verify:orphan-active` goes green (active/ is empty). `verify:completed-tracked` (#3264) is **not** on `task check` and is **not** in the drive-to DONE contract. Master never gets the artifact until a later land PR.

This is the #1358 / #3264 residue class firing again on the through-merge path those issues were meant to close.

## Field evidence (2026-08-18, deftai/directive)

Cohort `3470-3469-3454-through-merge`. Three `drive-to: merge-ready` leaves merged:

| Issue | Product PR | Merge | Completed file on origin/master after DONE? |
| --- | --- | --- | --- |
| #3454 | #3473 | `58234460` | no |
| #3469 | #3474 | `cbae6e53` | no |
| #3470 | #3475 | `1907fe76` | no |

Each worktree showed `?? xbrief/completed/2026-08-18-….xbrief.json` after the worker announced DONE + `scope:complete` + `verify:orphan-active` exit 0.

Parent declared the cohort done without Phase 6 `task swarm:finalize-cohort` / Step 2b commit+push to the delivery branch. Copied completed files onto the parent checkout (still on merged `chore/land-completed-3382-remainder`) — still untracked, still not on master.

## Root cause

Three contracts that do not meet:

1. **`scope:complete` does not git add/commit.** It only moves `active/` → `completed/` on disk (`packages/core/src/scope/`).
2. **Timing.** Drive-to doctrine says complete **after** merge. The product PR cannot carry a file that does not exist yet. A commit on the already-merged feature branch does not land on master.
3. **DONE is the wrong gate.** Preamble / #3429 require `verify:orphan-active` (active residue). They do not require `verify:completed-tracked` (delivery-tip tracked completed/cancelled). #3264 is a standalone verb, explicitly **not** on `check:framework-source`, so a land PR is not deadlocked — and so a leaf can exit DONE while master is still missing the record.

The designed land path already exists: Phase 6 Step 2b (#1358) and `task swarm:finalize-cohort` (`commitLifecycleMoves` + optional sweep PR). Drive-to leaves do not run it. The parent can skip it and still emit `done` under #2934 (terminal status with a reason).

[...]

Issue #2814, opened 2026-07-24
feat(review-cycle): PR-anchored review-owner lease replaces local review-monitor.json (#2810 class)


MScottAdams:
## Summary

Replace the machine-local `.deft/review-monitor.json` ledger (#2655) with a **PR-anchored review-owner lease** as the sole source of truth, so two maintainer agents on different machines cannot both claim review-cycle ownership of the same PR.

Observed pain: PR #2810 was double-babysat across maintainers because each clone had its own ignored `.deft/review-monitor.json`. Local register/verify cannot see cross-machine ownership. The file on this dogfood clone already accumulates many `ended_at: null` records (including #2810) that `verify` treats as stale after ~30 minutes but never cleans up — a graveyard, not a mutex.

## Problem

| Gap | Effect |
|-----|--------|
| Ledger is gitignored under `.deft/` | Invisible to other clones / maintainers / cloud agents |
| `registerReviewMonitor` last-writer-wins per PR | Second local agent silently steals; no conflict exit |
| No real `release` path in practice | Open records accumulate forever |
| `verify:review-monitor` only proves *this checkout* has a record | Passes while another human's agent is already driving fixes |
| Dual local+remote cache (rejected in design) | Would split-brain / go stale immediately across machines |

GitHub has **no first-class exclusive "work lease on this PR"** API. Assignees, review requests, labels, check runs, and Copilot `assign_to_agent` are soft or vendor-shaped — not a structured, stale-aware mutex for arbitrary Cursor/swarm review owners.

## Decision (locked for this issue)

- **Source of truth:** one sticky PR issue-comment with stable HTML markers, edited in place (same durability pattern as umbrella current-shape comments).
- **No local cache:** do **not** dual-write `.deft/review-monitor.json`. Delete / stop writing that file; `register` / `release` / `verify` speak only to GitHub.
- **Still local (out of lease scope):** `.deft-scratch/subagent-status/` heartbeats remain the *liveness* signal for the poller process on the host that spawned it — orthogonal to *who owns the PR*.

### Claim comment shape (illustrative)

```text
<!-- deft:review-owner -->
owner: <github-login>
monitor_agent_id: <stable-id>
head_sha: <sha-or-empty>
started_at: <iso8601-utc>
expires_at: <iso8601-utc>
platform_primitive: cursor-task|spawn_subagent|start_agent
<!-- /deft:review-owner -->
```

### Verb semantics

1. **`review-monitor:register`** (claim)
[...]

Issue #2307, opened 2026-07-05
fix(security): umbrella:current-shape emits attacker-controlled issue comments as authoritative agent state


MScottAdams:
## Summary

`umbrella:current-shape` selects the canonical `## Current shape (as of pass-N)` comment by **highest pass-N (tie-break highest comment id) across all commenters**, with no author/provenance check, and prints the comment body verbatim. Because AGENTS.md makes reading the current-shape comment a mandatory umbrella-planning step, any GitHub user who can comment can forge a higher-pass comment that agents then consume as authoritative state — durable workflow/prompt injection and false status/wave-order steering.

Surfaced by the scheduled application-security review at commit `6449c04e6633c14e00026b1ac52bfbfd2ca6d6c7` (finding #3, Medium).

## Attacker / input

Any GitHub user who can comment on the umbrella issue authors a `## Current shape (as of pass-N)` comment with a pass number >= the maintainer's.

## Impact

The forged higher-pass comment is printed verbatim and consumed by agents per the required umbrella-reading workflow, allowing durable workflow/prompt injection and false status / wave-order steering.

## Affected

- `packages/core/src/umbrella-current-shape/index.ts` — `selectCurrentShapeComment` ranks by pass/id with no provenance; `emitCurrentShape` prints `result.body` verbatim; `--strict` checks only section structure.
- `defaultFetchComments` / `mapCommentEntry` currently drop `user.login` / `author_association`, so provenance is not even carried through today.

## Remediation

1. Carry `user.login` and `author_association` through `mapCommentEntry`, then restrict `selectCurrentShapeComment` candidates to maintainer-authored comments (`author_association` in `OWNER` / `MEMBER` / `COLLABORATOR`), defeating the forged-higher-pass primitive.
2. Run the `scan()` quarantine/fence transform (`packages/core/src/cache/scanner.ts`) over the selected comment body in `emitCurrentShape` before printing.
3. Make `--strict` require BOTH structure (existing) AND provenance (a maintainer-authored current-shape comment was found); otherwise exit non-zero.

## Acceptance criteria

- [ ] A higher-pass current-shape comment authored by a non-maintainer is ignored; the maintainer-authored comment wins regardless of pass number.
- [ ] Selected comment body is scanned/fenced before emission.
- [ ] `--strict` fails when no maintainer-authored current-shape comment exists, even if a structurally-valid non-maintainer one is present.
- [ ] Regression test with a forged higher-pass non-maintainer comment.

## Related

[...]

PR #3864, opened 2026-08-28
feat(review-cycle): pass-open advisory marker on issue threads (#3607)

### LINKED ISSUE #3607: feat(triage,swarm): no cross-machine signal that a pass is open on an issue thread
## Problem

Structured passes -- triage, design-critique, review-response -- run on GitHub issue threads. Nothing tells an agent arriving at a thread that a pass is currently open on it.

Agents are dispatched per-worktree, but issues are the shared coordination surface. Cross-cutting work naturally lands on an issue someone else's pass owns. Worktree occupancy (#3433) cannot help: it is keyed on a path, and is single-machine by construction, while the population running these passes includes cloud and headless agents on other hosts.

## Observed on #3599

| Time (UTC) | Actor | Event |
| --- | --- | --- |
| 17:13:32 | Grok 4.6 | Triage analysis posted; becomes the critic's ceiling (`5372982342`) |
| 17:19:26 | Claude (other worktree) | "Additional evidence" comment -- inside the open pass |
| 17:21:46 | Grok 4.5 | Design critique; could not see the 17:19 comment |
| 17:23:57 | Grok 4.6 | Verified synthesis -- had to note the amendment was outside the critic's allowed inputs |
| 17:26:53 | Claude | Scope-widening comment, after synthesis closed |
| 17:29:24 | Claude | Split a child issue out, against the triage's explicit "do not decompose children" |

The last row is the cost that matters here. A decomposition was filed against a just-issued disposition **because the arriving agent had no way to know a pass had just closed on that thread**. Neither agent malfunctioned.

## What the round-1 panel established

Three critics, open charter. Synthesis: 5430557487.

**Blocking is refuted, 3/3.** An exclusive pass lock has no coherent owner -- the design-critique protocol puts a parent and N critics on one thread by construction, so "the pass owner holds the thread" names no actual actor. Measured across 27 critic rounds since 25 Aug, all sixteen mid-round writes were the **pass owner's own**, none from another actor, and the two highest-value ones were corrections that a hold would have suppressed while critics reasoned from refuted premises.

**The incumbent lease is the substrate.** `<!-- deft:review-owner -->` is already an *issue*-comment lease: every GitHub call in register / release / verify runs against the `issues/` REST namespace, `--pr` is a parameter name rather than a constraint, and `head_sha` is the only PR-shaped field and is optional. Extending it is roughly a rename, two fields, and a call-site.

**The gap is the trust boundary, not the keyspace.** The incumbent's author filter excludes `CONTRIBUTOR`-authored comments, and the second actor on this repository is exactly that association. Extending the lease without fixing the filter reproduces the invisibility this issue is about.

**The label option is disqualified.** A label cannot state its own expiry, and expiry is what makes a crashed or abandoned pass survivable.

**Crash-release is not inherited.** Occupancy's hazard was a missing release call site, not missing expiry -- it always had a TTL, and #3604 shipped a release verb. Passes last minutes, and an advisory marker carrying `expires_at` that self-clears on read needs no heartbeat. The hazard only returns if the marker blocks, which is the posture the panel rejected.

## Remedy

Extend the review-owner lease to issue-thread passes as an **advisory, observe-and-mark** signal.

- Set at pass start, carrying the pass kind, the owner, the declared ceiling, and `expires_at`.
- Cleared at synthesis; self-clears on read once expired.
- Oldest-id-wins for concurrent claims, matching the incumbent.
- Fix the `CONTRIBUTOR` author filter so the marker is visible across trust associations.
- An arriving agent **reads and is informed**. It may post. Guidance is to mark the comment as post-ceiling, or hold voluntarily.

Nothing blocks. The goal is that "a pass is open and your comment will fall outside its ceiling" is observable before the write, not that the write is prevented.

## Acceptance

- [ ] A pass-open signal readable by an agent in a different worktree and on a different host
- [ ] Set at pass start, cleared at synthesis, and self-clearing on expiry so an abandoned pass does not mark a thread forever
- [ ] Visible regardless of the author's trust association, including `CONTRIBUTOR`
- [ ] Arriving agents are informed and never blocked
- [ ] Reuses the `review-owner` lease rather than introducing a second mechanism, or records why that was not possible

**Removed from the original sketch:** "the synthesizer's *amended after ceiling* note becomes an exception path rather than a routine occurrence." That is unreachable by construction -- the design-critique contract **requires** a panel-deposit after the ceiling on every panel round. This issue's own deposit landed fourteen seconds past its own ceiling for exactly that reason.

## Out of scope

- **Reproducibility of pass inputs (#3780).** A ceiling pins a position, not content; comments can be edited or deleted afterward, including the ceiling itself. Two ceilings from this cluster already return 404. No coordination marker addresses that, and it is the higher-value defect.
- Blocking, holding, or any exclusive lock. Refuted 3/3.
- A label-based interlock. Disqualified above.
- Worktree occupancy (#3433). Different keyspace, single-machine.

## Related

- #3434 -- design-critique panel + verified synthesis (the protocol with the ceiling)
- #3452 -- review-response loop, same exposure on PR threads
- #2655 -- the review-monitor register / verify pair this extends
- #3780 -- the reproducibility half, split out of this issue
- #3599 / #3604 -- where this was observed

MScottAdams:
## What

Extends the incumbent `<!-- deft:review-owner -->` comment lease so a structured pass (triage, design-critique, review-response) can mark its **issue thread** as open. The mark is advisory observe-and-mark: it says a pass is open, who owns it, the declared ceiling, and when the mark expires. It never holds, blocks, or gates another actor's write.

No new mechanism, no lock, no label. The lease was already an `issues/`-namespace comment lease, so this is adoption plus two fields plus a call site.

## Why this shape

The #3607 panel refuted blocking 3/3: a pass has N+1 writers by construction (parent plus panel), so an exclusive hold names no actual actor, and 16 of 16 measured mid-round writes across 27 critic rounds were the pass owner's own. A label was disqualified because it cannot state its own expiry.

## Behaviour

- `kind: pass` on the shared marker carries `pass_kind`, `owner`, optional `agent_id` and `ceiling`, `started_at`, `expires_at`, `ended_at`.
- Expired or synthesis-cleared marks are ignored on read, so an abandoned pass never marks a thread forever and no heartbeat is needed.
- Concurrent marks resolve **oldest comment id wins**, matching the lease; the loser deletes its duplicate and is told which mark stands.
- `openPassMarker` returns `observed` when another owner already holds an open mark. The caller is informed and may still write. Nothing is held.
- Pass marks are a **disjoint population** on the same marker: a `kind: pass` body never parses as an ownership lease, so it cannot satisfy `verify:review-monitor` / `verify:l4-owner` and `registerReviewMonitor` never claims one as its lease anchor.

## Author trust boundary

The brief carried an unresolved divergence: the panel claimed the lease excludes CONTRIBUTOR-authored comments; the retro verified-claims table (5455218052) searched `github-lease.ts` / `lease-comment.ts` / `l4-owner.ts` for `author_association` / `OWNER` / `MEMBER` and reported "not reproduced".

**The filter does exist.** It is `isMaintainerAuthored(entry.authorAssociation)` at the top of `listReviewOwnerComments`, delegating to `MAINTAINER_ASSOCIATIONS` (`OWNER`, `MEMBER`, `COLLABORATOR`) in `packages/core/src/umbrella-current-shape/index.ts` (#2307). The earlier search missed it because the snake_case field name never appears in this directory. That location is now named in a code comment.

The boundary is split by consequence, and stated in the module:

- **Ownership leases gate.** `verify:review-monitor` / `verify:l4-owner` exit 0 on a live lease, so a forged lease defeats a fail-closed gate. The maintainer filter stays exactly as it was.
- **Pass marks do not gate.** A forged mark costs at most an unnecessary courtesy, so `listPassMarkerComments` reads every association including `CONTRIBUTOR` -- which is what the issue's acceptance requires.

Tests pin both halves using the real field instance: comment `5429316778` on PR #3775 is CONTRIBUTOR-authored and carries the marker. The pass reader sees it; the lease reader still drops it.

No exclusion was added that the code did not have, and no gate was weakened.

## Tests

`pnpm exec vitest run packages/core/src/review-monitor` -- 108 passed.
`pnpm exec vitest run packages/core/src/content-contracts` -- green in the combined run (3728 passed).
Module coverage after the change: `github-lease.ts` 98.2% statements / 88.9% branches, `lease-comment.ts` 96.0% / 90.4%; every remaining uncovered line is pre-existing.

## Out of scope

Content pinning and input reproducibility belong to #3780. No thread interlock was added to `content/contracts/design-critique.md` -- that contract forbids it.

## Note for the reviewer

The story xBRIEF is deliberately **not** in this change set. `verify:scope-provenance` fails closed on an active xBRIEF that declares a `file_scope` without an operator-minted approved-scope digest, and no digest exists for `issue-3607-pass-open-advisory-marker`. Minting one is human-presence-gated, so this PR follows the gate's own remediation: keep `active/` out of the product change set and land the completed artifact in a follow-up lifecycle PR after merge (#3476).

Closes #3607
Refs #3613, #3780, #2307, #2814


MScottAdams:
Fix batch pushed (`24519ff7`) for both P1 findings.

**Reused markers skip race arbitration** — valid, fixed. `openPassMarker` now re-lists after PATCHing a reused marker comment and returns `observed` with the actual winner when its own write lost, mirroring the post-create arbitration the lease already does. Two owners reopening the same inactive marker can no longer both report `opened`. Covered by a test that lands a competing older mark between the update and the re-list, plus the re-list error path.

**Marker lifecycle is unreachable** — valid, addressed in the skill rather than a new CLI verb. The review-cycle skill now carries the marker block verbatim and the concrete lifecycle: open with `task scm:body:comment:create`, read with `gh api repos/<owner>/<repo>/issues/<N>/comments` taking the oldest unexpired `kind: pass` block, clear at synthesis with `task scm:body:comment:edit`. That is the same gh-only path the lease already sanctions when review-monitor verbs are unavailable (#2878), so an agent running a pass can publish and clear the advisory state without engine bindings.

SLizard P3 (redundant `listPassMarkerComments` in `openPassMarker`) is acknowledged and not changed: the second call is the arbitration read that the P1 above requires, so removing it would reintroduce the race.

MScottAdams:
Fix batch 2 pushed (`8bdb2bb1`).

**Readback cannot arbitrate overwrites** — valid, and the earlier fix treated the symptom. A post-write readback can never settle two writers on one comment, because each one reads back its own PATCH. The root cause was recycling *any* marker comment, so that is what changed: `openPassMarker` now recycles only a comment the caller authored. Distinct owners therefore always land on distinct comments and resolve deterministically by oldest comment id, which is the path that already deletes the loser's duplicate. The remaining readback has a single writer per comment, and when an older mark from another owner won it now clears its own losing mark rather than leaving two live marks on the thread.

Two runs of the *same* owner can still both write that owner's comment. Both marks then name that owner, so no owner's pass disappears and the observable state is identical; that is called out in the code.

**SLizard P2 (403 editing another user's comment)** — same root cause, closed by the same change. A CONTRIBUTOR can no longer deadlock trying to PATCH a maintainer's comment, which matters here because CONTRIBUTOR authorship is exactly the case this issue exists for.

Tests: 111 in `packages/core/src/review-monitor`, including a foreign-authored expired mark that must be created-around rather than edited, the losing-mark clear, and its failure path. `github-lease.ts` is at 98.5% statements / 90.2% branches; the one uncovered line predates this PR.

MScottAdams:
Fix batch 3 pushed (`5f3f378c`). Both findings are valid and they are the same defect, so this batch removes the shared cell instead of patching another race.

A marker comment was a mutable cell chosen by scanning the thread, and GitHub gives no compare-and-swap, so any second writer silently replaced the first. My previous batch argued that same-login collisions were harmless because both marks name the same owner. That was wrong, and the review is right: a mark also carries pass kind, ceiling and expiry, and on this repo every agent runs under one login, so same-login concurrency is the common case rather than an edge case.

A marker comment now belongs to the one pass that created it:

- **open** never edits a comment it did not create. It creates its own, or reports the mark already on the thread as `observed`. Concurrent creates still resolve oldest-comment-id wins, and the loser deletes its own duplicate.
- **refresh** requires the `commentId` that open returned, so every comment has exactly one writer for its whole life.
- **close** targets that same `commentId` and renders `ended_at` over the marker as freshly read from the thread, so it can neither erase a different pass nor resurrect the metadata its caller opened with.

The post-write readback and its clearing path are gone. They existed to arbitrate a race that no longer exists.

Tests (109 in `packages/core/src/review-monitor`) now pin the two cases the review named as uncovered: a second pass sharing one login is informed and the first mark's `pass_kind` and `ceiling` survive intact; and a close after a renewal writes the renewed metadata plus `ended_at`, never the opener's snapshot. Also pinned: an open never PATCHes an existing comment, whoever authored it. `github-lease.ts` is at 98.2% statements / 88.7% branches, with the single uncovered line predating this PR.

MScottAdams:
## Review-cycle dual stop reached — halting, not merging

Three fix batches have been spent on one failure class (marker write safety), which is the `#2442` cap in `deft-directive-review-cycle` § Dual stop. I am stopping rather than opening a fourth.

**Batches spent**

1. `ef526327` — post-write arbitration on the reused comment, plus the concrete marker lifecycle in the skill. Cleared the "Marker lifecycle is unreachable" P1.
2. `8277c84d` — recycle only a comment the caller authored. Cleared the distinct-owner overwrite and SLizard's 403.
3. `177e6cd9` (was `5f3f378c`) — one comment per pass: an open never edits a comment it did not create, refresh requires the comment id open returned, close renders `ended_at` over the freshly read marker. Cleared both same-login findings.

**Residual, unfixed (class A, already-touched file)**

`Observed markers allow overwrites` (P1, `packages/core/src/review-monitor/github-lease.ts`): the renewal branch trusts `commentId` without verifying the target. A caller that feeds back the id from an `observed` result overwrites another pass's mark and gets `renewed`. SLizard's P2 on `closePassMarker` is the same hole on the close side — its author guard is skipped when `authorLogin` is empty.

Both are one guard: before any PATCH, confirm the target comment exists, parses as a pass mark, was authored by `owner`, and declares `owner`; otherwise return `observed` / `held-by-other`. That is a small, in-AC fix on a file this PR already owns — it needs operator consent to spend a fourth batch, not a new story.

**State**

- HEAD `177e6cd9`, rebased onto current master (the branch was `dirty`; the CHANGELOG `[Unreleased]` collision was union-merged, both entries kept).
- `task check` exit 0 on this HEAD. 109 tests in `packages/core/src/review-monitor`; `github-lease.ts` 98.2% statements / 88.7% branches.
- Greptile last reviewed `5f3f378c` at confidence 3/5; a re-review of `177e6cd9` is in flight.
- Resolved floor is `minGreptileConfidence` = **5** (framework dogfood, `task policy:show --field=plan.policy.review.minGreptileConfidence`). Not lowered, and not to be lowered for this PR.

RESUME: residual=observed-marker-renew-guard leftover=A conf=3/5 floor=5 standing=one-shot PR=https://github.com/deftai/directive/pull/3864

MScottAdams:
Fix batch 4 (operator-authorized residual pass) for the two remaining findings. Both are the same hole on opposite write paths, so they are closed by one guard.

**Observed markers allow overwrites** (Greptile P1) -- valid, fixed. `openPassMarker` trusted the caller's `commentId` and PATCHed it unread, so an id taken from an `observed` result overwrote that pass's kind, ceiling, expiry and owner and still returned `renewed`. The renewal branch now re-reads the thread and writes only when the thread itself proves the target is the caller's own mark.

**Missing authorization check on closePassMarker** (SLizard P2) -- valid, fixed by the same guard. The old close skipped its author check whenever GitHub reported no author, so a marker body naming the caller as `owner` was PATCHed on a comment the caller may not own -- a 403 for a non-maintainer, and the exact case the close path claimed to refuse.

The shared proof, read from the thread and never from the caller's input:

- the comment's GitHub author must be the caller, and an author the API did not report is refused rather than assumed;
- the marker's `owner` must be the caller;
- a refresh must additionally name the same pass -- same `agent_id` and `pass_kind` -- because every agent here runs under one login, so owner alone does not separate two passes. A refresh moves the ceiling and the expiry window only.

A refused renewal never writes. It falls through to the existing read-then-create path, which reports the mark that is already open as `observed` or opens its own comment, so no branch can edit a comment it cannot prove is its own. Close refuses with `held-by-other`.

Tests (`packages/core/src/review-monitor/github-lease.test.ts`), all failing against the previous commit and passing now:

- `refuses to renew a mark it cannot prove is its own` -- four refusal cases: another author, another owner in the body, an author GitHub did not report, and same login with a different pass. Each asserts zero PATCH calls and a byte-identical comment body.
- `opens its own mark when the comment id it was given is no longer on the thread`.
- `refuses to close a mark whose author GitHub did not report` -- SLizard's case, asserting no PATCH.
- Renewal error paths (list failure, PATCH failure) kept.

Residual, stated rather than hidden: two runs that agree on owner login, `agent_id` and `pass_kind` are one pass identity by construction; the thread carries nothing more to tell them apart, and a refresh then moves only that pass's own window. GitHub offers no compare-and-swap on a comment, so that is the floor for an advisory mark.

`task check` is green on this tree.

MScottAdams:
## Batch 4 landed; dual stop reached again at 4/5 -- halting, not merging

Batch 4 was the operator-authorized residual pass. Both named findings are closed and confidence moved **3/5 -> 4/5**, but the resolved floor is **5/5** (`plan.policy.review.minGreptileConfidence`, source: dogfood), so this is not merge-ready and I am not opening a fifth batch (#2442).

**State at halt** (HEAD `5f35249`)

- `task check`: exit 0 on this tree. CI: 22 success / 3 skipped / SLizard neutral. `mergeable_state: clean`.
- `task pr:watch -- 3864 --one-shot`: `NEW_P0_P1`, P0=0, P1=0, sha_match=true, `clean_gate_holdout=has_blocking` -- the holdout is the summary verdict prose plus confidence 4 < 5, not an open inline finding.

**What batch 4 closed**

`Observed markers allow overwrites` (Greptile P1) and `closePassMarker` missing authorization (SLizard P2) -- one guard, described in the previous comment. Four refusal tests plus a close-refusal test, each failing against `e5313695` and passing now.

**New residual, same class (5th variant)**

Greptile 4/5 now names: *synthesis can still erase a concurrent renewal of the same pass marker* -- close and renewal both read the comment, then unconditionally replace its body, so a close that lands after a renewal deactivates the renewed pass and both report success.

That is a read-modify-write race on one comment, and the GitHub comments API has no compare-and-swap. Every batch in this arc has closed the variant named and been handed the next one from the same class. Patching the close path again (adding pass identity to the close guard, as the renewal guard has) narrows the window; it cannot remove it, so I am not spending a batch on it.

**Options for the operator** (choose one; each is a separate decision, not a retry)

1. **Accept 4/5 by exception** and merge with the race documented in code as the floor of an advisory, non-gating mark. Nothing here holds a lock or blocks a write; the worst case is an advisory mark that reads as closed while a pass is still running, and expiry already bounds the opposite error.
2. **Change the shape to append-only**: a close posts a *new* terminating comment instead of editing the mark, and readers take the newest terminal record for a mark. No write ever replaces a body, so the class ends rather than shrinking. This is a design change to #3607, not a fix batch -- it wants its own issue and review.
3. **Drop the close write** and rely on expiry alone (`ended_at` disappears from the write path entirely). Cheapest terminator; costs the fast synthesis signal.

**Lifecycle**

Not merged, not completed. The xBRIEF is deliberately untracked in the worktree (#3874: `verify:scope-provenance` fails closed on an untracked/modified active brief without an operator-minted digest, which no agent can mint); it lands in a lifecycle PR after merge, per PR #3876.

---

**Resume line** -- leftover class: none (both authorized findings closed, CI green, `task check` 0). Residual class: **marker write safety, concurrent close vs renewal on one comment**, Greptile-blocking at 4/5. Resolved floor: **5/5** (dogfood, `minGreptileConfidence`). Authorization: **one-shot** batch 4, now spent -- no standing authorization for batch 5. PR: https://github.com/deftai/directive/pull/3864. Skill: `content/skills/deft-directive-review-cycle/SKILL.md` § Dual stop / operator follow-up (#3273). To resume, pick option 1, 2 or 3 above and say so explicitly; a bare "continue" would re-enter the same loop.

MScottAdams:
## Operator-authorized merge exception -- this PR only

Merging at Greptile confidence **4/5** against a resolved floor of **5** (`plan.policy.review.minGreptileConfidence`, source dogfood). Authorized by the operator on 2026-08-28. Recorded as the human-merge override trail.

**The policy floor is unchanged.** It was not lowered and `pr:merge-ready` still evaluates against 5 -- a one-PR exception, not a gate edit (#3156).

### The residual, named and filed

Synthesis can still erase a concurrent renewal of the same marker: a read-modify-write on one GitHub comment, with no compare-and-swap available in the API. **Filed as #3885** with three candidate directions and its severity stated.

### Why merging beats a fifth batch

- **The branch is strictly better than master.** Batch 4 closed the actual overwrite hole: a write to an existing mark now requires the comment's GitHub author, the marker's `owner`, and -- for a refresh -- the same `agent_id` and `pass_kind`, all read from the thread; an author GitHub did not report is refused rather than assumed. Greptile's `Observed markers allow overwrites` and SLizard's `closePassMarker` P2 are both closed. Confidence moved 3/5 to 4/5.
- **The residual is low severity by design.** Pass marks are advisory observe-and-mark, bound that way by this issue's synthesis. A lost renewal means an arriving agent briefly sees no open pass; the marker never blocks a write and holds no authority.
- **The fix exceeds the bound scope.** This arc bound the work as adoption of the incumbent lease plus two fields and a call site. Append-only semantics or an external lock is a new mechanism, which the synthesis said must beat the incumbent rather than be bolted on.
- **Zero P0 and zero P1**, no open inline threads, CI green, `mergeable_state: clean`.

Leftover class B, parked as its own story. Post-merge lifecycle -- `scope:complete` and the two verify gates -- is owned by the orchestrating session.

Issue #3858, opened 2026-08-28
bug(scm): the enforcing SCM readiness gate never invokes the worker-principal validator


MScottAdams:
# The enforcing SCM readiness gate never invokes the worker-principal validator

Split from the #3693 arc, 2026-08-28. Found by critic A (`5455120168` finding F2) and re-verified by the parent before binding.

## The finding

`validateGithubAuthForWorker` — the function that performs worker-principal verification, raises `FAILURE_INSTALLATION_IDENTITY_UNVERIFIABLE`, and carries the deferral pointers to #3693 — **is not reachable from any gate that authorizes SCM work.**

Measured at `d0c015a6`:

| Anchor | Finding |
|---|---|
| `readiness.ts:219` | `const depth: ScmProbeDepth = options.depth ?? "shallow"` |
| `readiness.ts:269` | the validator call is inside `if (depth === "deep")` |
| `readiness.ts:294` | `validateGithubAuthForWorker(...)` — the only readiness call site |
| `readiness.ts:528` | `requireScmReady` passes `depth: options.depth ?? "shallow"` |
| `scm/main.ts:31` | `requireScmReady({ whichFn: options.whichFn })` — **no depth** |
| `issue-ingest-cli.ts:35` | `requireScmReady()` — **no depth** |
| `reconcile-issues-cli.ts:34` | `requireScmReady()` — **no depth** |

A repository-wide grep for `depth: "deep"` outside tests returns **nothing**. The only production paths that reach deep are `session-start.ts:675`, gated on `--with-network` and wrapped in a try whose comment reads *"Never throws — session:start must not hard-block on SCM absence"*, and `scm:status --deep`, an explicit operator flag. **Neither authorizes anything.**

The only other production caller of the validator is `swarm/launch.ts:289`, which pins the mode to `injected-token` for credential injection — a different surface from SCM verb authorization.

## Why it matters

The identity gate is **diagnostic, not enforcing**. Three consequences:

1. **Decisions made about the validator have no effect on authorization.** #3693 is deciding whether App installation credentials may be accepted as worker principals. Implementing any answer inside `validateGithubAuthForWorker` would not change whether such a credential can drive `scm:*`, because nothing enforcing calls it.

2. **"We fail closed on installation credentials" is true of the function and false of the gate.** The parent asserted the stronger claim in the #3693 triage write-back and it was wrong — traced one level too shallow. Recording "keep failing closed" as a decision, while this gap exists, would retire the question behind a safety property the gate does not have.

[...]

Issue #3469, opened 2026-08-18
Grok Build TUI is detected as routing provider `unknown`, so leaf pins never use the grok rows


MScottAdams:
## Summary

A live Grok Build session (spawn_subagent present, parent Grok 4.6) resolves `task verify:routing` as **provider `unknown`**. Engine host detect requires `GROK_BUILD` or `DEFT_AGENT_RUNTIME=grok-build`. This TUI sets neither. Routing then uses the `unknown` pins (`leaf-implementation` / `review-monitor` → `grok-4.6`) instead of the `grok` / harness-default rows.

The swarm skill still says: Grok Build = `spawn_subagent` present. The engine no longer uses that. Two ledgers.

## Field evidence (2026-08-18, deftai/directive)

Printed in-process:

- `GROK_BUILD=""`
- `DEFT_AGENT_RUNTIME=""`
- `CURSOR_AGENT` / `CURSOR_COMPOSER` / `CI` empty

`task verify:routing --advise` printed: `provider 'unknown': all 1 gated role(s) decided.`

`.deft/routing.local.json` on this machine:

- `unknown.leaf-implementation` and `unknown.review-monitor` pinned `grok-4.6` (2026-08-14)
- `grok-build.leaf-implementation` / `review-monitor` also pinned `grok-4.6` — but live resolve maps that runtime to key **`grok`**, so the `grok-build` key never applies
- `grok.*` is harness-default (correct for a harness-bound provider)

Session-start host-surface was `class=file-first source=assumed` (same "no signal → invent a default" pattern).

## Root cause

`resolveDispatchProvider` (`packages/core/src/swarm/routing.ts`) only returns `grok` when `GROK_BUILD` is truthy or `DEFT_AGENT_RUNTIME=grok-build`. Else `unknown`.

Claude Code sets `CLAUDECODE` in tool/hook subprocesses. Grok Build does not set an analog. OpenClaw has env **and** a `~/.openclaw` filesystem fallback. Grok Build has neither.

`unknown` can still look "decided" if someone pinned it. That hid the miss.

## Other places that expect Grok Build and miss the same way

| Surface | What happens without the env |
| --- | --- |
| `verify:routing` / `swarm:launch` `resolved_model` | provider `unknown`; stamps the unknown pin |
| `probeSubagentBackends` | `grok-build` `available=false` |
[...]

PR #3869, opened 2026-08-28
fix(scm,platform): deny managed cloud by positive metadata read, require explicit host-gh selection when ambiguous

### LINKED ISSUE #3859: bug(scm,platform): CURSOR_AGENT treated as proof of cloud on local desktop, silently skipping every SCM gate
# Local desktop is misclassified as cloud, and no OS predicate can safely fix it alone

**Recut 2026-08-28** after the audit `5455280918` returned STEER. The original direction — `process.platform === "win32"` as a positive discriminator — is **withdrawn**. See the parent correction `5455305166`. The thread preserves the earlier text.

## The defect, unchanged

`intake/platform-capabilities.ts:32-33` and `platform/platform-capabilities.ts:119` treat `CURSOR_AGENT` as proof of a cloud runtime. Local desktop Cursor sets it too. So local desktop classifies `cloud-headless`, infers `injected-token`, finds no `GH_TOKEN`, and returns not-ready — while a healthy keyring `gh` login sits unused.

Every SCM-dependent gate is then skipped: `triage:queue`, `issue:ingest`, `umbrella:current-shape`, the design-critique chip write path. **The skip is advisory**, so the agent is never told its work-selection surface is gone. It proceeds believing it consulted a queue it never reached.

## Why the obvious fixes are both wrong

**Removing the `CURSOR_AGENT` hop** grants `host-gh` on a real cloud VM by marker *absence*. Ruled out.

**Adding `process.platform === "win32"`** was the original direction here and does not survive. Two measured reasons:

1. **Cursor supports Windows "My Machines."** The agent loop runs in Cursor's cloud while terminal commands and file edits execute on a Windows machine. `CURSOR_AGENT` + Windows therefore does **not** entail local desktop.
2. **"Windows implies not managed cloud" is a versioned fleet assumption**, not an invariant. Cursor's docs do say managed agents run on Ubuntu today, but that is a fact about a third party's infrastructure that can change without notice. Being on Windows is a positive OS fact; *therefore not cloud* is an inference from Windows currently being absent from someone else's fleet.

The classifier was collapsing two different questions — *is this a Cursor-managed VM*, and *is host credential use appropriate on this execution boundary*. `process.platform` answers neither.

## Bound contract

**Operator decision, 2026-08-28** (human-origin): take the explicit-selection route. Do **not** couple this issue to #3858.

### 1. A positive cloud deny, at highest priority

Cursor-managed VMs serve a local metadata API on `CURSOR_AGENT_SOCKET` where `agent/runtime` is always `managed`. Self-hosted workers do not serve it. A successful read is a **positive assertion of managed cloud**.

Probe it, and classify `cloud-headless` when it reports `managed`. Give this higher precedence than any other Cursor signal.

**⊗ Socket absence must never select `host-gh`.** Absence means "not a managed VM, or the API was unreachable" — treating it as proof of local recreates the exact marker-absence grant this issue exists to prevent. Absence leaves the runtime *ambiguous*, which item 2 handles.

### 2. Ambiguous runtime requires explicit selection

When `CURSOR_AGENT` is set and the managed probe does **not** report `managed`, the runtime is ambiguous. Do not guess from the OS.

Require an **explicit operator or dispatcher `host-gh` selection** before host credentials are used. There is precedent: `githubAuthMode` is already an explicit field on the swarm launch envelope (`swarm/launch.ts`), consumed alongside `inferGithubAuthMode`. The readiness path needs an equivalent explicit opt-in. Choose the surface deliberately — environment variable, policy field, or CLI flag — and state in the PR body why, along with how an operator discovers it. Absent that selection, behaviour is unchanged from today.

### 3. Tell the operator the gates are off

The current failure is silent, which is what makes it costly. When SCM-dependent gates are skipped, that must be **visible**, not advisory-only. At minimum the reason must appear in `scm:status --json`; if there is a session-start surface that can carry it, say so.

### 4. Preserve CI precedence

The CI loop at `intake/platform-capabilities.ts:24-28` (`GITHUB_ACTIONS`, `CI`, `BUILDKITE`, `DEFT_AGENT_RUNTIME`, `GROK_BUILD`) runs **before** the `CURSOR_AGENT` hop at `:32`. Nothing you add may disturb that ordering. Verify it yourself against current HEAD.

## Diagnostics — required, but not the safety case

Keep all three: a reason string in `scmReadinessToDict`, a local-desktop-authenticated regression, and a CI-ordering regression.

**They must not be presented as the deterministic control.** The audit established why: on the day a Windows managed VM appears, the CI-ordering test does not fire because a managed VM need not set `CI`; the local-desktop test keeps passing; and the reason string keeps truthfully reporting which branch was taken while saying nothing about whether that branch's premise still holds. A test asserting `win32 => local` *freezes* the assumption rather than falsifying it. The safety case is items 1 and 2; these are observability.

## Acceptance

1. A managed-runtime probe classifies a Cursor-managed VM as `cloud-headless` on a positive read, at higher precedence than other Cursor signals.
2. Probe absence never selects `host-gh`.
3. On an ambiguous runtime, host credentials require explicit operator or dispatcher selection; without it, behaviour is unchanged.
4. With that selection present on local desktop with a healthy authenticated `gh`, `scm:*` verbs succeed and SCM-dependent gates are not skipped.
5. A Windows GitHub Actions runner still classifies `cloud-headless`.
6. Skipped SCM gates are operator-visible, not advisory-only.
7. No path selects host credentials from marker absence or from an OS predicate alone.
8. CHANGELOG `[Unreleased]` line.

## Out of scope

- **⊗ The cloud half** — whether a file-backed `ghs_` App installation login may be authorized. #3656, blocked on #3693.
- **⊗ Principal validation on the enforcing path.** That is #3858, deliberately excluded by the operator to keep this shippable. Do not pull it in.
- **⊗ The `GROK_BUILD` contradiction** between the two classifiers (#3469).
- **⊗ The `isHeadlessSession()` TTY bug.**
- **⊗ Generic container detection** as a cloud signal — it misclassifies a developer running Linux locally in Docker. Filesystem and cgroup fingerprints (`/opt/cursor/cloud-agent-tools/`, `/pod-daemon`) may serve as corroborating diagnostics, never as the authorization invariant.

## On the string `"win32"`

If any platform check survives as a diagnostic: `process.platform` returns `"win32"` on **all** Windows including x64 and ARM64. It is Node's OS-family identifier, not a bitness claim, and there is no `"win64"`. Measured at `5455237296`. Do not narrow it with `process.arch`.

## Provenance

Direction originated in the #3693 arc (`5455120168`, synthesis `5455183890`) and was **materially corrected** by the targeted audit `5455280918`, which reached first-party Cursor documentation the earlier panel did not. The audit investigated `CURSOR_AGENT_SOCKET` despite this issue's own prose prohibiting it, and was right to — that prohibition was mine, inherited unchecked, and is withdrawn.

Refs #3469, #3656, #3693, #3858.

MScottAdams:
## What was broken

`CURSOR_AGENT` was treated as proof of a cloud runtime. It is not — local desktop Cursor sets it too. So a local desktop session classified `cloud-headless`, inferred `injected-token`, found no `GH_TOKEN`, and returned not-ready while a healthy keyring `gh` login sat unused.

Every SCM-dependent gate was then skipped — `triage:queue`, `issue:ingest`, `umbrella:current-shape`, `pr:*`, `cache:fetch-all` — and **the skip was advisory**, so the agent was never told its work-selection surface was gone. It proceeded believing it had consulted a queue it never reached.

Reproduced live on this machine before the fix, and again after it.

## Why not the obvious fixes

**Removing the `CURSOR_AGENT` hop** grants `host-gh` on a real managed VM by marker *absence*. Ruled out.

**`process.platform === "win32"`** was the originally bound direction and is **withdrawn**. Two measured reasons:

1. Cursor supports Windows **"My Machines"** — the agent loop runs in Cursor's cloud while terminal commands execute on a Windows machine. `CURSOR_AGENT` + Windows does not entail local desktop.
2. "Managed VMs are Ubuntu" is a **versioned fact about a third party's fleet**, not a runtime invariant. The day that changes, a win32 carve-out silently grants `host-gh` on a managed VM.

The classifier was collapsing two questions — *is this a Cursor-managed VM* and *is host credential use appropriate on this execution boundary*. `process.platform` answers neither, and is not an input to either classifier in this PR.

## The design

**1. A positive cloud deny, at the highest Cursor precedence.** Cursor-managed VMs serve a metadata API on `CURSOR_AGENT_SOCKET` whose `agent/runtime` is always `managed`; self-hosted workers do not serve it. A successful read is a *positive* assertion of managed cloud, so it classifies `cloud-headless` above every other Cursor signal — including above the sandbox hop, which previously inferred `host-gh`, and above the operator opt-in below.

**2. Absence is ambiguous, never local.** Socket absence, an unreachable socket, a non-200, or a body that does not assert `managed` all leave the runtime *ambiguous*. No path selects host credentials from marker absence. Both probe error directions are safe: a false `managed` denies host credentials, and a failed read preserves today's behaviour.

**3. An ambiguous runtime requires explicit selection.** When `CURSOR_AGENT` is set and the probe does not report `managed`, host credentials require `DEFT_GITHUB_AUTH_MODE=host-gh`. Absent it, behaviour is unchanged from today.

**4. Both classifiers share one precedence order:** CI markers → positive managed read → Cursor sandbox → ambiguous Cursor hop. This moves `platform/platform-capabilities.ts`'s Cursor hop *below* its CI hops, so neither a CI marker nor a managed read can be overridden by the opt-in. The CI loop in `intake/platform-capabilities.ts` still runs first — verified against HEAD, not assumed.

### Why an environment variable, and how an operator finds it

The ambiguity is a property of the **execution environment**, not of the repository. A policy field in `plan.policy` would be committed and would then propagate to every other machine and to managed VMs — exactly the wrong scope for "this box's credential store is the right one." A CLI flag would only reach `scm:status`, not the other SCM-dependent verbs. An env var matches the lifetime and blast radius of the decision.

It is named `DEFT_GITHUB_AUTH_MODE` after the existing `github_auth_mode` dispatch-envelope field (`swarm/launch.ts`, consumed near `inferGithubAuthMode`), so a dispatcher exports the same label it already records.

Discovery is at the point of failure rather than in a doc an operator would have to know to read:

- `scm:status --json` reports `runtime_mode_reason` (`cursor-marker-runtime-ambiguous`).
- The `[deft scm]` skipped-gates line carries that reason, and session-start shows it.
- The remediation an operator sees when blocked names the exact variable and value, and states that a positive managed read always wins.
- `content/scm/github.md` § Ambiguous Cursor runtime and the host-gh opt-in.

Only the exact value `host-gh` is honoured; `hostgh` / `host_gh` / `1` / `yes` are not treated as a selection.

## On the diagnostics

The reason string and both regressions are here because the issue requires them, but they are **observability, not the safety case**. A test asserting a classification *freezes* an assumption rather than detecting its invalidation. The safety case is the precedence contract: a positive managed read denies, absence never grants, and ambiguity requires an explicit opt-in.

The default probe is verified against a real metadata server on a local socket in its own process, covering `managed`, not-managed, non-200, and unreachable — not only stubbed verdicts.

## Verification

- Live repro on this Windows Cursor session: ambiguous → not-ready with the reason and opt-in named; with `DEFT_GITHUB_AUTH_MODE=host-gh` → ready, `host-gh`, authenticated, zero skipped gates; with `GITHUB_ACTIONS=true` added → `ci-marker`, still denied.
- Full suite green; targeted coverage on the changed modules is 82.75% branch against a 75 floor.
- `task check` green.

### Classification matrix

| Environment | Before | After (no selection) | After (`DEFT_GITHUB_AUTH_MODE=host-gh`) |
| --- | --- | --- | --- |
| Local desktop, Cursor | `cloud-headless`, gates skipped silently | `cloud-headless`, gates skipped **with reason + opt-in named** | `local-unsandboxed` → `host-gh`, ready |
| Cursor-managed VM (metadata reports `managed`) | `cloud-headless` | `cloud-headless` (`cursor-managed-runtime-probe`) | `cloud-headless` — opt-in **cannot** override |
| Windows My Machines worker | `cloud-headless` | `cloud-headless` (ambiguous) | `local-unsandboxed` → `host-gh` |
| Windows GitHub Actions runner | `cloud-headless` | `cloud-headless` (`ci-marker`) | `cloud-headless` — CI keeps precedence |

**Not verified against real infrastructure:** I have no Cursor-managed VM or My Machines worker to run against, so those two rows are verified by the probe contract and a live local socket server, not by a managed fleet. The managed row depends on Cursor's documented `agent/runtime = managed`; the exact response shape is documented only as that value, so the body parser accepts a bare string and a JSON object with a `runtime` field, and over-matching there fails closed.

## Review round 2 -- the inconclusive-probe cache

Greptile flagged the memo as the one place this change could still fail open, and it was right. The first cut memoized every probe verdict per socket path. On a managed VM that lost a startup race with the metadata API -- socket not yet listening, or a transient non-200 -- the first read recorded `unavailable`, that verdict stuck for the life of the process, and an explicit `DEFT_GITHUB_AUTH_MODE=host-gh` would then be honoured on a machine the probe would have denied a second later. That is precisely the fail-open direction this PR exists to close.

The memo is now asymmetric:

- **A `managed` read is memoized.** A VM cannot stop being managed mid-process, so caching the deny is sound, and it keeps the memo doing what it was for -- one child spawn per socket on the machines where the probe actually fires.
- **Every other verdict re-probes.** `unavailable` and `not-managed` only say the API did not assert `managed` on that attempt. Re-probing costs one spawn bounded by the existing timeouts; caching one is unbounded in the fail-open direction.

The missing coverage is `re-probes an unavailable read and denies once the socket reports managed`: it reads through a socket path with no listener, gets `unavailable`, then starts a real metadata server on that same path and asserts the second read is `managed` and that the classification is `cloud-headless`. **It fails against the pre-fix memo** -- the second read returns the cached `unavailable` -- and passes after. Two narrower tests pin each half of the asymmetry on its own.

One more finding from this round: the probe's contract is that it never throws, but `JSON.parse` returning `null` or a primitive would have reached a property access on that result. Body parsing is now a named `parseProbeResponse` helper that returns `undefined` for anything that is not a JSON object, so the defensive path is directly testable rather than implicit.

Rationale recorded in `xbrief/decisions/2026-08-28-memoize-only-a-positive-managed-runtime-read-re-probe-every-inco.decision.json`. No gate, verifier, or coverage floor was touched.

## Out of scope

The cloud half (#3656), principal validation on the enforcing path (#3858), the `GROK_BUILD` classifier contradiction (#3469), the `isHeadlessSession()` TTY bug, and generic container detection — filesystem and cgroup fingerprints are corroborating diagnostics at best, never the authorization invariant.

Closes #3859
Refs #3469, #3656, #3693, #3858


MScottAdams:
ended_at: 2026-08-28T22:50:03.483Z

Issue #3433, opened 2026-08-17
feat(session,swarm): fail-closed worktree occupancy lease for overlapping mutation sessions


MScottAdams:
## Summary

A swarm parent (or any mutation session) on a shared checkout and a second session that starts mutation `session:start` can both believe they own the same worktree. Existing file locks and freshness binds do not prevent that.

Add a **worktree occupancy lease**. Mutation ceremony fail-closes when another live occupant exists, unless the newcomer presents an explicit grant. Coordinated overlap is allowed only when the environment can actually reach the occupant.

This issue is the first ship: claim / deny / steal / TTL. A join-request protocol is the follow-up, not this PR.

## Problem

Swarm isolates **workers** (one worktree per leaf, file-overlap audit). It does not isolate the **parent checkout**.

The swarm parent usually sits on the primary tree and keeps mutating shared local state: ritual, freshness bind, `xbrief/` lifecycle, cache, CHANGELOG, worktree add/remove, merge close-out.

A second interactive session that runs mutation `session:start` on that same tree is treated as the new owner:

- Cold `session:start` always mints a new `session_id` and overwrites `.deft/ritual-state.json`. There is no occupant check.
- Re-arm can inherit the live parent's ritual when worktree + HEAD look continuous, so the newcomer never notices the parent is still live.

On one worktree these cannot be shared even if the agents "agree" in chat:

- git HEAD and the index
- the dirty tree
- `.deft/ritual-state.json` (single slot)

Resulting failures: stolen ritual / hook deny-or-wrong-allow, branch or index races during worktree add or cohort close-out, `xbrief/` lifecycle races, CHANGELOG races, `scope:complete` against the other session's leftovers.

The same class applies to **any** two mutation sessions, not only swarm-vs-implement. Swarm is the loud case.

## Why existing protections are not enough

| Existing | What it serializes | Gap |
|---|---|---|
| Swarm worktrees + file-overlap audit | Parallel leaf writers | Not a second parent on the same tree |
| #1260 PROJECT-DEFINITION lock | Short RMW of one JSON file | Not session ownership |
| #1329 `agents:refresh` lock | Managed AGENTS block RMW | Same |
| #3117 freshness bind + `DEFT_SESSION_ID` | Stops one session certifying another as current | Does not stop the second session from writing |
| #2814 review-owner lease | One babysitter per PR (GitHub sticky comment) | Different resource; right lease *shape* |
[...]

Issue #2113, opened 2026-06-30
Cursor harness: session-start ritual + branch gates silently dropped on summary-based resume


nheroux-bit:
## Environment

- **Harness: Cursor IDE agent (Cursor harness).** This was observed specifically in Cursor, where long sessions are continued via **context summarization/compaction** rather than a verbatim transcript.
- Deft Directive engine: `@deftai/directive-core@0.56.2`, vendored `.deft/core`.
- Consumer project with `vbrief/` lifecycle, `allowDirectCommitsToMaster` default (OFF).

## Summary

When a Cursor session is **resumed from an auto-generated summary** (the model's prior context is compacted into a summary blob and the live transcript is discarded), the agent does **not** re-run the Deft session-start ritual or the pre-implementation gate stack before resuming work — including creating/activating vBRIEFs and writing code. AGENTS.md is still injected as a rule, but the *ritual* (which is an action triggered "on every interactive session start") is not re-triggered, because the harness treats a resumed session as mid-session, not session-start.

## What happened (concrete)

After a summary-based resume, the agent:

- Did **not** read USER.md content or emit the Deft alignment confirmation echoing the addressing-name (the "unfakeable read" check).
- Did **not** run `deft session:start` / surface `deft verify:tools` / `deft triage:welcome` / cache-fresh warnings.
- Did **not** emit the branch-policy disclosure.
- Did **not** run the Story Start Gate (`git status --short --branch`) or the pre-`start_agent`/pre-code gate stack (`verify:session-ritual`, `verify:story-ready`, `vbrief:preflight`, `verify:branch`) before writing code.
- Created and activated a new scope vBRIEF and wrote/edited multiple source files directly.
- Continued **Scope 9 and Scope 10** work on a **Scope 8 feature branch** (`feat/scope8-banker-formatting`), violating one-story-per-branch hygiene. (The branch was at least not the default branch, so no direct-to-main violation — but nothing *checked*.)

The user had to notice and ask "did you read AGENTS.md / is deft installed / why are we building on this branch" — the gates that should have caught all of this never fired.

## Why this happens (hypothesis)

- The session-start ritual (#1149) and `verify:session-ritual` (#1348) are framed around an **interactive session start**. A summary-based **resume** is not detected as a session start, so neither the ritual nor the staleness check is re-triggered.
[...]

PR #3870, opened 2026-08-28
Decide occupancy before any ritual persist, and keep the compact stale-mark fail-open (#3769)

### LINKED ISSUE #3769: bug(session): an excluded session can rewrite the lease owner's ritual state and lock it out of its own worktree
## Summary

A session the occupancy lease is supposed to exclude can still write the lease holder's `.deft/ritual-state.json`, and can lock the holder out of its own worktree. Two distinct paths, both measured.

This is a **cross-session denial-of-write against the lease owner**. It is not caused by the spawn exemption and cannot be closed by fixing that; split out of #3755 on its critic's disposition.

## Path 1 -- a denied write has already mutated the owner's state

The gated ritual verifier runs at `dispatcher.ts:963-976`, **upstream of** the occupancy gate at `:996`.

`agent_hooks` is exempt from the pass-cache and re-executes on every mutation boundary (`verify-session-ritual.ts:685-697`), and `runGatedStep` writes the state file (`verify-session-ritual.ts:266-271`).

**Measured:** a foreign session's direct write, ultimately denied `occupancy-occupied`, had already executed the gated step and rewritten the owner's `.deft/ritual-state.json` -- replacing `gated_steps.agent_hooks` with the foreign process's result, while `session_id` stayed as the owner's.

Denial does not prevent the write, because the write is ordered before the denial.

## Path 2 -- a foreign compact locks the owner out

`session.compact` routes to `markRitualStaleAfterCompact` with **no identity and no occupancy check** (`dispatcher.ts:1557-1559`). It never enters `inspectMutationGates`, so occupancy cannot see it by construction.

**Measured:** a foreign session's compact event set `rearm_needed: true` on the owner's record. The owner's next direct write was then denied `ritual-not-ready` -- "Compact/resume marked re-arm needed."

So an excluded session can revoke the holder's write authority in its own leased tree.

## Why the lease cannot help

Both paths are outside what the lease covers. Path 1 is an ordering problem: the mutation happens before the gate that would have rejected it. Path 2 is an entire event class the gate never sees.

Worth noting the asymmetry in locking, too: `ritual-sentinel.ts:160-191` writes ritual state with temp-plus-rename and **no cross-process lock**, while occupancy takes an append lock (`occupancy.ts:604-619`). Lost updates between concurrent same-tree actors are structurally available. **Unverified** -- no stress test was run.

## Why this is reachable in normal operation

Same-worktree dispatch is not exotic, it is mandated. `core-phase-5-6.md:26` requires a Tier-1 monitor to background-dispatch a continuation leaf in the **same worktree**, repeated at `core-phase-0.md:36` and `review-cycle/SKILL.md:514`. Those leaves write. So multiple sessions operating in one leased tree is a supported, required shape -- not a misuse.

## Acceptance

- [ ] A session that fails the occupancy gate does not mutate the owner's `.deft/ritual-state.json` -- order the gate before the gated-step write, or scope the write to the resolved actor
- [ ] `session.compact` cannot mark another session's record `rearm_needed` -- it needs an identity check, or must route through the mutation gates
- [ ] Regression tests for both: a denied foreign write leaves the owner's ritual state byte-identical; a foreign compact does not affect the owner's `rearm_needed`
- [ ] A decision on whether ritual-state writes need the same cross-process lock occupancy already uses

## Out of scope

- The spawn exemption itself (#3755). Neither path here involves a spawn tool.
- Whether same-worktree dispatch should be gated at all (#3755). This defect exists either way.
- Occupancy admission semantics and the cooperative-versus-authorization framing (#3611, closed).

## Provenance

Both measurements are from the #3755 round-1 critic (comment 5427665352), run via a `tsx` harness against source at `0a82ac9665880210b4b7da3c0e841733ad927ea5`, using the same seam shapes as `dispatcher.test.ts`, with occupancy claimed only in throwaway temp roots. No repository state was written. **The parent has not independently re-run them** -- re-derive before building.

Line numbers in this issue have already drifted three times during this cluster as merges landed. Cite behaviour and deny codes rather than lines.

Refs #3755, #3433, #3611, #3613.

MScottAdams:
## What

Two paths let a session that does not own a worktree lease rewrite the ritual state of the session that does.

**Path 1 - occupancy decides before any ritual persist.** At 52a5fddc `inspectMutationGates` called `verifySessionRitual` at :1110 and `evaluateOccupancyWriteGate` at :1143, so a ritual persist happened before the occupancy verdict. That verifier executes `agent_hooks` and rebinds a forward HEAD, and both write `.deft/ritual-state.json`. A foreign session the gate was about to deny had therefore already rewritten the occupant record by the time the deny returned. Occupancy now runs first, and the deny paths read the ritual through `inspectSessionRitual`, which runs no gated step and never rebinds. Only a session that occupancy admits reaches the verifier, and no `session_id` retitle happens on a denied path.

**Path 2 - compact stays always-allow and fail-open.** `session.compact` keeps its own branch in `decideHook` and never routes through `inspectMutationGates`. The stale-mark writes even when the acting identity cannot be bound. Per operator amend 2026-08-28 that is deliberate: compact has no issue context and runs inside a hook budget, so a skipped write has no surface a human reliably sees. The accepted cost, stated plainly: an unidentified compact re-arms the owner. Both call sites now name the decision, and a test pins the direction.

## Shape

The bound contract (comment 5455499276) dropped two limbs that were filed as OR-pairs. Neither is implemented here:

- routing compact through `inspectMutationGates`, which would re-run the Path 1 writes or occupancy-deny an owner compact
- scoping the ritual write to the resolved actor, which is a `session_id` retitle on a denied path

Ritual-state locking stays a recorded decision item and is out of scope.

## Tests

The three Path 1 assertions were run against the pre-fix dispatcher and fail there:

- `occupancy.test.ts`: a denied foreign write leaves the owner ritual state byte-identical on a fresh record. The ritual verifier is deliberately not stubbed in that suite (only the gated entrypoint runner and the scope reader are), so a persist would be visible.
- `dispatcher.test.ts`: the persisting verifier seam is never called on an `occupancy-occupied` deny or an `occupancy-identity-unavailable` deny. The no-write inspect answers instead.
- `compact-ritual.test.ts`: an unbindable compact still sets `rearm_needed`, does not retitle the record, does not disturb the lease, and never consults the mutation verifier.

A fourth test pins the premise the ordering rests on: the write-gate verifier does persist when it runs.

## Note

`InspectSessionRitualOptions` gains `requiredGatedSteps` so the no-write inspect reports the same readiness its verifying counterpart would (`WRITE_GATED_REQUIRED_STEPS`, #3738). Without it the deny-path note would report `ritual-not-ready` for a red `cache_fresh` that the write gate deliberately treats as optional.

Closes #3769. Refs #3613, #3433, #3611, #2113, #3738.
## Checklist

- [x] `/deft:change` - N/A: seven files, all tracing to the single scope xBRIEF for #3769; no cross-cutting or architectural expansion.
- [x] `CHANGELOG.md` - entry added under `[Unreleased]` / Fixed.
- [x] `ROADMAP.md` - N/A: not a roadmap-tracked item.
- [x] Tests pass locally - `task check` exit 0 on aa0dedf6.

## Post-Merge

- [ ] Verify issue #3769 auto-closed after squash merge; close manually if the keyword did not fire (#167).


MScottAdams:
## Review residual: ritual-state locking is the remaining limb

HEAD `ffcb6b70bed2e665bd552774024983efa6a5dd32`. Greptile confidence 4/5, P0=0, two inline P1 threads. Resolved floor for this repo is **5** (`plan.policy.review.minGreptileConfidence`, dogfood). CI merge gate green; SLizard neutral (advisory only).

Both P1 threads name one defect class, and the second is the honest residual of the first fix.

**1. Ritual persists after lease transfer (`dispatcher.ts`) - addressed.** Both persisting writes in the verifier (the gated-step record and the forward-HEAD rebind) now compare-and-swap against the record on disk, so a session that lost the lease mid-verification is refused rather than overwriting the new occupant. Regression test `refuses an in-flight ritual write once the tree is re-armed under a new owner` in `occupancy.test.ts`, verified red against the pre-fix helper.

**2. Ownership check remains non-atomic (`verify-session-ritual.ts`) - not addressed, deliberately.** The interval between the compare read and the write cannot be closed by any read-modify-write over a plain JSON file. Closing it means serializing ritual writes behind a lock. That is the ritual-state locking item, which the bound contract for this issue records as a decision item and places out of scope (issue comment 5455499276, item 5). Per the review-cycle out-of-AC gate (#2881) a new cross-cutting reliability contract needs its own story rather than a mid-review redesign.

Net effect on the named hazard: the window in which a session the gate will deny can rewrite the occupant record shrinks from the whole gated verification (which executes `agent_hooks` and can rebind HEAD, so seconds of wall clock and at least one subprocess) to the interval between a read and a write inside one function. The entry-deny path, which is the defect this issue was filed for, no longer persists at all.

Not merging: the resolved floor is 5 and confidence is 4, and `pr:merge-ready` also reports the #3225 advisory should-not-merge. Handing back for an operator decision: accept this residual on the record, or authorize the ritual-state locking story and re-enter.

RESUME: residual=greptile_p0_p1 leftover=B conf=4/5 floor=5 standing=one-shot PR=https://github.com/deftai/directive/pull/3870

MScottAdams:
## Operator-authorized merge exception -- this PR only

Merging at Greptile confidence **4/5** against a resolved floor of **5** (`plan.policy.review.minGreptileConfidence`, source dogfood). Authorized by the operator on 2026-08-28. Recorded here as the human-merge override trail.

**The policy floor is unchanged.** It was not lowered, and `pr:merge-ready` still evaluates against 5 -- this is a one-PR exception, not a gate edit (#3156).

### The residual, named

The surviving P1 is "Ownership check remains non-atomic" in `packages/core/src/session/verify-session-ritual.ts`. It is correct, and it cannot be closed by a read-modify-write over a JSON file: the gap between the compare and the write is structural, and removing it requires serializing ritual writes behind a lock.

That work is **explicitly out of scope** for #3769 -- its bound contract records "Ritual lock remains a decision item, not scope" as item 5 (comment 5455499276) -- and it is a cross-cutting reliability contract of the #2881 class.

**Filed as #3872**, with the lock-scope, contention, crash-recovery and blast-radius questions the fix has to answer.

### Why merging is the better call than another fix batch

- The branch is **strictly better than master** on this axis: it added compare-and-swap on both persisting writes, so a session that loses the lease mid-verification is now refused instead of clobbering the new occupant. The residual is a narrower window than what exists today.
- The three brief acceptance criteria all shipped, with mutation-verified tests (the new Path 1 assertions fail against the pre-fix dispatcher, the CAS test fails against the unguarded helper).
- `task check` exits 0; CI is green; `mergeable_state` is clean.
- A fourth fix batch could only either expand into #3872's scope or perform cosmetic churn. The worker stopped at one batch of three rather than thrash, which is the correct dual-stop behaviour.

Leftover class B (new subsystem, parked as its own story) per the review-cycle taxonomy. Post-merge lifecycle -- `scope:complete` and the two verify gates -- is owned by the orchestrating session, not by the exited worker.

Issue #3888, opened 2026-08-28
bug(hooks): a repo-local deft-hook runner is not portably registrable from one tracked hooks.json (#3785 items 1+2)


MScottAdams:
## Split from #3785

#3785 bound a four-part remedy. Two parts landed (items 3 and 4, plus an always-on AGENTS.md card); items 1 and 2 did not, and this issue carries them with the evidence for why they need their own change.

**Landed on #3785**
- Item 3: `content/docs/hook-runtime-unavailable.md` documents the out-of-band recovery `deft policy:disable-host-hooks --host cursor --confirm` and forbids the `failClosed` hand-edit that the next `deft update` reverts.
- Item 4: `deft init` / `deft update` warn when an agent-hook registration is git-tracked while no committed `package.json` dependency on `@deftai/directive` lets a clone obtain `deft-hook`.
- Legibility card in AGENTS.md (propagated via `content/templates/agents-entry.md`), because a session denied on exit 127 can still read files.

**Still open (this issue)**
1. Publish a **no-Node `deft-hook` artifact**, and/or deposit a **repo-local runner**, so the command is hostable from a clone.
2. Deposit a tracked wrapper so a missing runtime renders an **explicit deny with a decision code** instead of an opaque exit 127. Still fail-closed, but legible. This also covers the #3736 timeout case.

## Why they did not land together

Both items reduce to the same question: what command string can a single **tracked** `.cursor/hooks.json` name so that it is executable on every host that receives the clone? The deposit writes one `command` string per entry, and that string is shell-evaluated by the host.

| Claim | Verdict | Method |
|---|---|---|
| Cursor shell-evaluates the hook `command` string | verified | Cursor hooks documentation; corroborated by a Windows report where the command is wrapped in a PowerShell base64-decode invocation |
| On Windows the evaluating shell may be PowerShell 5.1 | verified | Windows default; PowerShell 7 is not guaranteed present |
| `\|\|` fallback chaining is a parse error in PowerShell 5.1 | verified | operator introduced in PowerShell 7 |
| A parse error exits non-zero, which under `failClosed: true` is a deny | verified | exit contract: exit 0 renders a decision, every non-zero exit is a non-decision |
| A relative path with forward slashes fails as the **command token** under `cmd.exe`, but is fine as an **argument** | verified | prior Directive deposit shipped `node <relative/forward/slash/path>` successfully |
| A shebang script is not directly executable on Windows | verified | requires `.cmd` / `.bat` or an explicit interpreter |
[...]

Issue #3571, opened 2026-08-20
bug(hooks,policy): disabling hostHooks writes {} and strips guardrails with no capability-cost disclosure — and verify:agent-hooks recommends it


rockbox37:
## Summary

Setting `plan.policy.hostHooks.<host> = false` and running `deft update` replaces that host's hook config with a bare `{}`, removing every `deft-hook` pre-execution guardrail for anyone who later opens the repo in that host. This happens with **no capability-cost disclosure and no confirmation prompt**, and the resulting state is committed to tracked files, so it propagates to every clone.

Two things make this more than a documentation nit:

1. **Directive's own remediation text recommends it.** The `verify:agent-hooks` failure message says:

   > If an affected host is unused, inspect `deft policy:show --field=hostHooks`, set `plan.policy.hostHooks.<host> = false`, and run `deft update`.

   An agent or operator following that advice to clear a *timing* failure (see #3570) silently disables safety gates as a side effect.

2. **Directive's own reviewer bot then flags the result as a security regression.** On the PR carrying this change, `deft-slizard` raised four separate P2 findings, e.g.:

   > `.cursor/hooks.json` is replaced with an empty object (Security Guardrail Bypass) … fail-closed safety gates are completely removed, allowing unchecked execution of destructive or sensitive operations without invoking deft-hook guardrails.

   …plus the equivalent for `.codex/hooks.json`, `.grok/hooks/deft.json`, and the `hostHooks` policy block itself. So the framework recommends an action, performs it silently, and then its own review surface reports that action as a guardrail bypass.

## Inconsistency with comparable policy changes

Other capability-reducing or capability-expanding policy flips are gated behind an explicit disclosure and `--confirm`:

- `deft policy:allow-direct-commits -- --confirm` (branch protection OFF)
- `deft policy:enable-value-feedback -- --confirm` — prints `⚠ Capability-cost disclosure …`
- `task product-signal:enable -- --confirm`

Disabling host hooks removes a *security* control and has no equivalent surface at all: no typed verb, no disclosure, no confirm. The only documented route is hand-editing `PROJECT-DEFINITION` and running `deft update`.

## Repro

1. In a consumer repo, set `plan.policy.hostHooks` to `{"claude": true, "cursor": false, "grok": false, "codex": false}`.
2. `deft update`
3. `cat .cursor/hooks.json .codex/hooks.json .grok/hooks/deft.json` → each is exactly `{}`.
[...]

Issue #2752, opened 2026-07-22
deft update re-projects .claude/settings.json with no persistent opt-out; fail-closed hook causes friction in non-ritual repos


rockbox37:
## Summary

`deft update` (and `directive init`) re-project a Directive-owned `.claude/settings.json` on every upgrade (per the "Agent-host direct-write hooks" feature, #2438 / #2596). For a consumer repo that deliberately does **not** want the fail-closed Claude Code hook, there is no persistent opt-out: removing the file is treated as drift and silently re-created by the next `deft update`. Combined with how the Claude Code `PreToolUse` gate behaves in practice, this produced real friction for a repo that uses Claude Code as a dev/agent host but not the interactive session-ritual workflow.

Reporting from a consumer deposit on **`@deftai/directive` 0.80.0** (dogfooding a separate project).

## What happened

Upgrading a consumer repo's vendored `.deft/core` from 0.75.0 → 0.80.0 projected `.claude/settings.json` (plus `.codex`/`.cursor`/`.grok` equivalents). The Claude `PreToolUse` gate then:

1. **Blocked every `Edit`/`Write`** until a gated `deft session:start` + `deft doctor` ritual was recorded — expected per design, but this repo's workflow doesn't run that ritual, so it was pure friction.
2. **Gated edits to files _outside_ the repo too.** Claude Code's `PreToolUse` matcher keys on the *tool name* (`Edit|Write|…`), not the target path, so once the hook is loaded for the session it fail-closes on edits to any path — e.g. it blocked edits to files in a different directory entirely (agent memory files under `~/.claude/...`).
3. **Went stale on every `git` HEAD change.** The compact/HEAD re-arm (#2113) marks the ritual stale; a routine merge+pull mid-session re-blocked all writes until `deft session:start` was re-run. In a repo with a normal PR cadence this fires constantly.
4. **Survived deletion.** After removing `.claude/settings.json` on purpose, the docs' own guidance ("Repair missing/drifted entries: `deft update`") means the next upgrade re-adds it — the removal isn't honored as intent.

Net effect: to do ordinary work in the repo after the upgrade, every session needs `deft session:start && deft doctor` before the first edit, re-run after each HEAD change — and a deliberate removal doesn't stick.

## Why this is a problem for this class of repo

[...]

PR #3890, opened 2026-08-28
fix(hooks): make the hook-runtime lockout legible and recoverable (#3785)

MScottAdams:
## What this fixes

A Cursor cloud-agent VM (and every CI runner, container, or fresh clone without a global npm install) fail-closes **every** mutation because `deft-hook` is not on PATH. The root cause is not hook semantics: Directive **git-tracks the fail-closed registration while gitignoring the runtime it names**. The registration travels; the implementation does not. Because the binary never runs, the consumer sees an opaque exit 127 and no Directive code is left to explain it.

`failClosed: true` is unchanged. Absent, crashed, and timed-out stay one non-decision class.

## What landed

**Item 4 - deposit-time warning.** `deft init` / `deft update` now report when an enabled agent-hook registration is git-tracked while no committed `package.json` dependency on `@deftai/directive` lets a clone obtain `deft-hook`. The committed pin (#2264) is the existing reconstitution anchor: with it, `npm install` produces the runtime the registration names. Warn-only by design - refusing init/update on this condition would be the same lockout from the other side. Disabled hosts are excluded, since their registration is stripped and cannot deny anything.

**Item 3 - recovery documentation.** New `content/docs/hook-runtime-unavailable.md` leads with the durable, out-of-band recovery `deft policy:disable-host-hooks --host cursor --confirm`, states the capability cost, and explains why `.deft-directive-disable` (read *inside* the missing binary) and `.no-deft-directive` (consulted at `session.start` only) cannot reach a `preToolUse` denial. It documents absent / crashed / timed-out as one non-decision class - which is also the answer left open on #3736.

**Legibility card in AGENTS.md.** A session already denied on exit 127 can still read files, so the tracked AGENTS.md is the only channel that still reaches it. The card names the exit-127 signature, the recovery verb, and the doc. Propagated through `content/templates/agents-entry.md` with a marker test per #1309.

**Upgrade trap, called out everywhere.** The `failClosed: false` hand-edit works once; the next `deft update` rewrites the deposit and silently re-arms the lockout. The doc, the AGENTS card, the deposit warning, and the CHANGELOG all say so, and a test asserts the warning copy carries it.

**CHANGELOG.** The `[Unreleased]` entry names `deft policy:disable-host-hooks --host cursor --confirm` as the durable workaround, explicitly not the hand-edit.

## What did not land, and why

**Items 1 and 2 are split to #3888** with the portability evidence. Both reduce to one question: what command string can a single **tracked** `hooks.json` name so that it is executable on every host receiving the clone? Findings:

- Cursor shell-evaluates the hook `command`; on Windows that shell may be PowerShell 5.1.
- `||` fallback chaining is a **parse error** in PowerShell 5.1 (the operator arrived in 7), and a parse error exits non-zero, which under `failClosed: true` is a deny. A fallback command string would fix the cloud-agent case and break every working Windows 5.1 user.
- A relative forward-slash path fails as the command token under `cmd.exe`; a shebang script is not directly executable on Windows; an explicit `node` interpreter does not help because Node is exactly what is missing; an absolute path is not trackable.
- No OS-keyed command variant exists in the documented `hooks.json` schema.

So item 1 needs either a no-Node standalone binary or a deliberate decision to stop tracking the registration (which trades auditability: a tracked file's deletion shows in `git status`, an untracked file's absence is silent). Item 2 inherits the same execution problem - a wrapper that renders the deny must itself be runnable without Node. Neither is a change to make quietly inside a release-gating PR.

**Honest answer to "does a no-Node consumer now get a legible deny instead of exit 127?"** Not in-session: the host still reports exit 127, because no Directive code runs. What changed is that the condition is caught **before it ships** (deposit-time warning), and a locked-out session now finds the recovery in AGENTS.md and the new doc instead of guessing. The in-session legible deny is #3888.

## Gate integrity disclosure (#3156)

One deliberate gate change: `plan.policy.agentsMdBudget` managed ratchet re-seeded from 162 to 166 lines and 16016 to 16530 bytes to admit the AGENTS card. The ratchet is a strict no-growth seed, so any card at all requires re-seeding; this is the remediation the gate itself prints. The card was cut to 4 lines first and the ratchet re-seeded to exactly the new size with zero slack. Recorded via `task decision:write` (`xbrief/decisions/2026-08-28-re-seed-the-agents-md-managed-section-ratchet-166-lines-16530-by.decision.json`). No verifier, floor, test, or eval fixture was touched.

## Verification

- `pnpm exec vitest run packages/core/src/init-deposit/hook-runtime-travel.test.ts packages/core/src/init-deposit/agent-hooks.test.ts` - 39 passed. New module: 93.3% statements, 91.3% branches, 100% functions.
- `pnpm exec vitest run packages/core/src/content-contracts/skills/agents_entry_contract.test.ts` - 76 passed, including the new propagation marker test.
- `task verify:agents-md-budget`, `task verify:encoding`, `task verify:ac` - clean.
- `task check` does **not** reach a verdict on this machine: it fails at `gate ts:check-lane` with `spawnSync task ENOBUFS`, an output-buffer limit in the check runner, not a product failure. Running the lane directly yields 16 failures across 6 files (doctor / session / occupancy); the **identical** 16 failures reproduce at the merge base `35da5f36`, so this branch introduces none. Filed separately.

Tracking: #3785. Split: #3888. Refs #3736, #3571, #2752, #2264, #3156.

Whether #3785 stays open is the operator's call - items 1 and 2 now live on #3888.


MScottAdams:
## Review dispositions (head `1e9a238c`)

**Greptile P1 - "Fresh registrations escape inspection" - valid, fixed.**

The probe asked git only for tracked paths, so the first `deft init` (which writes `.cursor/hooks.json` and does not stage it) stayed silent on exactly the run that precedes the commit carrying the fail-closed fence into every clone. The probe now asks for tracked **or** trackable-and-not-ignored paths (`git ls-files --cached --others --exclude-standard`), which is what "travels with the tree" means in the first place. An ignored registration still stays silent, because it cannot reach a clone.

Two real-git regression tests pin both directions in `packages/core/src/init-deposit/hook-runtime-travel.test.ts`:
- `warns on a first deposit, before the registration has been staged` - asserts `git ls-files` is empty and the warning still fires.
- `stays quiet when the consumer ignores the registrations: nothing travels`.

The same widening applies to the runtime anchor, so a brand-new project whose `package.json` is not yet committed is not warned at spuriously: registration and anchor are judged by the same travel rule.

**SLizard P3 - `CHANGELOG.md:18` embeds the `policy:disable-host-hooks` invocation - not taken.**

The command is in the entry deliberately, and is the release-notes obligation named in the issue bind: a consumer already locked out cannot read the PR body, and the alternative they will otherwise find (hand-editing `failClosed`) is silently reverted by the next `deft update`. The durable workaround has to be in the release notes to be reachable.

**SLizard advisory - `.deft/core/docs/hook-runtime-unavailable.md` "does not exist" - false positive.**

`content/docs/` in this repository is the source for `.deft/core/docs/` in a consumer deposit. AGENTS.md already references sibling pages by their consumer path (for example `.deft/core/docs/deft-directive-disable.md`), and the consumer path is the one a locked-out reader can actually open.

**SLizard advisory - brief statuses still `running`/`proposed` - expected.**

Lifecycle terminal transition is post-merge (`scope:complete`, refs #2321 / #3476). Marking the brief completed before merge is the failure mode those gates exist to catch.

## CI note - `verify:orphan-active` is red from a master-side condition

The merge gate fails at `verify:orphan-active`: *"2 active/running xBRIEFs look shipped but still consume WIP"*. Those two are `origin/master`'s `xbrief/active/` entries for #3835 and #3859 - both issues closed earlier today with their briefs left in `active/`. This branch adds exactly one active brief (#3785, open, unmerged) and the gate passes locally against it. The red clears when those two briefs are landed to `completed/` on master by their owners; it is not addressable from this branch.

MScottAdams:
## Review dispositions (head `74226f41`)

**Greptile P1 - "Nested repositories suppress warnings" - valid, fixed.**

The probe short-circuited on a missing `.git` entry in the project root. A project deposited into a subdirectory of a repository has no `.git` of its own, yet the parent is free to commit `.cursor/hooks.json` and carry the fail-closed fence to clones with no runtime - so the one shape where the deposit root is not the repository root was exactly the shape that went unwarned. The probe now asks `git ls-files` directly, which answers from any depth (paths relative to the deposit root) and still declines outside a repository. Regression test: `warns for a project nested inside a repository it does not root`, which asserts the nested directory has no `.git` and the warning fires anyway.

**SLizard P1 - "Missing test implementation ... `hook-runtime-travel.test.ts` is omitted from the diff" - false positive.**

The file is added by this PR and is present at every head it has had. `git show --stat` on the branch lists it; the acceptance command runs it green (17 tests at this head).

**SLizard P1 - "Missing recovery card in the managed section" - false positive.**

The card is present in both places the #1309 propagation contract requires: `AGENTS.md` inside the `deft:managed-section` block, and `content/templates/agents-entry.md`. `packages/core/src/content-contracts/skills/agents_entry_contract.test.ts` fails closed if either copy drops the exit-127 signature or the `policy:disable-host-hooks` recovery, and it passes at this head.

Previously dispositioned findings (CHANGELOG command path, `.deft/core/docs/` consumer path) are unchanged - see the earlier comment on this PR.

MScottAdams:
## Fix-batch cap reached (head `898fc53b`) - one open Greptile P1, awaiting operator consent

Three fix batches have been spent on this review ownership, which is the dual stop this worker was dispatched with. Halting automatic re-fix rather than opening a fourth.

**State at `898fc53b`:** CI green (0 failures), SLizard 0 findings, Greptile confidence 4/5 with one open P1.

**Batches spent (each a distinct, valid Greptile P1, each closed with a regression test):**
1. `4e4c34fc` - a registration untracked at deposit time escaped the probe, which is the state of every first `deft init`.
2. `74226f41` - a project nested inside a repository it does not root escaped the probe, because the check looked for a local `.git`.
3. `898fc53b` - an uncommitted `package.json` was credited as the runtime anchor, silencing the warning in the case that actually strands the clone.

**Open P1 - "Unresolvable anchors suppress warnings" - valid, not fixed.**

`readPin` reports `rawSpec` verbatim, so a committed `"@deftai/directive": "file:../directive"` (or `link:`, `portal:`, `workspace:`) counts as an anchor even though a fresh clone cannot install from it. The failure is narrower than the three above - it needs a committed non-portable spec, which is a local-development shape rather than a shipped one - but the direction is the same: the warning goes quiet where it should fire.

Proposed fix, if the operator authorises a fourth batch: reject non-registry spec prefixes (`file:`, `link:`, `portal:`, `workspace:`) when judging the anchor, with a warning line naming the spec as non-portable, plus a test per prefix. Contained to `hook-runtime-travel.ts` and its test.

**Also open: confidence 4/5 against the framework dogfood floor of 5.** With the P1 resolved this is a class C confidence-only hold under the review-cycle skill, which offers document / operator override / minimal polish rather than redesign.

No merge until an owner decides. The four merge preconditions in the dispatch envelope (CI green, Greptile 5/5 with zero P0/P1 on the merged head, `verify:orphan-active`, `verify:completed-tracked`) are not all met at this head.

MScottAdams:
## Operator disposition — merging at 4/5 with the staged-manifest finding accepted

**Human-origin decision, operator, 2026-08-28.** Recorded by the parent; not agent-authored. This waives the 5/5 dogfood floor for this PR only.

### The finding is accepted as real and not fixed

Greptile's current review (updated `22:31:34`, pinned to head `62a3d88c`) reports that an index-only staged `package.json` dependency is credited as a runtime anchor. That is correct and it is the **fifth** distinct hole found in the same reachability condition, after untracked-at-deposit-time, nested-project, uncommitted-anchor, and non-portable-spec — each of which was real and each of which was fixed, the last one at `62a3d88c` with an allowlist of what a clone can actually fetch rather than a denylist of prefixes.

### Why it is being accepted rather than fixed

**The deposit-time warning is best-effort legibility, not an authorization gate.** Nothing grants or denies on its verdict. An over-credited anchor produces **one missed warning** at deposit time — it does not admit a write, weaken `failClosed`, or change what the hook does. The consumer's protection is the fail-closed hook itself, which this PR does not touch, plus the recovery documentation it adds.

Contrast with the holes that *were* fixed: those were the common cases. An untracked registration at deposit time is the state of **every first `deft init`**, so leaving it would have made the warning near-useless in the situation it exists for. An index-only staged manifest is a narrow local-development shape.

The fix budget is spent — four batches, four real holes closed. A fifth round on a non-gating warning is not the best use of the remaining pre-release window, and the marginal case it closes is far narrower than the four already closed.

### What is NOT being waived

- `failClosed: true` is untouched. Absent, crashed, and timed-out remain one non-decision class.
- CI is green; SLizard reports zero findings; P0 count is zero.
- The PR still does **not** fix the exit-127 lockout, and its acceptance language says so. Legibility arrives before the fence and after the lockout, never during. Bound items 1 and 2 remain on #3888 with the portability evidence.

### Follow-up

The staged-manifest case should be carried on #3888 alongside items 1 and 2, so the reachability condition gets one more pass when the repo-local runner work is done rather than being lost here.

Refs #3785, #3888, #3156.

MScottAdams:
**Correction to my disposition comment above (`5458496177`) — it is wrong, and this PR did not merge at 4/5.**

`#3890` merged at **Greptile 5/5, P0=0, P1=0**, SHA-matched to head `5b7afc82` (final review updated `22:43:21`: *"No blocking failure remains"*). Merge commit `2a66cb43`.

**The staged-manifest finding was fixed, not accepted.** By the time my interrupt reached the worker, the fix was already pushed and re-reviewed clean, and the merge had completed. I wrote the disposition on the assumption the worker was still mid-fix; it was not. The waiver was never exercised and no finding was accepted unfixed.

Verified on master:

- `init-deposit/hook-runtime-travel.ts:103-108` reads the anchor **as committed at `HEAD`** via `readCommittedFile`, so a staged-only or working-tree-edited manifest reports `is not committed` rather than counting as an anchor.
- `:79` documents the predicate as **"Allowlist, not denylist"** — only a registry version, range, dist-tag, `npm:` alias, or remote URL silences the warning. That is broader than the four prefixes I originally specified, and closes the class rather than the instances.

**What has NOT changed:** this PR still does not fix the exit-127 lockout. Legibility arrives before the fence (deposit-time warning) and after it (recovery doc, AGENTS.md card), never during. Bound items 1 and 2 remain on #3888, and #3785 stays **open** because its acceptance criteria are not met — which is why lifecycle PR #3902 was closed.

The error was mine: I interrupted on an assumption about worker state instead of checking the review status first, then reported the resulting non-event as fact. Anyone reading this thread for release notes should take **5/5, finding fixed** as the record.

Issue #3893, opened 2026-08-28
bug(gates): a brief stranded in active/ by one merge reddens the merge gate on every other open PR


MScottAdams:
**Recut 2026-08-28 after a design critique.** Synthesis accepted at comment 5458204912; disposition map 5458200859; verified-claims table 5458204775. Corroborating evidence from a second session at 5458168565.

## The defect

`verify:orphan-active` runs inside `check:merge` and enumerates **every** running brief under `xbrief/active/` with no diff ownership. It sits in both `FRAMEWORK_CHECK_GATES` and `CONSUMER_CHECK_GATES`. So one brief stranded by any merge fails the merge gate on every other open PR, naming a file those PRs never touched.

Measured on 2026-08-28: PRs **#3879, #3883, #3887 and #3889** failed the same two lanes at 21:14, all on this verifier. #3879 then passed a re-run with its **head unchanged** once the strandings landed, and master's own run reported 0 running briefs, exit 0. Separately, two single-brief lifecycle PRs each removed one stranded brief and neither could pass alone, because each still evaluated the other's; a third PR had to carry both.

That deadlock is structural, not bad luck: a repo-wide predicate evaluated per PR means k open PRs go red for one stranded file, and n strandings mutually block their own n cleanups.

## The sharper finding: the gate is not preventive

`orphan-active` decides a linked PR shipped by reading REST `merged_at`. **During a candidate's own merge gate, that candidate is not merged, so its `merged_at` is null.** The gate cannot detect the residue the candidate is about to create. It detects debt already in the base, or an issue closed earlier.

So the current placement fails to stop the merge that creates the debt and then blocks unrelated PRs after the debt exists. It shifts recovery cost onto bystanders without closing the creation window.

## Two arguments that do NOT work, recorded so they are not retried

**Sibling-check precedent is false.** `verify:completed-tracked` is excluded from the merge gate, but not as placement precedent -- `gate-lists.test.ts` pins that exclusion so an in-flight lifecycle land PR is not deadlocked. The two checks differ in substance: `orphan-active` keys on issue closure **and** linked-PR `merged_at` (with a test requiring failure for an open issue whose linked PR merged), while `completed-tracked` treats an open issue as nonterminal and asks whether a tracked terminal artifact exists on the delivery tip. Relocating on the analogy would silently drop the merged-PR/open-issue class.

[...]

Issue #3284, opened 2026-08-11
feat(check,intake): product-first done-gate - stated acceptance criteria run FIRST and are never skippable; hygiene gates run second and degrade under pressure (architecture for #3267)


ape-deft:
**Evidence (internal benchmark transcripts, private - no benchmark name or scores on public surfaces):** every observed point loss across two harnesses and two framework versions is a verification-DIRECTION failure, not a capability failure:

- One run verified a self-built 7-check suite instead of the task's stated checks, found an out-of-scope defect, and spent its remaining budget on it.
- One run verified a self-chosen approximation of the stated commands; the only arm that passed was the one that ran the task statement's commands verbatim (#3267's evidence).
- Two runs spent 44-47% of all turns making the framework's own hygiene gates pass in a toolchain-incomplete container (#3282), squeezing product verification into the final 1-2 turns, where a real output divergence was rationalized away.

In the same runs, the process gates caught zero product defects. The framework points its verification budget inward (briefs, branches, toolchain, managed files) and leaves the product verified by whatever the agent improvises. A competing lightweight framework wins on the same tasks with a fraction of the machinery because its small process points outward at the task statement.

## Change

**(1) Intake captures stated acceptance criteria as executable AC.** xbrief schema gains `acceptance.commands[]` (verbatim command strings from the task statement, plus optional expected outputs). Empty is allowed only with an explicit `none_stated: true` marker so absence is a decision, not an accident.

**(2) `deft verify:ac` runs them verbatim.** Same commands, same flags, same working directory; output comparison; machine-readable pass/fail per command. Self-chosen verification is supplementary, never a substitute.

**(3) `directive check` inverts its ordering.** verify:ac runs FIRST and fails fast; hygiene gates (branch, toolchain, managed-file integrity) run second. Under budget pressure or in degraded environments (#3282 preflight), hygiene degrades to advisory - AC never degrades when commands exist.

**(4) Rapid profile = AC-only.** The ceremony dial's lightest setting keeps exactly one verification behavior: run the stated commands before declaring done. This is the rapid path's positive content, per the #3267 discussion.

## Acceptance sketch

Schema field + intake-skill guidance for capturing stated commands; verify:ac task with verbatim execution and output diff; check runs AC before hygiene with fail-fast; pressure/degraded profiles keep AC mandatory and hygiene advisory; docs in build + pre-pr skills. Test: a consumer task with stated commands where the product is subtly wrong fails check even when every hygiene gate passes.

[...]

Issue #3267, opened 2026-08-10
feat(verification,strategies): literal acceptance-command verification - capture exact stated commands at intake, run verbatim before done (extends #973)


ape-deft:
**Evidence (internal benchmark transcripts, private - no benchmark name or scores on public surfaces):** on one task, all three arms - no-framework, directive, and a competing lightweight framework - declared success in their final messages. Only one passed the official checker: the arm whose workflow had captured the task statement's EXACT commands as acceptance criteria and ran them verbatim before finishing (its final message quotes the literal invocations, shows matching outputs, and even flags a caveat about the exact command form). The other two verified with self-chosen approximations of the ask and shipped subtle mismatches they never saw.

Cheap, decisive, and currently nowhere in directive's light path.

## Proposed rule

At intake, record the task statement's literal acceptance commands / stated invocations (when present) as executable acceptance criteria. Before declaring done, run them VERBATIM - same paths, same flags, same working directory - and compare outputs. Self-chosen verification is supplementary, never a substitute.

Make this the rapid path's positive content: when the ceremony dial (#3214) selects rapid/minimal, this is the one verification behavior that must survive, because it has demonstrated score value at trivial cost.

## Acceptance sketch

Intake step captures stated commands into the plan/AC surface; done-gate runs them verbatim; rapid/minimal profiles include this check; guidance in build + pre-pr skills.

Extends #973 (machine-verifiable-spec - atomic claims with shell commands as acceptance gates). Composes #1395 (verify-ac), #1201 (pre/post conditions), #3214 (rapid path), #1176 (plan completeness). Relates #3266 (budget-aware sizing - bank THIS pass first).

Issue #3264, opened 2026-08-10
bug(lifecycle,gates): completed xBRIEFs can remain untracked after merge (#1358 recurrence)


MScottAdams:
## Summary

Post-merge `scope:complete` can leave completed xBRIEFs **only on a laptop** (untracked / never on origin). Product merges and issue closes still succeed. Release lifecycle sync and `verify:orphan-active` stay green because they watch **active** residue, not **completed-but-not-tracked**.

Operators repeatedly land hygiene PRs (e.g. #3262, #2882, #3247). That is recovery, not prevention.

## Prior art (not re-opened)

- **#1358** (closed) + PR **#1689**: swarm Phase 6 Step 2b skill text — must commit+push lifecycle moves after merge.
- **#1487** (closed): cohort `scope:complete` sweep on close.
- **#734** (closed): release refuses while closed issues still sit in non-terminal folders.
- **#2321** (closed): orphan **active**/running on master.

#1358 made the rule **discoverable**. Agents under **`drive-to: merge-ready` leaves** often never run Phase 6 Step 2b / `swarm:finalize-cohort`. Skill-only is not enough.

## Recurrence evidence

- 2026-08-10 epic/field cohort: many closed issues with product on master; `xbrief/completed/` missing on origin until chore **#3262**.
- Same family as post-#1358 chore completes (#2882, #3247).

## Design mode (v1)

**Deterministic, low-token, simpler-is-better.** One fail-closed check + one remediation. No owner-matrix theater, no LLM judgment, no multi-skill rewrite as the only fix.

### Rule (v1)

If a GitHub issue is **closed** and Directive has a **scope xBRIEF** for it (origin ref / known plan id under the lifecycle tree), then the **delivery branch tip** (default `master`/`main`) must contain a **tracked** artifact under `xbrief/completed/` or `xbrief/cancelled/` that references that issue. Else fail closed.

### Remediation (single path)

Prefer policy-safe land (branch policy #747):

```text
task swarm:finalize-cohort -- --stories <ids> --repo <owner/repo>
```

or an equivalent lifecycle PR (same class as #3262). Do **not** require direct commits to the default branch as the only path.

### Non-goals (v1)

- Replacing release active-folder sync (#734)
- Full leaf/parent DONE protocol rewrite (may be v2)
- Prompt NLP / free-text markers
- Issues that never had a scope xBRIEF
- Re-opening #1358 as incomplete skill work — this issue is **enforcement**

## Acceptance criteria

[...]

PR #3902, opened 2026-08-28
chore(xbrief): land completed scope for the hook-runtime lockout (#3785)

MScottAdams:
## Summary

`scope:complete` for #3785. PR #3890 squash-merged as `2a66cb43` and left the brief in `xbrief/active/`; this moves it to `xbrief/completed/` with per-item `x-directive/evidence`.

`verify:orphan-active` scans repo-wide inside `check:merge`, so a leftover active brief reddens the merge gate on every other open PR (#3893). Landing this immediately rather than batching.

## Verification

- `task scope:complete` exit 0; literal acceptance-command gate ran all four stated commands verbatim, all exit 0 (#3267 / #3284).
- Lifecycle-only change: one brief moved, no product code.

Refs #3264, #1358, #2321.

### DIFF
--- a/xbrief/completed/2026-08-28-3785-bughooks-fail-closed-hook-registration-travels-without-a-hos.xbrief.json
+++ b/xbrief/completed/2026-08-28-3785-bughooks-fail-closed-hook-registration-travels-without-a-hos.xbrief.json
@@ -2,11 +2,11 @@
   "xBRIEFInfo": {
     "version": "0.8",
     "description": "Scope xBRIEF for GitHub issue #3785 (partial: bound items 3 and 4; items 1 and 2 split out on portability evidence)",
-    "updated": "2026-08-28T20:30:45Z"
+    "updated": "2026-08-28T22:50:08Z"
   },
   "plan": {
     "title": "bug(hooks): warn when a fail-closed hook registration travels without a hostable runtime",
-    "status": "running",
+    "status": "completed",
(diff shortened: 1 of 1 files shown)


MScottAdams:
Closing on operator instruction — **#3785 stays open deliberately, because its acceptance criteria are not met.**

This PR renames the brief into `completed/` and sets `plan.status: completed`. That would assert full delivery for a scope that is partially delivered.

What actually shipped in PR #3890: bound items **3** (recovery documentation) and **4** (deposit-time travel warning). Bound items **1** (a no-Node `deft-hook` artifact / repo-local runner) and **2** (a legible deny with a decision code instead of exit 127) did **not** ship — they are split to #3888 with the portability evidence, along with the staged-manifest finding dispositioned at 4/5.

Concretely: a consumer with no Node still gets exit 127 at the moment of denial. Legibility arrives before the fence and after the lockout, never during. Marking the scope complete would record the opposite.

The brief correctly remains in `xbrief/active/` on master. `verify:orphan-active` is quiet because #3785 is open, so this is not lifecycle debt and it does not redden any merge gate.

No criticism of the worker — it was following a standing instruction to run `scope:complete` immediately after merge, which is right in the ordinary case and wrong for a deliberate partial landing. The operator resolved the ambiguity after the PR was opened.

Re-open when #3888 lands the remaining items.

Issue #3930, opened 2026-08-30
bug(ci,contracts): master task check is red -- a content contract pins `Closes #3755` to `[Unreleased]` and the v0.108.0 cut moved it


MScottAdams:
## Summary

`task check` is red on `origin/master` and has been since the v0.108.0 cut. One content-contract case fails:

```
FAIL packages/core/src/content-contracts/standards/occupancy_membership_docs.test.ts
  > occupancy boundary naming (#3755) > carries one Unreleased changelog line for the membership change
AssertionError: expected [] to have a length of 1 but got +0
```

This blocks the `Merge gate (task check)` required check on every PR branched from master, independent of what the branch changes.

## Evidence that it is pre-existing

Master tip `c6761881914d3eb02f44f379f9e1e02a94ff708f` check-runs:

```
Merge gate (task check)          | completed | failure
TypeScript (build + lint + test) | completed | failure
Merge gate (Blacksmith primary)  | completed | failure
```

Re-deriving the assertion's own logic against pristine `origin/master:CHANGELOG.md` gives `0`, so the case cannot pass on master:

```
$text -split "## [Unreleased]" -> [1] -> split "\n## " -> [0]
lines containing "Closes #3755" = 0
non-blank lines in that section  = 4   (the four ### headings, all empty)
```

## Mechanism

The assertion pins a specific changelog line to the `## [Unreleased]` section forever:

```ts
const unreleased = changelog.split("## [Unreleased]")[1]?.split("\n## ")[0] ?? "";
const mentions = unreleased.split("\n").filter((line) => line.includes("Closes #3755"));
expect(mentions).toHaveLength(1);
```

`Closes #3755` was under `[Unreleased]` when the contract was written. The v0.108.0 release cut moved it into `## [0.108.0] - 2026-08-29`, which is what a release cut is supposed to do. The contract has no notion of "or the section it was released into", so it goes red on the first cut after it lands and stays red.

This is a class, not a one-off: any content contract that asserts a literal under `## [Unreleased]` becomes a scheduled failure at the next release.

## What needs deciding

Not prescribing the remedy; the options differ in what they protect:

1. Scope the assertion to "present in `[Unreleased]` **or** the most recent released section", keeping the documentation guarantee across cuts.
2. Re-point the contract at the durable surfaces it already checks in its two sibling cases (`occupancy.ts`, `commands.md`) and drop the changelog case, on the grounds that the changelog is a release artifact rather than a durable contract.
[...]

Issue #3877, opened 2026-08-28
bug(session,tests): mint-path session fixtures read ambient DEFT_SESSION_ID, reddening the suites and scope:complete


MScottAdams:
**Recut 2026-08-28 after a design critique.** Synthesis accepted at comment 5457549697; disposition map 5457450968; verified-claims table 5457549558. Duplicate #3836 folded in at 5457372427 -- its evidence names both failing tests verbatim.

## The defect

Two **mint-path** session fixtures omit an env bag, so `resolveOccupancySessionId`'s `input.env ?? process.env` reads whatever the caller's shell exported:

```text
FAIL packages/core/src/session/occupancy.test.ts > worktree occupancy lease (#3433)
  > mutation session:start claims and a second session is denied
  expected '<exported id>' to be 'first-sess'

FAIL packages/core/src/session/session-start-rearm.test.ts > session re-arm vs cold ceremony tiers (#2992)
  > re-arm refreshes ritual without tools/triage/release/tickler
  expected '<exported id>' to be 'rearm-session'
```

Both pass a `newSessionId` stub and then assert the minted id. The same occupancy suite already uses `stubEnv` on close-out cases, so the authors isolate when they mean "unset" -- these two cases simply do not.

It matters beyond a red suite: `scope:complete`'s literal acceptance-command gate (#3267) runs a brief's `verify_commands` verbatim, so a story whose commands include the session suite cannot complete while the variable is set. Observed live completing #3769, then passing after unsetting, same tree, no code change.

## What this issue is NOT

**Env-over-mint is shipped product law, not a fixture bug.** `resolveOccupancySessionId` resolves explicit id, then `DEFT_SESSION_ID`, then mint, and two named tests lock it: "prefers `DEFT_SESSION_ID` over a mint" and "re-arm with `DEFT_SESSION_ID` keeps the occupant id and does not mint". An earlier draft of this issue proposed asserting that fixture ids are not overridable by the environment. That would invert the order, make `session:start` ignore a set variable, and split from the lease its own freshness line asks the operator to adopt. **Do not build it.**

**The literal-acceptance runner is not the bug.** It inherits `process.env` because #3267 requires verbatim execution. Sanitising it would green `scope:complete` while the identical command stayed red in a worker shell -- a false green against the contract that exists to prevent that.

## Correction to this issue's original framing

[...]

PR #3931, opened 2026-08-30
fix(test,session): stub the package-manager seam and drop the local timeout override (#3901)

### LINKED ISSUE #3901: bug(test,windows): the --with-network session:start case makes a real npm registry request under a 20s budget and is load-flaky
Bound by design-critique arc: successor lean 5466043459, verified-claims table 5466045455, synthesis accepted 5466045514, critic 5466038275. This body carries the bound design; it is the bind product, not a recut.

## Summary

`session:start --with-network probes only the disclosed public npm registry` (`packages/core/src/session/package-manager-network.test.ts`) makes a **real network request to registry.npmjs.org** and is load-flaky on Windows. It reddens `task check` and `task verify:ac` for reasons unrelated to the change under test.

Found during the #3755 delivery run on 2026-08-28. Confirmed pre-existing: the worker reverted `packages/core/src/session/occupancy.ts` to `origin/master` and reproduced the failure.

**There are two causes, not one.** The registry round trip is the visible one. The other is a local timeout override that contradicts the project's own recorded win32 policy, and it survives any fix to the first.

## Observed

Same host, same day, no source change between runs: failures at 14:27, 14:44, 14:47 and 16:04; passes at 14:33, 15:12 and 16:11. Four failures, three passes. It failed `task check` / `verify:ac` three times and cost roughly 25 minutes of re-runs in a single story.

## Cause 1: the mock is a pass-through spy, not a stub

The file header says the package-manager seam "is stubbed at the lowest level". It is not:

```ts
vi.mock("node:child_process", async (importOriginal) => {
  const actual = await importOriginal<typeof import("node:child_process")>();
  return { ...actual, spawnSync: vi.fn(actual.spawnSync) };
});
```

`vi.fn(actual.spawnSync)` records **and executes**. A direct Vitest runtime check confirms one real invocation plus one recorded call.

So the case really runs `npm view @deftai/directive version --registry=https://registry.npmjs.org/ --ignore-scripts` on an inner `timeout: 5_000`.

**Correction to the original filing:** this call does **not** come from `doctor/payload-staleness.ts`. The `--with-network` session path reaches `probeSessionReleaseAvailability`, whose `defaultNpmView` in `packages/core/src/session/release-availability.ts:47-59` builds that argv with the 5-second timeout. The doctor seam is separate and uses 15 seconds. The original attribution was inherited from the test file's own header comment and was wrong.

## Cause 2: the case overrides the project's win32 timeout policy

`vitest.config.ts:210-225` sets `testTimeout: 120_000` on win32 and records why: Windows process creation drains through a global lock, "a git-fixture test issuing 10-30 sequential spawns therefore takes 25-80s", the value is "120s, not 20s (#3616)", and lowering `maxWorkers` or re-serialising files is explicitly forbidden as a fix.

This case overrides that back down to `timeout: 20_000`, while `initPrivateScopeRepo` performs **six sequential git subprocesses** (init, two config calls, add, commit, rev-parse). Those six spawns remain after npm is stubbed. A 20-second cap against a documented 25-80 second cost is the second cause, and it is independent of the network.

## What stubbing does and does not cost

The #2182 assertion is the **recorded npm/pnpm argv**, not registry behavior. The expected value is an exact singleton naming `@deftai/directive`, the canonical public registry, `--ignore-scripts`, and the timeout. An added private-registry call breaks the singleton comparison; a dropped public probe empties the list. The test permits exit code 0 or 1 and never asserts npm status or stdout, so the real response supplies no assertion evidence.

Stubbing therefore preserves the security boundary exactly. The real round trip buys no coverage of any claim this test makes.

**Implementation shape:** use the low-level command-selective stub already used at `packages/core/src/doctor/npm-view.test.ts` -- stub npm/pnpm command execution, return a well-shaped deterministic process result, keep the exact singleton argv assertion. Do **not** replace this with a high-level `probeReleaseAvailability` injection: that bypasses construction of the security-relevant argv and weakens #2182. High-level seam injection is the right pattern for `release-availability.test.ts`, not here.

## Scope boundary on "no network"

`allowOptionalNetwork: true` reaches `resolveSessionScmReadiness` with `depth: "deep"`, and the deep runner can execute **live `gh` API validation** on a 30-second subprocess timeout (`packages/core/src/scm/readiness.ts:285-315`). The case stubs triage, staleness and orientation, but not `probeScm`.

So a package-manager stub cannot support a process-wide no-network claim. Either scope the claim to npm/pnpm, or inject a deterministic SCM probe and add a control establishing no other outbound path.

## Acceptance criteria

1. The `--with-network` case issues **no npm or pnpm network request**. If the stronger claim of no outbound request at all is wanted, the SCM probe is injected too and a control test establishes it; otherwise the criterion stays scoped to npm/pnpm.
2. It still fails when a flow contacts the private-scope registry, and still fails when the public-registry probe is dropped -- both as explicit mutation checks, proving the #2182 assertion is intact and non-vacuous.
3. The local `timeout: 20_000` override is removed, or retained with a recorded justification that survives the `vitest.config.ts:210-225` rationale. Reducing or reusing the six-spawn git fixture is an acceptable alternative route.
4. A deterministic no-registry control is the gate-worthy evidence. Any repeated-run smoke check names an exact command and an explicit concurrency and load shape; "twenty green runs" without that is not a reproducible oracle.
5. The file header no longer claims a stub where a pass-through spy is used, and no longer implies the doctor payload-staleness seam.
6. Any remaining real-network test in this suite is marked so it can be excluded from gate runs.

Refs #2182, #3214, #3286, #3480, #3616, #3755.

MScottAdams:
## Summary

`session:start --with-network probes only the disclosed public npm registry` made a real request to `registry.npmjs.org` on every run and was load-flaky on Windows. This fixes both recorded causes.

**Cause 1 -- the mock was a pass-through spy.** `vi.mock("node:child_process")` returned `spawnSync: vi.fn(actual.spawnSync)`, which records *and* executes. The file header claimed the seam was "stubbed at the lowest level"; it was not. The call comes from `probeSessionReleaseAvailability` -> `session/release-availability.ts:47-59` (`defaultNpmView`, `timeout: 5_000`) -- not `doctor/payload-staleness.ts` (15 s), which the old header implied.

**Cause 2 -- the case overrode the win32 timeout policy.** `vitest.config.ts:210-225` sets `testTimeout: 120_000` on win32 and records why: sequential process creation drains through a global lock, so a git-fixture test issuing 10-30 spawns takes 25-80 s. This case overrode that back to `timeout: 20_000` while `initPrivateScopeRepo` runs six sequential git subprocesses. That cause survives any network fix, and the measurement below shows it was about to bite on its own.

## What changed

`packages/core/src/session/package-manager-network.test.ts` only, plus the CHANGELOG entry and the scope xBRIEF.

- The `spawnSync` mock is now a **command-selective stub**: npm and pnpm return a well-shaped deterministic `SpawnSyncReturns`, every other command still executes because the git fixture needs it. This is the low-level pattern from `packages/core/src/doctor/npm-view.test.ts`. The exact singleton argv assertion is unchanged, so the #2182 security boundary is preserved. A high-level `probeReleaseAvailability` injection was **not** used -- that would bypass construction of the security-relevant argv.
- The stub also serves doctor's offline `npm config get` reads, which lets the `defaultRitualRunner` case drop its local `mockImplementation`. That block called `spawnSyncMock.getMockImplementation()` from inside its own implementation -- a latent self-recursion that stayed dormant only because no other command reached it.
- `probeScm` is injected in every session-start case (see the scope decision below).
- The local `timeout: 20_000` override is **removed**. The case now inherits the project policy instead of contradicting it. No `maxWorkers` change and no re-serialisation -- `vitest.config.ts:220` forbids both.
- The file header now names the session release-availability seam and no longer claims a stub where a spy was used.

## Scope decision on acceptance criterion 1

Criterion 1 allowed either scoping the claim to npm/pnpm, or injecting the SCM probe and adding a control. **I took the stronger claim.**

`allowOptionalNetwork: true` reaches `resolveSessionScmReadiness` with `depth: "deep"`, whose runner spawns live `gh` on a 30 s subprocess timeout (`scm/readiness.ts:285-315`). The shallow path used by the two sibling cases spawns `gh auth status` on a 15 s timeout (`readiness.ts:355-380`). `probeScm` is already a first-class injectable seam on `SessionStartOptions`, so injecting it costs about fifteen lines and removes both an outbound request and a second wall-clock hazard that criteria 3 and 4 exist to remove. Scoping to npm/pnpm would have satisfied the criterion on paper while leaving live `gh` inside a package-manager assertion.

The control is enforced structurally rather than by inspection: the stub records every command it passes through to the real `spawnSync`, and `afterEach` fails closed if any of them is network-capable (`npm`, `pnpm`, `gh`, `ghx`, `curl`, `wget`). That satisfies criterion 6 by construction -- no real-network case remains in this file, and a future one has to be marked and excluded before it can land.

## Evidence

**Criteria 1 and 6 -- the control is non-vacuous.** Deleting only the `probeScm: hermeticScmProbe(observedDepths)` line and re-running:

```
AssertionError: expected [ 'gh' ] to deeply equal []
AssertionError: expected [ 'gh', 'gh', 'gh' ] to deeply equal []
AssertionError: expected [] to deeply equal [ 'deep' ]
Tests  5 failed | 3 passed (8)
```

Suite wall time also rose from 18.8 s to 25.9 s. Restored, all 8 pass and `outboundExecutions()` is empty for every case.

**Criterion 2 -- two mutation checks.** Both run the real `runSessionStart`, then assert the shared `expectOnlyDisclosedPublicRegistryProbe` predicate throws:

- *added private-scope registry call*: the real run is asserted clean first, then a second flow reaches `https://npm.internal.example.com/` through the same recorded seam; the singleton comparison fails.
- *dropped public probe*: same configuration with `probeReleaseAvailability` injected to issue nothing; the recorded list is empty and the predicate fails.

**Criteria 3 and 4 -- the deterministic control is the gate-worthy evidence; the load run is supporting.**

Deterministic (gate-worthy):

```
pnpm exec vitest run packages/core/src/session/package-manager-network.test.ts
-> 8 passed, 15.9 s, zero outbound executions
```

Repeated-run smoke, exact shape:

- command: `pnpm exec vitest run packages/core/src/session`
- concurrency: project default `maxWorkers` = 8 (`min(12, floor(32 * 0.25))`, `vitest.config.ts:18`), win32, 32 logical cores
- load shape: 55 files / 572 cases per repetition executed at that worker count; 10 sequential repetitions; no other vitest process on the host
- result: **10/10 exit 0, 0 failures.** Per-repetition file totals 29.7-87.6 s.

The load run also settles cause 2. With `--reporter=verbose` under that same contention the target case took **18,784 ms with zero network I/O**. Its removed cap was 20,000 ms -- 1.2 s of headroom, spent entirely on six git spawns. Stubbing npm alone would have left the case within 6% of its own timeout, which is why removing the override was required rather than cosmetic.

**Criterion 5.** Header rewritten; it names `probeSessionReleaseAvailability` -> `release-availability.ts` `defaultNpmView` and states that `doctor/payload-staleness.ts` is a separate seam not exercised here.

## Gate integrity (#3156)

The removed `timeout: 20_000` is a **local override of the project gate**, not the gate. The project cap (`vitest.config.ts:225`) is untouched, `maxWorkers` is untouched, and no file was re-serialised -- both are explicitly forbidden at `vitest.config.ts:220`. Nothing was weakened to go green: the change removes work (a network round trip and three `gh` spawns) rather than widening a budget, and it adds three assertions that did not exist before.

## `task check` result and pre-existing failure

```
Test Files  1 failed | 1072 passed | 6 skipped (1079)
     Tests  1 failed | 15485 passed | 152 skipped (15638)
```

The single failure is **pre-existing on master and unrelated to this branch**: `occupancy_membership_docs.test.ts` pins a changelog literal to the `[Unreleased]` section, and the v0.108.0 cut moved that line into the released section. Tracked as **#3930**, filed from this branch with the evidence. Master tip `c6761881` already shows `Merge gate (task check)`, `TypeScript (build + lint + test)` and `Merge gate (Blacksmith primary)` as `failure`, and re-deriving the assertion's own logic against pristine `origin/master:CHANGELOG.md` returns 0. It is not fixed here because that would mean editing the failing gate from an unrelated branch (#3156).

Two further failures appeared in an earlier local run and are **not** reproducible in a clean environment: `session-start-rearm.test.ts` and one `session-start.test.ts` case read an ambient `DEFT_SESSION_ID` exported by the operating shell. That is **#3877**, already open. With the variable unset, both pass.

## Related Issues

Closes #3901

Refs #2182, #3616, #3480, #3286, #3214, #3755, #3877, #3930.

## Checklist

- [x] `/deft:change <name>` — N/A, 3 files and not cross-cutting (one test file, CHANGELOG, scope xBRIEF)
- [x] `CHANGELOG.md` — entry added under `[Unreleased]` → `### Fixed`
- [x] `ROADMAP.md` — N/A, not a roadmap-tracked item
- [x] Tests pass locally — 15485 passed, 152 skipped, 1 failed (pre-existing #3930, see above); targeted suite 8/8; 10/10 load repetitions

## Post-Merge

- [ ] **Verify issue auto-close**: confirm #3901 closed after squash merge
- [ ] `task scope:complete` on the active xBRIEF


MScottAdams:
## Review cycle state — merge-ready on review, blocked on a pre-existing CI red

**Greptile Step 6 fail-closed all-of is satisfied on HEAD `22af6fde4418498106bb3887d6a99b71f3c11810`:**

```
sha_match=true  last_reviewed=22af6fde4418  confidence=5  p0=0  p1=0  errored=false
clean_gate_holdout=ci_failures
```

Greptile review 2: *"The PR appears safe to merge. No blocking failure remains."* Confidence 5/5 meets the dogfood floor of 5.

### Fix batch (one commit, `22af6fde`)

| Finding | Severity | Disposition |
|---|---|---|
| CHANGELOG.md — historical releases gain false fix | P1 | Fixed. Valid and mine: the entry was inserted with a whole-string replace, so it landed in all 15 sections with an empty `### Fixed` block. Restored from `origin/master` and re-inserted once, bounded to `[Unreleased]`. Diff is now a single `+` line. |
| Outbound guard misses remote git | P2 | Fixed. Both `spawnSync` and `execFileSync` are now wrapped into one record, and a `git` call counts as outbound when its subcommand contacts a remote (`fetch`, `pull`, `push`, `clone`, `ls-remote`, `remote update`). Local git stays exempt — it is the fixture's own cost. A new case proves the classifier in both directions and that the fixture's six git spawns are recorded and local. |
| Discarded return value in the private-scope mutation check | P2 | Fixed. The injected call is asserted to have been answered by the stub, which is also what makes the mutation traffic-free. |

Addressed 1/1 P1 and 2/2 P2, 0 deferred.

### The remaining blocker is not this branch

```
ci_failed_checks: Merge gate (task check), TypeScript (build + lint + test),
                  Merge gate (Blacksmith primary), TypeScript (Blacksmith primary) / run
```

All four resolve to one failing case, `occupancy_membership_docs.test.ts`, which is **red on `master` and unrelated to this PR**. Tracked as **#3930**.

CI on this HEAD:

```
packages/core/src/session/package-manager-network.test.ts (9 tests) 424ms   <- this PR, green
Test Files  1 failed | 1077 passed | 1 skipped (1079)
  FAIL occupancy_membership_docs.test.ts > carries one Unreleased changelog line
```

Master tip `c6761881` shows the same three checks as `failure`, and re-deriving that assertion's own logic against pristine `origin/master:CHANGELOG.md` returns 0. Master has not moved since this branch opened, and #3930 is still unfixed.

It is not fixed here. The remedy is a decision among three options (see #3930), and choosing one silently inside an unrelated PR is what #3156 forbids without its own issue and review. Admin-merging past it is also not available: the outage playbook covers CI weather, not a genuine failing assertion.

### Resume

```
RESUME: residual=external-ci leftover=n/a conf=5/5 floor=5 standing=one-shot PR=https://github.com/deftai/directive/pull/3931
BLOCKER: #3930 must land on master, then this branch needs only a re-run (or a rebase) to go green.
```

Post-merge duty for whoever lands this: `task scope:complete` on `xbrief/active/2026-08-29-3901-*.xbrief.json` with `--merge-commit` and `--pr 3931`, then `task verify:orphan-active -- --issue 3901` and `task verify:completed-tracked -- --issue 3901`.

Issue #3448, opened 2026-08-18
feat(review-cycle,swarm): standing residual until resolved conf floor or same-fingerprint loop


MScottAdams:
## Summary

`#3273` made dual-stop resume discoverable, but **pursue residual** is still **one pass then park**. A through-merge / drive-to:merge-ready cohort then leaves PRs sitting at **below the resolved Greptile floor** when the bot still names **one leftover on files the PR already touches**.

That reading is wrong for **directive dogfood (5/5)** and for **consumers** (default **4/5**, or whatever `plan.policy.review.minGreptileConfidence` is). The floor is already resolved by `#3095`. The missing rule is **when to keep going** vs **when to park**.

Field class (2026-08-18, deftai/directive): after `pursue residuals until told otherwise or an endless loop`, the parent kept moving CI-red / dirty PRs and **parked** PR 3432 and PR 3408 at 4/5. Both had a single concrete leftover on already-touched files. The operator had to re-point: work those until 5/5 or loop.

## Problem

Three rules collide and no skill says which wins:

| Rule | What an agent does |
|------|--------------------|
| Dogfood / `#3095` floor | PR is not CLEAN below resolved `minGreptileConfidence` |
| `#3273` | `pursue residual` = **one** pass on the unit that just dual-stopped, then re-stop |
| `#2881` | Confidence-only hold (0 P0/P1) = at most one polish; out-of-AC = follow-up issue, do not expand |

So a standing order like **pursue residuals until a loop** gets read as **only new leftovers** (CI red, dirty rebase, operator-named hole). A 4/5 (or consumer 3/5) PR with one named file leftover waits for the human to re-point.

`#2881` also over-fires: **dropped return value / wrong remediation on a file this PR already owns** gets labeled new-subsystem and parked. The P0 carve-out (defect on already-touched files) should have won.

## Goals

1. **Portable floor.** Keep going until confidence meets the **resolved** `#3095` floor (`plan.policy.review.minGreptileConfidence`: typed project policy > framework-source dogfood **5** > consumer default **4`). ⊗ Hard-code 5/5. ⊗ Lower the project policy to clear one PR.
2. **Typed leftovers** so park vs continue is not invented:
   - **A — named leftover on already-touched files** (wrong remediation, dropped field, fixture, same-module hole) → **in-AC residual**. Keep going until floor or same-fingerprint loop.
   - **B — new subsystem / AC fight** → park or file a follow-up. Do not expand mid-babysit.
[...]

PR #3935, opened 2026-08-30
fix(contracts): pin the #3755 changelog entry by content, not by section (#3930)

### LINKED ISSUE #3755: fix(session): the occupancy lease admits a session lineage, and nothing says so
## Summary

The occupancy lease admits **a session lineage** -- the holder, and the children it dispatches -- not a single process. That is the intended design, it is what the code already does, and it is stated nowhere.

Recut after a design critique (successor lean 5427856058) and an operator decision. The original framing ("spawn tools bypass the gate, document it") was refuted: gating spawn fixes neither failure case, and the exemption is what makes the situation silent rather than what creates it.

## The decision

**A child dispatched by the lease holder operates under the holder's authority and inherits its rights.**

Operator-decided 2026-08-26 after the round-1 critique. This resolves a contradiction the tree has carried since #3433:

- #3433's record states the guarantee as "one live writer per worktree" and asserts "Children still use their own worktrees."
- `core-phase-5-6.md:26` requires a Tier-1 monitor to background-dispatch a continuation leaf `drive-to: merge-ready`, **same worktree**. Repeated at `core-phase-0.md:36` and `review-cycle/SKILL.md:514`. Those leaves write.

Both cannot be true. The decision keeps the mandated flow and corrects the guarantee: mutual exclusion is **between session lineages**, not between writers.

## Why gating spawn was the wrong remedy

Recorded so it is not re-proposed. The lease resolves identity from `DEFT_SESSION_ID`, which a spawned process inherits by default.

- **A child that inherits the id** presents the owner's identity, so a spawn gate would admit it, and its writes are admitted today. Gating spawn is a **no-op** for this case -- which is the common, mandated one.
- **A child with a different id or none** is dispatched successfully and denied at its first product write. A spawn gate would deny it earlier. Better failure timing, no hole closed, and it would break the mandated continuation-leaf flow if leaves receive fresh identities.

So the original proposal was either inert or harmful. The exemption is a faithful implementation of #3433's acceptance criterion ("PreToolUse denies **product-path writes** when occupied"), which never considered dispatch -- scoped by omission, not chosen. It entered in the same commit as the lease, with no comment at the site, no test pairing a lease with a spawn tool, and no decision record.

## What to build

**1. State the guarantee.** `occupancy.ts:9-15` currently says "mutual exclusion under crash-free operation" with no actor definition. It must say mutual exclusion holds **between session lineages**, and that a child dispatched by the holder operates under the holder's authority. Mirror it wherever the occupancy contract is described for consumers.

**2. Fix the stranded-child case.** A child dispatched into a leased tree without the lineage identity should either receive it at dispatch or not be dispatched there. Dispatch-then-deny-at-first-write is the worst of both: the work is already underway when it fails.

**3. Stop advertising steal against the parent.** `occupancy.ts:107-109` hands a denied caller a copy-pasteable `session:start --steal --confirm --occupant <reported-session-id>`. For a stranded child that is a command to evict its own parent, delivered through the channel agents treat as instruction. Deny text for a same-tree child must not be an escalation affordance. (The same string is flagged from the operator's side on #3729.)

**4. Pin the three deny codes by behaviour.** `actor = null` for spawn disables the identity and ritual-owner checks as well as occupancy, so three codes are unreachable for spawn -- `occupancy-occupied`, `occupancy-identity-unavailable`, `occupancy-ritual-mismatch`. A test written from the old framing would have pinned only the first. Pin all three by deny code, **not by line number**: cites in this area have drifted three times during this cluster.

**5. Write the decision record** for the lineage guarantee, so the next reader finds the decision rather than re-deriving the contradiction.

## Deliberately not in scope

- **Making inheritance explicit rather than ambient.** Today a child is the owner because it inherited an environment variable. Under the ratified cooperative model (`occupancy.ts:9-15`, and the #3611 decision record) that is adequate fidelity, and an explicit grant is ceremony until something needs the lineage to be legible. Revisit if #3769 or close-out reasoning needs to attribute a write to a lineage.
- **The ritual-state race** -- split to #3769. A denied foreign write still rewrites the owner's ritual state, and a foreign `session.compact` locks the owner out. Neither involves a spawn tool.
- **Other gate exits.** Assist/scratch writes, `ephemeral` and `explore` spawns, and unclassified Shell redirects all bypass occupancy too. Item 1's prose must be scoped to what the gate actually covers -- non-spawn, non-scratch, classified direct writes by a mismatched lineage -- and must not imply spawn is the only exception.
- Occupancy admission semantics generally (#3611, closed).

## Already done, do not re-land

`commands.md:403` already reads "spawn tools remain outside occupancy enforcement under #3755", and `:399` already corrects the authorization reading. Both landed in `8f5dd900`. State the **rationale** beside that disclosure rather than repeating the disclosure.

## Acceptance

- [ ] The occupancy module contract states mutual exclusion between session lineages, and that a dispatched child operates under the holder's authority
- [ ] A decision record captures the lineage guarantee and why "one live writer per worktree" was withdrawn
- [ ] A child dispatched into a leased tree either carries the lineage identity or is refused at dispatch -- never dispatched then denied at first write
- [ ] Deny text shown to a same-tree child does not advertise stealing from its parent
- [ ] Tests pin all three spawn-unreachable deny codes by code, not by line
- [ ] Consumer-facing prose is scoped to the gate's real coverage

## Provenance

All `measured` claims come from the round-1 critic (comment 5427665352), run via a `tsx` harness against source at `0a82ac96` using the same seam shapes as `dispatcher.test.ts`, with occupancy claimed only in throwaway temp roots. The parent verified that `commands.md:399` and `:403` landed in `8f5dd900`, and that the "one live writer" phrasing survives only in #3433's completed brief rather than in live prose.

Refs #3433, #3611, #3729, #3769, #3613.

## What the lineage guarantee costs -- state it, and pin it

Added after the operator decision, from the follow-on discussion. This is the consequence of the decision and belongs in the same change.

**Occupancy provides no mutual exclusion inside a lineage.** If the holder dispatches N children that inherit its identity, all N are admitted as the owner. The gate cannot order them, cannot detect them, and will never deny one. That was already true; the decision makes it stated rather than implied.

So the honest scope is: **a coarse inter-lineage mutex.** It stops a second operator session, another cohort, or an unrelated agent from mutating the tree. It says nothing about the holder's own children.

What actually bounds intra-lineage concurrency is topology and protocol, not the lease:

- **Worktree isolation**, which does most of the work -- each swarm worker gets its own tree, and `resolveWorktreeMap` refuses same-path collisions. A real gate.
- **The delivery-attempt ledger**, `DENY_DUPLICATE_ACTIVE` with `maxActiveAttempts: 1`. Deterministic, but keyed on *story*, not on worktree, so it does not generalise.
- **"Dispatch ONE continuation leaf"** -- singular, and the monitor polls rather than writes while it runs. This is the weakest of the three because it is prose. **Unverified:** whether anything enforces the "one", and whether a Tier-1 monitor writes concurrently with its leaf.

**Residual risk, recorded rather than resolved:** a parent and its continuation leaf writing the same tree concurrently is permitted by every mechanism listed above. Reachability is unmeasured.

### Additional acceptance

- [ ] A test pins that a **same-lineage second writer is admitted** -- asserting the permissiveness, not just documenting it. Without this, a later reader takes "mutual exclusion" in the module contract to cover their children and builds on a guarantee that never existed. The test exists to make the boundary legible and to stop it being filed as a bug.

### Note for #3769

This decision makes #3769 **more** serious. If the lineage is the unit of exclusion, the inter-lineage boundary is the only thing occupancy is still solely responsible for -- and that is exactly where a foreign session can rewrite the owner's ritual state and lock it out.

### LINKED ISSUE #3930: bug(ci,contracts): master task check is red -- a content contract pins `Closes #3755` to `[Unreleased]` and the v0.108.0 cut moved it

MScottAdams:
## Summary

`task check` has been red on `origin/master` since the v0.108.0 cut. One content-contract case bounded its changelog search to the `## [Unreleased]` section:

```ts
const unreleased = changelog.split("## [Unreleased]")[1]?.split("\n## ")[0] ?? "";
const mentions = unreleased.split("\n").filter((line) => line.includes("Closes #3755"));
expect(mentions).toHaveLength(1);
```

The cut moved that entry into `## [0.108.0] - 2026-08-29`, which is what a cut is supposed to do, so the bounded slice went empty. The assertion encoded a false invariant -- the entry's section is transient, the entry is not -- and it blocked the `Merge gate (task check)` required check on every PR branched from master, independent of what the branch changed.

The case now searches the whole file and requires exactly one entry carrying both of its tokens:

```ts
const entries = changelog
  .split("\n")
  .filter((line) => line.includes("Closes #3755") && line.includes("occupancy:grant"));
expect(entries).toHaveLength(1);
```

The changelog is append-only across cuts, so exactly-one holds wherever the entry lives -- no section parsing, no new release machinery.

## Remedy provenance

Design-critique bound: accepted lean 5466430201, critic 5466425900, verified-claims table 5466430284. This is the critic's fourth option, which the issue body did not name. The three options in the body were each rejected: option 1 ("Unreleased or the most recent released section") is a one-cut grace period that also needs latest-section logic not exposed today; option 2 (drop the case) weakens the gate, since the two sibling cases pin occupancy behaviour and operator documentation while this third case independently pins release-note discoverability; option 3 (post-cut release obligation) is prose rather than a control.

The class premise was refuted by measurement (denominator 1): of 15 candidate test files containing both `CHANGELOG.md` and `Unreleased`, 13 use literals, mocked seams, or temp-root changelogs, 2 read the repository CHANGELOG, and only this one bounded its assertion to the Unreleased section. `standing_residual_floor_3448.test.ts` is the other repository reader but is not section-bounded, so it is untouched here and still passes.

## Gate integrity (#3156)

This edits a currently-failing gate, which is normally forbidden. It is authorized because the arc bound this specific repair and the assertion encoded a false invariant. The guarantee is preserved, not relaxed: still exactly one, still both tokens.

## Evidence

Mutation proofs against the repository CHANGELOG, each mutation applied by index and reverted (changelog restored byte-identical):

| Mutation | Expected | Observed |
|---|---|---|
| Entry line removed | fail | fail -- `expected [] to have a length of 1 but got +0` |
| Entry line duplicated | fail | fail -- `expected [ ...(2) ] to have a length of 1 but got 2` |
| `occupancy:grant` dropped from the entry | fail | fail -- length 0 |
| `Closes #3755` dropped from the entry | fail | fail -- length 0 |
| Synthetic pre-cut: entry moved under `[Unreleased]` | pass | pass |
| Synthetic post-cut: two later release sections inserted above the entry | pass | pass |

The last two demonstrate section-independence against synthetic changelogs rather than by assertion: a future cut that moves the entry again cannot break the case.

## Review fix batch

Greptile raised one P1 (inline, not counted in the rolling summary): the scope xBRIEF was tracked under `xbrief/active/` with `plan.status: running`, and its only `github-issue` reference is #3930 -- which this PR closes on merge. That is exactly the `verify:orphan-active` shipped signature, so master would have gone red again on a different gate the moment the issue closed.

The finding is valid. `collectGithubRefs` (`packages/core/src/orphan-active/refs.ts`) collects only `github-issue`-typed references, so the `x-xbrief/refs` entries for #3755 and #3156 are not collected and the brief resolves to #3930 alone; the unscoped sweep then returns `shipped("all referenced issues are closed")`.

Verified against the real evaluator with a stubbed post-merge GitHub state (#3930 closed, #3785 open), on two synthetic roots that both actually run the sweep:

| Tree | `verify:orphan-active` |
|---|---|
| WITH the #3930 active brief (what the first push shipped) | **exit 1** -- "1 active/running xBRIEF look shipped but still consume WIP" |
| WITHOUT it (what this branch ships) | **exit 0** -- "no orphaned active/running xBRIEFs (scanned 1 running brief in active/)" |

Fix: the brief is no longer tracked in this PR. It stays `running` locally so the worker keeps its lifecycle, and lands as a stamped `completed/` artifact after merge via `scope:complete` -- which #3476 requires regardless. Master therefore has no orphan exposure at any point, rather than a red window that a follow-up lifecycle PR would have had to close.

## Related Issues

Closes #3930. Refs #3755, #3156, #3448.

## Checklist

- [x] `/deft:change <name>` — N/A (2 files, not cross-cutting)
- [x] `ROADMAP.md` — N/A
- [x] Tests pass locally — full `task check` green

## Post-Merge

- [ ] **Verify issue auto-close**: after squash merge, confirm #3930 closed; #3755 must stay closed and must not be re-referenced by a closing keyword.
- [ ] `scope:complete` + land the completed xBRIEF (`verify:orphan-active` / `verify:completed-tracked` both exit 0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- **A content contract no longer pins its changelog line to the `[Unreleased]` section (#3930).** A release cut moves an entry into the section it cut, so the v0.108.0 cut emptied the bounded slice and left `task check` red on master for every branch, independent of what that branch changed. The occupancy membership case now requires exactly one entry anywhere in the changelog carrying both of its tokens. The changelog is append-only across cuts, so that holds for every future release with no section parsing and no new release machinery. Closes #3930. Refs #3755, #3156.
 
 ### Removed
 
--- a/packages/core/src/content-contracts/standards/occupancy_membership_docs.test.ts
+++ b/packages/core/src/content-contracts/standards/occupancy_membership_docs.test.ts
@@ -36,11 +36,15 @@ describe("occupancy boundary naming (#3755)", () => {
     expect(text).not.toContain("Join (`occupancy:request`) is named in remediation only");
   });
 
-  it("carries one Unreleased changelog line for the membership change", () => {
+  // Searched over the whole file, not the Unreleased section: a release cut
+  // moves the entry into the cut section, so a section-bounded search is a
+  // scheduled failure. The changelog is append-only, so exactly-one holds
+  // wherever the entry lives (#3930).
+  it("carries exactly one changelog entry for the membership change", () => {
     const changelog = readText("CHANGELOG.md");
-    const unreleased = changelog.split("## [Unreleased]")[1]?.split("\n## ")[0] ?? "";
-    const mentions = unreleased.split("\n").filter((line) => line.includes("Closes #3755"));
-    expect(mentions).toHaveLength(1);
-    expect(mentions[0]).toContain("occupancy:grant");
+    const entries = changelog
+      .split("\n")
+      .filter((line) => line.includes("Closes #3755") && line.includes("occupancy:grant"));
+    expect(entries).toHaveLength(1);
   });
 });


MScottAdams:
Valid, and fixed in `26ea3e84`.

Confirmed the mechanism rather than assuming it: `collectGithubRefs` (`packages/core/src/orphan-active/refs.ts`) only collects references whose type matches `github-issue`, so the `x-xbrief/refs` entries for #3755 and #3156 are not collected and the brief resolves to #3930 alone. With that issue closed on merge, the unscoped sweep returns `shipped("all referenced issues are closed")` and exits 1 -- master red again on a different gate, which is precisely what this PR exists to prevent.

Verified against the real evaluator with a stubbed post-merge GitHub state (#3930 closed, #3785 open), on two synthetic roots that both genuinely run the sweep:

- WITH the #3930 active brief: exit 1 -- "1 active/running xBRIEF look shipped but still consume WIP"
- WITHOUT it: exit 0 -- "no orphaned active/running xBRIEFs (scanned 1 running brief in active/)"

Fix: the brief is no longer tracked in this PR. It stays `running` locally so the worker keeps its lifecycle, and lands as a stamped `completed/` artifact after merge via `scope:complete`, which #3476 requires regardless. That removes the orphan window entirely instead of leaving one for a follow-up lifecycle PR to close.

Issue #3463, opened 2026-08-18
feat(languages,doctor): Rust consumer lane — executable rust.md, Cargo-aware doctor, check recipe


MScottAdams:
## Summary

One issue for a **Rust consumer lane**: people who already write Rust and consume Directive. Guidance exists (`content/languages/rust.md`). The operating lane does not. TypeScript gets copy-paste configs, a doctor check, and a check contract agents can run. Rust gets prose and "see `commands.md`."

This is **not** "rewrite Directive in Rust" and **not** `#3461` (prefer Rust-backed *tools* while keeping a JS/Python app). If the app *is* Rust, `task check` should run Cargo.

**Grade:** immediately actionable docs + thin doctor/setup. No engine rewrite.

## Current gap (verified in-tree)

- `languages/rust.md` is standards-complete (rustfmt, clippy, rustdoc, thiserror/anyhow, unsafe, cargo-audit, ≥80% coverage) and content-contracts keep its shape.
- Setup lists Rust and *looks* at `Cargo.toml` during inference, then treats it as just another language name.
- `task doctor` has no `rustc` / `cargo` / `rustfmt` / `clippy` probe.
- After `directive init`, include-only `Taskfile.yml` + `task deft:check` does **not** run `cargo test` or clippy (`docs/consumer-check-contract.md` / #3218).
- `content/templates/` has no Cargo / rustfmt / clippy scaffold.
- `REFERENCES.md` does not name `languages/rust.md`.
- `coding/testing.md` has no Rust subsection.
- Coverage floor in `rust.md` is **80%**; Go/TS/JS files say **85%**; USER.md cannot weaken the project bar.
- `rust.md` see-also still points at legacy `PROJECT.md`.
- This repo dogfoods TS + Go, not Rust. Quality will come from the language file + gates, not from us living in Cargo.

### A — Make `rust.md` agent-executable

Add drop-ins agents can copy, matching `languages/typescript.md`:

- `Cargo.toml` bin+lib / workspace skeleton
- `rustfmt.toml` (100-col, already required in prose)
- clippy config matching `deny(clippy::all)` / pedantic-for-libs
- Taskfile fragment: `fmt` → `cargo fmt --check`, `lint` → `cargo clippy -- -D warnings`, `test` → `cargo test`, `test:coverage` → llvm-cov or tarpaulin, `audit` → `cargo audit`, `check` composing those
[...]

Issue #3961, opened 2026-08-30
bug(swarm,harness): grok-build parent poll has no fail-closed host-budget stop, so overnight babysit OOMs the CLI


MScottAdams:
> [!IMPORTANT]
> **The remedy specified in this issue was refuted by its own accepted design-critique arc (record 5470766807).** A three-critic panel found, on independent grounds, that a deterministic per-poll host-budget verb run by the parent cannot bound the failure it targets. The Ask and Acceptance sections below have been retired in place; their original wording is preserved in this issue's edit history.
>
> **Do not ingest this issue as a build instruction.** The measured problem statement and the gate-coverage table below still stand and are worth reading.
>
> Successors: #3983 and #3984, both dispositioned by the set-level arc anchored on #3983 (record 5470998520), and #3987, a hook matcher-coverage defect found while checking.

## Problem

The official Grok Build swarm monitor path is a live parent poll (`get_command_or_subagent_output` — host adapter #1331 / #1342). Dual-stop (#2442), parent-turn-shape (#3131 / FC14), delivery-attempt (#3143), envelope SLA (#3153), and effort-budget surfacing (#3266) do not fail-close that poll when the **host process** is the resource being consumed.

Measured dogfood crash (2026-08-30, win32): `grok.exe` PID 51484 aborted with Rust `memory allocation of 1101824 bytes failed` and Windows exit `0xc0000409` (abort / `__fastfail`, not a stack smash). Last log was `shell.turn.inference_start` at `2026-08-30T09:14:38Z` (4:14 AM CDT) on session `01a05086-…`, **loop 283**, after `get_command_or_subagent_output`. That turn had been running ~6.6 hours. The same session's last successful prompts were **478k–480k tokens**. Peak across the process was **569,282 tokens**. Thirteen sessions lived in one process. Grok memtrace on this host records jemalloc purge as `hook_installed: false` (Unix-only path), so retained pages are not returned.

A 1.1 MB alloc fails only when the process is already at commit limit. The prescribed poll loop filled the heap; Directive had no machine gate that could stop it.

## Why existing gates miss

| Surface | What it stops | Why it did not stop this |
| --- | --- | --- |
| Dual stop #2442 | Skill-prose iteration / no-progress halt | Not a `task` on the grok-build poll path; parent continued to loop 283 |
| `evaluateParentTurnShape` #3131 | Repeated assistant text with no tools (FC14) | Parent was issuing tools (`get_command_or_subagent_output` / `monitor`) |
[...]

PR #3972, opened 2026-08-30
fix(release): classify Closes position and require privileged census labels

### LINKED ISSUE #3969: bug(release): the consumer hard-stop gate has no ships-past disposition, so its only escape hatch is a false Closes claim
## Summary

`verify:consumer-hard-stops` (#3900, landed today at `a4f70040`) offers a cut exactly two dispositions for an open consumer hard-stop: fix it, or declare `Closes #N` in the CHANGELOG `[Unreleased]` section. There is no way to record "this cut knowingly ships past #N, and here is why."

That missing third disposition is not hypothetical. It is the decision this project already made once, in writing, and it is blocking a cut right now.

## Measured on first real use

Running the gate against the current tip:

```
Consumer hard-stops: fail -- 6 open hard-stop(s) not in this cut Closes set:
  #3463, #3602, #3612, #3614, #3615, #3737
Recovery: close those issues in this cut (or list them in the Unreleased Closes set).
```

`parseClosesSet` (`packages/core/src/release/consumer-hard-stops.ts`) reads only `Closes|Fixes|Resolves #N` from the text between `## [Unreleased]` and the next `## [` heading. Nothing else clears an entry.

## Why the existing recovery line is not a disposition

The operator's decision for this cut is to fix four of those and **ship past #3463** (a Rust consumer lane — a feature, not a defect fix). Taking the printed recovery literally would mean writing `Closes #3463` in the CHANGELOG. That is false on its face, and it has a second-order effect: GitHub's closing-keyword parser would **auto-close #3463** when the release PR merges. That parser ignores negation, which was measured on this repository today — a lifecycle PR body reading "Does not close #3899" closed #3899 anyway (#701).

So the only available escape from the gate is to assert something untrue and silently close a real open issue. A gate whose sole escape hatch is a false statement pressures every future operator toward exactly the misrepresentation #3156 exists to prevent.

## The precedent this should encode

On 2026-08-28 the operator recorded a risk acceptance on #3899 in prose: the pending cut shipped as-is past three known consumer BLOCKERs, with a stated argument (delaying 21 unrelated fixes cost more than the marginal exposure, given every finding already reproduced on the published 0.107.0), a named affected path (first-session onboarding), a named remediation target (the release after), and an explicit "this is not precedent" clause.

That is a well-formed disposition. It was invisible to any machine. The gate landed the following day and cannot read it.

## Proposed direction (for the arc to weigh, not settled)

A third declaration in the `[Unreleased]` section, parsed separately from `Closes`, that satisfies the census while recording the acceptance — carrying at minimum the issue number and a reason, and printed in the gate output rather than silently absorbed, so the cut's own release notes state what it ships past.

1. Whether the disposition belongs in the CHANGELOG at all, or in a decision record (`deft decision:write`) that the gate reads. The CHANGELOG is consumer-facing release notes; a risk acceptance may be exactly what belongs there, or may be operator metadata that does not.
2. Whether it should require authority — a waiver that any agent can write is not a control. #3156 governs who may clear a red gate, and the whole point of this gate is that an agent cannot wave it through.
3. Whether it should expire. The #3899 acceptance named "the release after" as a remediation target; nothing enforced that, and the issues are still open two cuts later.
4. Whether `adoption-blocker` and `BLOCKER`-titled entries deserve the same disposition path. #3713 makes the label privileged and body-evidence gated, so waiving one may warrant more than waiving the other.
5. Whether the printed recovery line should stop advertising the `Closes` route for issues the cut does not actually close.

## Acceptance

1. A cut can record that it ships past a named open hard-stop, with a reason, without asserting that it closes it.
2. The declaration does not trigger GitHub's closing-keyword parser.
3. The gate prints what the cut ships past, rather than passing silently.
4. Whatever authority rule is decided is enforced by the gate, not by convention.
5. A test proves the disposition is not vacuous: a cut that ships past an entry without a well-formed declaration still fails.

## Design-critique

`mechanism-shaped: true`

Refs #3900, #3713, #3156, #3899, #701, #3463.

MScottAdams:
## Summary

Bound SoT is successor lean 5470479357. The issue body claimed the consumer hard-stop gate only escaped via a false closing-keyword claim. That is false: parseClosesSet had no position classification.

This PR:
1. Composes citation-grammar classifyPosition into parseClosesSet so a closing keyword inside a fence, inline code span, blockquote, strikethrough, or explicit negation does not clear the census. A plain control still clears. Tests for each of the five bypass forms failed before the change.
2. Makes census blocking privileged labels only (adoption-blocker, blocks-release-tag). Title-only BLOCKER is an inbound flare (#3713), not a release gate. A blocks-release-tag entry like #3899 becomes visible.
3. Does not ship a ships-past hatch. Authority would have to be a flag carrying an issue citation; the critic steelman for refusing the hatch is the recorded outcome. Whether the tag waits stays the operator decision.
4. Wires task pr:check-closing-keywords into branch-gate.yml (including PR body edits, FP mode) and into the framework check graph via verify:closing-keywords.

Issues 3463 and 3899 stay open. This PR does not claim to unblock the pending tag.

## Related Issues

Closes #3969

## Checklist

- [x] /deft:change N/A -- bound design-critique story with active xBRIEF; no separate change proposal
- [x] CHANGELOG.md -- added entry under [Unreleased]
- [x] ROADMAP.md N/A
- [x] Tests pass locally (targeted vitest + verify:closing-keywords)

## Post-Merge

- [ ] Verify issue auto-close for #3969 after squash
- [ ] Enable branch protection N/A


MScottAdams:
Batch 4 (class A leftover, no live git fetch on every task check):

Greptile asked for a live git fetch so a stale origin/master cannot enlarge merge-base..HEAD. Putting the network on every task check would be a different product than the rest of the origin-based check graph, and we will not weaken the gate to go green (#3156).

Third option shipped on 968626f5: detect the locally observable stale-range enlargement without fetching. If origin/<default> is an ancestor of local <default> AND HEAD merge-bases differ, verify:closing-keywords now returns stale-base and exits 2 with git fetch origin <default> recovery. If merge-bases match, the origin range is still used. Missing origin merge-base remains fail-closed (exit 2). CI still prefers --pr via GITHUB_PR_NUMBER / PR_NUMBER.

This does not fetch. It fails closed on the evidence this checkout can see. GitHub-unknown origin HEAD remains the same contract as every other origin-based gate; the merge-gate path is --pr and does not use the local range.

If that still reads as insufficient, we will refuse a live fetch on the whole check graph rather than add one.

MScottAdams:
Refusal (fetch leftover, HEAD 32172870):

Greptile still wants a live git fetch so a stale origin/master cannot enlarge merge-base..HEAD when no newer local default exists to reveal the lag.

Judged false-positive as a merge-blocking finding, not as a description of git:

1. GitHub-unknown origin HEAD is locally undecidable. A current origin and an origin that is equally stale with local default are the same graph. Distinguishing them requires the network.
2. Scoping fetch to this wrapper is still network on every task check, because verify:closing-keywords is in the framework check graph. That is a different product than the rest of the origin-based gates. We will not add it, and we will not weaken the gate to go green (#3156).
3. Missing origin already fails closed (exit 2, fetch recovery). Locally observable lag (origin ancestor of a descendant local default, HEAD merge-bases differ) already fails closed (stale-base, fetch recovery). CI / merge-gate uses --pr and never takes this local range.
4. When origin is stale and we cannot see it, the range is too large, so the local check is noisier (false fail on already-merged keywords), not weaker. Recovery is git fetch origin master -- the same recovery the missing-base path already prints.

P3 ships-past hatch refusal stands. No live fetch on the check graph.

MScottAdams:
Batch 6 (class A leftover, scoped fetch):

Detect-staleness was not enough for Greptile after a rebase re-review. Remaining hole is GitHub-unknown origin HEAD when no newer local default exists.

Third option the operator named: fetch is scoped to this wrapper only, not the whole check graph. resolveClosingKeywordsSource now runs git fetch origin <default> -- before merge-base on the local path. CI still uses --pr and does not fetch. If fetch fails, missing-base and locally observable stale-base still fail closed.

Does not put fetch on other origin-based gates. Does not weaken the gate (#3156). P3 hatch refusal stands.

MScottAdams:
Ping-pong halt on the fetch leftover (HEAD 5760a027):

Greptile is now asking for the opposite of the leftover that blocked merge two reviews ago.

1. Detect-staleness without fetch: 4/5, leftover "stale range still accepted".
2. Scoped fetch in this wrapper: 4/5, leftover "when fetch fails you still use a stale origin".
3. Fail closed when fetch fails and freshness is unproven: CLEAN 5/5 P0=0 P1=0 on 680e82a6.
4. xBRIEF evidence stamp only (#3781 closeout-attestable): 3/5, leftover "when fetch fails you discarded a valid cached merge-base / live fetch blocks local checks".

Those two leftovers cannot both be true. Using a cached origin after fetch fail is the stale-range finding. Discarding it is the offline-hard-fail finding. CI still uses --pr and never takes this path.

I will not reverse the 5/5 product to chase the new opposite leftover. That is the thrash the dual-stop forbids. P3 hatch refusal stands. Closeout evidence is stamped; task verify:pr-closeout-attestable -- --pr 3972 passed locally.

MScottAdams:
Declined leftover (operator accept-5/5-and-merge):

The product reached CLEAN 5/5 on 680e82a6 (P0=0, P1=0, no has_blocking). The only delta on the current HEAD (5760a027) is the #3781 acceptance-evidence stamp on the active xBRIEF -- no product change.

The leftover posted against that stamp asks for the opposite of the finding that produced the CLEAN: keep the cached origin merge-base when a freshness fetch fails. Using the cached origin after fetch-fail is the stale-range finding. Discarding it is the offline hard-fail finding. Both cannot be satisfied in this PR.

Fail-closed on an unproven origin after a failed fetch stays. The design question is not dropped: it is filed as #3988 (when a freshness fetch fails, should an origin-comparing gate fail closed or fall back to the cached ref) for its own design-critique arc.

P3 ships-past hatch refusal stands. Does not close #3463, #3899, #3984, or #3961. Does not change the confidence floor, verifier, required checks, coverage, or policy (#3156).

Issue #3984, opened 2026-08-30
rfc(swarm,process): the parent-must-not-own-a-multi-hour-babysit rule did not bind, and another prose restatement will not bind either


MScottAdams:
## Summary

Split from the #3961 arc (synthesis accepted; lean `5470761146`, table `5470765129`). That arc refuted the remedy #3961 specified and named this as the second of two surviving Directive-side items. #3961 is not re-opened by this issue and stays as filed.

The claim here: the overnight crash was not a monitor loop consuming a host. It was **a parent implementing inside its own babysit for hours**, which existing rules already forbid — so the gap is that those rules were not binding on the actor, not that a rule is missing.

## What the arc measured

- **The fatal turn was working, not polling.** In its retained window, **54% of tool calls were terminal commands and 24 were direct file edits**. That is the prohibited shape of a parent implementing as the leaf, not a monitor loop.
- **A mandated per-poll three-state verb of exactly the requested shape already exists on this path**, and a bounded wait with a thirty-minute cap already ships.
- **An existing rule already forbids the loop that ran.** #3032 / #1880 Gap C require the parent to dispatch rather than babysit product loops when worker dispatch is available.
- The turn had been running **~6.6 hours** at the point the process aborted, across a process holding thirteen sessions.

So the honest description is an **enforcement-locus and ownership** problem: the rules existed, were mandatory, and did not bind.

## Why another prose rule is the wrong answer

#3961's own framing invokes thin fail-closed design (#3265): one check plus one remediation, because more AGENTS.md sentences will not catch the next overnight parent. That argument applies with equal force to this item. A restatement of "the parent must not babysit" adds a sentence to a document that already contains the sentence.

What would make the rule bind is the open question, and it is a genuine one — the arc did not answer it.

## Not asserted -- for the critic and builder to establish

1. Whether an envelope rule can be made machine-checkable at all. `evaluateParentTurnShape` (#3131 / FC14) already inspects parent turn shape but keys on repeated text with no tool calls; a parent issuing many terminal commands and file edits is the opposite signal and passes.
2. Whether "a parent is implementing rather than orchestrating" is detectable from turn shape — for example a ratio of edit and shell calls to dispatch calls, sustained over a window — or whether that misfires on legitimate parent work such as writing arc records, filing issues, and reading state. **This one matters**: a parent that maintains design-critique records and files scopes issues many writes per turn while orchestrating correctly.
[...]

Issue #3614, opened 2026-08-21
BLOCKER: bug(codex): proposed-scope exemption cannot extract target from raw apply_patch payload


dbcall2:
## Summary

The active-scope planning exemption for `xbrief/proposed/*.xbrief.json` does not work for Codex's `apply_patch` tool when the hook payload contains the raw patch but no separate `path` field. A single-file proposal addition is denied for having no active xBRIEF even though the denial message advertises the proposed-scope exemption.

This is a Codex residual/regression after #2625 and #2738: master supports the exemption when `hookWriteTargetPath()` receives `file_path`, `filePath`, or `path`, and the ApplyPatch regression test supplies `tool_input.path`. The Codex tool surface observed here supplies the patch document itself, whose `*** Add File:` target is not extracted.

## Environment

- Consumer: `3Ci-Consulting/requirementor`
- Host: Codex desktop on macOS, `apply_patch` direct-write tool
- Installed Directive/deposit: `0.105.0`
- Latest Directive `master` checked before filing: `2679d9baa44e2794b5492b32c5655c5506342c29`
- Mutation ritual: green (`deft session:start --rearm`; gated session ritual passed)
- Initial lifecycle state: no active normalization xBRIEF yet; the attempted write was the proposal intended to create it

## Reproduction

1. In a consumer repository with no applicable active xBRIEF, invoke Codex `apply_patch` with one patch that adds a valid file such as:

   ```text
   *** Begin Patch
   *** Add File: xbrief/proposed/2026-08-21-normalize-requirementor-onboarding.xbrief.json
   +{ ... valid proposed xBRIEF ... }
   *** End Patch
   ```

2. The hook denies the write:

   ```text
   Directive denied apply_patch: No active xBRIEF ...
   Write a new proposal to xbrief/proposed/*.xbrief.json (planning exemption)
   ```

3. Repeat with an absolute target in the `*** Add File:` header. It is denied identically.
4. Run the sanctioned CLI workaround, `deft xbrief:create -- --out xbrief/proposed/...`. Creation succeeds, after which normal promote/activate works.

The two rejected `apply_patch` attempts made no file changes.

## Latest-master analysis

At `master` commit `2679d9b`:

- `packages/core/src/hooks/dispatcher.ts` correctly exempts a proposed lifecycle artifact once `hookWriteTargetPath()` returns its path.
[...]

Issue #3186, opened 2026-08-07
sec(authz,hooks): AppSec scan @ 4bf54335 — UAT Shell fail-open (authz obfuscation, kill-switch/policy, assist-scratch symlink)


MScottAdams:
## Summary

AppSec scan of **deftai/directive @ `4bf54335`** produced **3 validated findings** (2 HIGH, 1 MEDIUM). All are **UAT / PreToolUse residual bypass** class: #3110 closed literal `.deft/authz/` Shell writes, but **unclassifiable** Shell and **lexical** assist-scratch still fail open.

**Impact theme:** Under active UAT, an agent can still clear or forge authz state, plant the Directive kill-switch, flip branch/merge policy, or escape the checkout via a symlink under the assist scratch carve-out — unlocking Write / push / PR / merge without a human grant.

---

### 1. HIGH — UAT residual: obfuscated Shell writes to `.deft/authz/` still fail-open

| | |
|--|--|
| **Severity** | HIGH |
| **Location** | `packages/core/src/authz/classify.ts` |
| **Commit** | `4bf54335` |

**Impact:** After #3110, literal authz store writes are denied, but `python3` / `node` (and similar) **byte/base64 path construction** returns `classifyShellAuthzOps=[]` → **`shell-op-unclassifiable` allow**. An agent can:

- Clear UAT state or forge operator-cli grants
- Empty `activeGrantIds` then activate grants → Write / push / PR / merge unlock

**Remediation:**

- Under active UAT, **fail-closed** on unclassifiable Shell that uses python / node / perl (or other) write bins; **or**
- Require opaque FS writes to **resolve realpath** under the project and treat destinations under `.deft/authz` as **settings** (same deny posture as #3110 literals)

**Tests:** Add fixtures for base64/byte-constructed paths to `.deft/authz/*` under UAT; expect deny, not unclassifiable allow.

---

### 2. HIGH — UAT residual: Shell can plant `.deft-directive-disable` / flip policy settings

| | |
|--|--|
| **Severity** | HIGH |
| **Location** | `packages/core/src/authz/classify.ts` |
| **Commit** | `4bf54335` |

**Impact:** Commands such as:

- `echo > .deft-directive-disable`
- `deft policy:allow-bot-merge` / `allow-direct-commits` / `disable-directive` (and peers)

still **classify empty** → Shell **allowed under UAT**.

Consequences:

[...]

PR #3990, opened 2026-08-30
fix(hooks): close shell write reissue and evaluate in-flight elapsed

### LINKED ISSUE #3983: rfc(harness,swarm): host-budget enforcement belongs at a pre-tool chokepoint reading a non-forgeable source, not at a cooperative parent poll
## Summary

Split from the #3961 arc (synthesis accepted; lean `5470761146`, table `5470765129`). That arc refuted the remedy #3961 specified — a per-poll host-budget verb the parent runs — on three independent grounds, and named this as one of two surviving Directive-side items. #3961 is not re-opened by this issue and stays as filed.

The surviving claim is about **where enforcement sits**, not about adding another check the parent must remember to run.

## What the arc established

Measured by three siblings across three model families, against the incident artifacts:

- **The proposed oracle carries no memory value.** One seat inventoried **7,066 sample records** holding only a timestamp and a kind. Every threshold built on it is a workload proxy, not a memory measurement.
- **The quantity the check wants is never written at the moment it would run.** Prompt size is recorded when a turn *finishes*. The process died *starting* a turn, so a check running before the next inference is always one turn stale, and the fatal allocation is precisely the one the log never captured.
- **Most of the footprint is unreachable from a poll-cycle check.** Summing per-session context in the crashed process, **84% of retained tokens sat in idle sessions**. A parent that halts, compacts or splits its own session removes at most about a sixth of the footprint, and idle sessions run no poll cycle for a per-poll check to gate.
- **Enforcement is cooperative, and that is the defect.** The mid-run enforcer would be invoked by the same actor that had already ignored the ownership rules covering that loop. A verb the parent must remember to run and choose to honour inherits the failure it is meant to prevent.
- **The oracle is attacker-writable shared state.** It is one append-only file owned by the account every agent on the host runs as, with **no provenance field**. A check keyed on it can be silenced by appending benign records, or turned into a cohort-wide denial of service by appending threshold-tripping ones.
- **A host pre-tool chokepoint already exists on this host** and was absent from #3961's candidate list. One seat identified it as already deposited.

## The claim this issue makes

If a resource gate is wanted at all, it belongs at a chokepoint that can **deny the call**, reading a **non-forgeable process-level source**, rather than at a parent that is asked to stop and reads a file any agent can write.

That is a different mechanism from the one #3961 asked for, with different inputs, a different trust model, and a different failure mode. It deserves its own critique rather than inheriting a refuted design's framing.

1. Whether the already-deposited host pre-tool chokepoint can read a process-level memory figure at all on this platform, or whether it is limited to the same proxies that failed. If it cannot, this issue may have no buildable core and should say so rather than shipping proxies at a new address.
2. What the non-forgeable source actually is — RSS, commit charge, allocator statistics — and whether it is available per-process on win32 without an upstream change. Note #3961's non-goals already fence host memory retention as upstream.
3. Whether denying a tool call is a safe failure mode. A chokepoint that denies mid-cohort could strand workers holding leases or half-written state; the arc found the cooperative design unsafe, and a coercive one has its own blast radius.
4. Whether this belongs to Directive at all, or is a host feature Directive should merely require. Shipping a Directive-side enforcer for a host resource problem may be the same category error #3961 made, moved one layer down.
5. What a threshold would be fitted to. The arc found #3961's proposed conjunction fires exactly once across 44 sessions, with two of its three conjuncts already firing on healthy processes — a curve fit through one point. Any threshold here needs a defensible derivation, not a reconstruction of one incident.
6. Whether the existing bounded-wait mechanism and worker dispatch already bound this in practice, making a resource gate unnecessary rather than merely misplaced.

## Acceptance

To be set by the arc. This issue states a premise and an enforcement-locus claim; it does not prescribe an implementation, and a critic concluding that no Directive-side gate should ship is a legitimate outcome.

## Design-critique

`mechanism-shaped: true`

Refs #3961, #3265, #3153, #3131, #3143, #1331, #1342.

MScottAdams:
## Summary

Bound remedy from the #3983/#3984 set-level arc. Not the #3983 issue body.

Part 1 (security): PreToolUse matchers now include Grok Build spellings (run_terminal_command, write, search_replace). Recognized in-repo shell file-write dests (Set-Content, python pathlib, IO.WriteAllText) are authorized through inspectMutationGates the way ApplyPatch authorizes every mutation target (#3614). A Write occupancy deny can no longer be reissued through the shell. git status, occupancy:release, and OS-temp body-file writes stay allowed. Absolute dests are resolved before containment so a /tmp/../ prefix cannot skip occupancy. Relative dest after cd to OS temp stays fail-closed (reconstructing cwd reopens occupancy).

The reissue-path test is packages/core/src/hooks/dispatcher-write-reissue.test.ts. The /tmp/../ occupancy test is in the same file.

Part 2: evaluateInFlight applies maxElapsedSeconds to a queued/running attempt. Pre-dispatch cannot see that case (DENY_DUPLICATE_ACTIVE).

Did not build a chokepoint that reads host memory logs. Did not build a parent-is-implementing detector. Left #3984 and #3961 open. Left #3987 open: matcher coverage is necessary and not sufficient for that issue liveness half.

## Related Issues

Refs #3987
Closes #3983

## Checklist

- [x] /deft:change N/A -- bound design-critique story with an activated scope xBRIEF
- [x] CHANGELOG.md -- Unreleased Fixed entry
- [x] ROADMAP.md N/A -- not a tracked ROADMAP item
- [x] Tests pass locally -- reissue path, matcher catalog, evaluateInFlight, dest-form, occupancy:release pin, canonicalize /tmp/../

## Post-Merge

- [ ] Verify issue auto-close for #3983 only
- [ ] Confirm #3987 remains open
- [ ] N/A branch protection


MScottAdams:
Halt: dual-stop after 3 Greptile batches. Not merge-ready.

HEAD 6333a8ab. Confidence 2/5 (floor 5). Mechanical mergeable_state=clean is not sufficient.

What landed in the PR:
- Part 1: Grok shell matcher coverage plus reissue-path tests (dispatcher-write-reissue.test.ts). Occupancy deny on Write also denies Set-Content and pathlib write_text to the same in-repo dest. git status, occupancy:release, and absolute/OS-temp dests stay allowed.
- Part 2: evaluateInFlight on the delivery-attempt ledger.

What was tried and reverted: a command-wide last-cd-temp exemption. Greptile showed it skipped occupancy when a write ran before a later temp cd. Unsafe. Reverted.

Outstanding leftover (class A, already-touched files):
1. Relative dest after cd to OS temp is fail-closed as unprovable. Reconstructing cwd reopens occupancy bypass. Documented Windows body-file workflow uses absolute temp or env TEMP paths, which still allow.
2. looksLikeTempPath is lexical and runs before resolve, so a dest containing /tmp/../ can skip mutation gates. Canonicalize first; do not reconstruct cwd.

Did not close 3983 (PR not merged). Left 3984 and 3961 open.

RESUME: residual=conf_floor leftover=A conf=2/5 floor=5 standing=one-shot PR=https://github.com/deftai/directive/pull/3990

MScottAdams:
## Halted by operator decision — the open findings are a design question, not review residue

This PR is **not merged and not abandoned**. Its review loop is stopped deliberately, at eight batches across two dispatches, because the findings stopped being about this diff.

## Why

The two most recent findings are the same defect in different clothes:

- *changed cwd bypasses write gates* — compound destinations discard working-directory context
- *path transforms misattribute write targets* — the scanner accepts arbitrary invoked methods before the write call, but records only the initial `Path` argument

Both reduce to: **the destination is not statically recoverable from the command text**. So do `cd`, `pushd`, variable expansion, symlinks and method chains. Earlier batches in this same PR closed compound relative destinations, a trailing `cd $env:TEMP` exemption that was itself an occupancy bypass, and a lexical temp-path check running before canonicalization.

That is not a queue of bugs. It is one posture, expressed once per batch:

```ts
if (!isInRepoShellWritePath(projectRoot, dest.path)) continue;
```

If a destination cannot be shown to be in-repo, the gate is skipped. The residual class is every command whose destination is not statically recoverable, and that space is unbounded — so a reviewer can keep drawing from it indefinitely and the confidence score will not trend to the floor. It went 2 → 3 → 4 → 3 → 2 → 1/5.

Continuing would let review set the design by attrition, one obfuscation at a time, which is what a design-critique arc exists to prevent.

## Where the question went

Filed as **#3997** — whether the gate should invert to fail-closed on destinations it cannot prove external. That issue carries this PR's batch history as its evidence, states the ergonomic cost honestly (fail-closed would deny legitimate computed-path writes, including the mandated `--body-file` workflow this repository requires on Windows), and records that a critic concluding "fail-open is the correct trade, land #3990 as written" is a legitimate outcome.

## What stands regardless

The work on this branch is real and is not being discarded:

- the host shell tool is in the shell matcher, closing the reissue bypass where a write denied through the edit tool succeeded through the shell
- recognized in-repo destinations route through the same mutation gates as the edit tool
- a reissue-path test denies a `Write` and then denies the identical path reissued through `run_terminal_command`
- the in-flight elapsed evaluator on the delivery-attempt ledger

Whether that lands as-is, lands with the posture inverted, or waits, is #3997's to decide.

## Refusals preserved from the review cycle

The worker refused, and those refusals stand: it did not reinstate the `cd $env:TEMP` exemption to buy a confidence point after measuring it as an occupancy bypass; it did not reconstruct cwd to relax the compound-destination denial; and it did not merge below the dogfood floor. None of the halt is a gate being weakened.

This PR is also **not** in the pending release. That was decided separately, on the grounds that shipping a partially-effective security fix retires the concern while leaving the surface open.

MScottAdams:
Refuse Greptile findings on HEAD 102d3ae7. They are the #3997-settled fail-open class, not this dispatch's defect.

What this dispatch fixed: extractAfterVerb treated -Value as the dest when it appeared before -Path. Tests that failed before the change:

- Set-Content -Value x -Path src/app.ts now extracts src/app.ts
- Set-Content -Value x with no -Path yields no dest (does not fabricate -Value)
- Set-Content -Path src/app.ts -Value x still extracts src/app.ts
- occupancy deny on the -Value-first form

What this dispatch refuses (same fingerprint as prior rounds; operator bound on #3997):

- Directory-change / cwd reconstruction. Compound dests stay fail-closed as unprovable. Reconstructing cwd reopens occupancy bypass. The previous worker already refused this.
- Pathlib transform chains that emit the original Path() argument. Recogniser recall is #3987, out of scope here.
- Inverting isInRepoShellWritePath to fail-closed on unprovable dests. #3997 synthesis 5472062522: fail-open at this predicate stays.

#3186 realpath containment on isInRepoShellWritePath is deferred: assertProjectionContained is escape-polarity (in-tree dest that realpaths out). The critic junction is re-entry (lexical out, realpath in) and needs a dest-side ancestor walk that is not that binding.

Did not reinstate the cd TEMP exemption.

Confidence 1/5 is below dogfood floor 5. P0=0 P1=0 on the current HEAD probe. Will not merge below the floor. Will not spend a second code batch on the settled posture.

Refs #3997. Refs #3987. Closes nothing.

MScottAdams:
Operator-authorized landing below the dogfood confidence floor (#3095). This is a this-PR-only override, not a gate change (#3156): minGreptileConfidence, the verifier, and policy flags are unchanged.

plan.policy.requireHumanMerge is false. The operator authorized merge of HEAD e167682b6e9036f621175858c8aca3fc61d711c0 at Greptile confidence 1/5.

Recorded at the last SHA-matched watch on the prior product head: P0=0, P1=0. The holdout was confidence, not findings.

The four refused findings stay out of scope by the bound #3997 arc (synthesis 5472062522, table 5472059705). Full refusal: https://github.com/deftai/directive/pull/3990#issuecomment-5472281394

1. Cwd reconstruction and compound-destination fail-open -- the posture the #3997 panel refuted 3-0.
2. Inverting isInRepoShellWritePath -- that same refuted proposal.
3. Pathlib transform misattribution -- recogniser recall at #3987's layer, not this dest predicate.
4. The cd $env:TEMP exemption -- measured as an occupancy bypass; not reinstated.

Scope statement: content/contracts/path-write-fence.md section Shell file-write reissue (#3983 / #3987), plus CHANGELOG [Unreleased]. Do not cite this merge as "the shell write path is gated."

Issue #4013, opened 2026-08-31
bug(hooks): a write to a target outside every worktree is gated by the payload root's occupancy lease


MScottAdams:
## Summary

A write whose target lies outside every worktree is gated by the **payload root's** occupancy lease. `admitEffectiveHookRoot` returns `admitPayload(payload, null)` when the target has no resolvable worktree, so a destination belonging to no repository -- a path under `%TEMP%`, a home-directory file, anything outside a checkout -- inherits the payload root's occupancy and ritual state. The lease then serialises writes anywhere the agent touches, not only inside the checkout it describes.

Split from #3954 section 3 on the accepted set-level design-critique arc (successor lean https://github.com/deftai/directive/issues/3954#issuecomment-5480338725, completed-arc record 5480360433). Accepted finding 11 of 24: this is neither an identity nor a lifecycle question, it shares no remedy surface with the actor split or with lease release, and holding it inside #3954 made that issue un-landable as one unit.

## Measured at `fb9ece05`

- `packages/core/src/hooks/dispatcher.ts:626-636` has three fallbacks to the payload root: a null write target, no existing ancestor directory, and no worktree for that ancestor. The third is the one that fires for ordinary non-repository destinations.
- Observed live on 2026-08-31T08:15Z: a write to a path under `%TEMP%` was denied naming the primary checkout's occupant (`Directive denied write: Worktree occupied by session host:claude:v1:...`), while the checkout itself was not the target.
- **Not relieved or worsened by #3987.** `isInRepoShellWritePath` (`packages/core/src/hooks/shell-write-targets.ts:240-249`) returns false for any destination resolving outside the project root, so the landed shell-write classifier does not extend payload-root gating to out-of-tree shell writes. The fallback fires on the direct-write and MCP surfaces, before and after the rest of #3987 lands.
- Scale: with 100+ worktrees under `.deft-scratch/worktrees/`, the coupling widens as cohort size grows -- one lease can serialise unrelated scratch writes across a whole cohort.

## Why this is its own issue

The remedy surface is root resolution, not actor resolution. Nothing about who the actor is, or when a lease is released, decides whether a target outside every worktree should inherit the payload root's lease at all. The #3954 arc kept the two actor-resolution sections and the lease-lifecycle work together and split this out for that reason.

## Candidate remedies, not prescribed

1. `effectiveRoot` declines to fall back when the target lies outside every worktree: no root, no gating.
[...]

Issue #3999, opened 2026-08-31
bug(session,swarm): a dispatched subagent claims the occupancy lease and exits without releasing it, stranding the parent's worktree


MScottAdams:
## Problem

A subagent dispatched with `spawn_subagent` claims the worktree occupancy lease and exits without releasing it. Nothing runs on that worker's exit, so the lease is stranded until its TTL expires.

This is a recurrence of the class #3604 closed on 2026-08-26 ("a finished session's lease is never released, so dead locks accumulate"). That fix covers close-out paths that exist for swarm cohorts and for `session:end`. An ad-hoc dispatched worker has neither, so nothing releases on its behalf.

## Measured, 2026-08-30

A parent dispatched a babysit worker to shepherd a documentation PR. The worker did its job correctly -- opened a PR, took a Greptile review to CLEAN 5/5, merged, deleted its branch and removed its worktree -- and exited. It left this behind in the repository root `.deft/occupancy.json`:

```
session_id:   host:grok:v1:MDFhMDU0YjktNDA0Mi03MmUxLTkyOWMtYjZhMTA3NGIzMWUz
claimed_at:   2026-08-30T22:10:39Z
heartbeat_at: 2026-08-30T22:18:13Z
intent:       mutation
```

That session id decodes to `01a054b9-4042-72e1-929c-b6a1074b31e3` -- the dispatched worker, not the parent (`01a05398-…`). The lease was still present at `2026-08-31T01:37Z`, roughly three hours after the worker exited, and was only cleared because the parent ran a deliberate lease sweep at the operator's request.

Worth noting for whoever picks this up: the worker's heartbeat **did** advance (22:10:39 -> 22:18:13), unlike the frozen-heartbeat case in #3987. Different context, different result; that contrast may be useful evidence there.

## What is not wrong here

Two things I checked before filing, so nobody re-investigates them:

1. **The cross-session release is correct behaviour, not a second defect.** `releaseOccupancy` denies a non-owner only when the lease is live: it computes `expired` and `owns`, and refuses solely on `!expired && !owns`. The stranded lease was hours past the 20-minute TTL, so a different session could clear it legitimately. Within the TTL the parent would have been **denied** its own worktree by its own dead child.
2. **The worker did not misbehave.** It followed its envelope, cleaned up its worktree and branches, and reported accurately. No instruction it was given told it to release the lease, and no verb ran on its exit.

## Why this matters more than one stale file

[...]

Issue #3926, opened 2026-08-30
bug(occupancy): fresh single-session containers report an occupied lease, the steal succeeds, and session:start still reports occupied (#3433 friction; candidate root shared with #3921)


MScottAdams:
**TL;DR: in a fresh single-session container the occupancy lease (#3433) reports the workspace as occupied by another session, the agent burns turns stealing it, and `session:start` STILL reports occupied afterwards. Observed in field transcripts (internal test results, private). This is pure friction in exactly the environment where no concurrency exists, and it is a candidate contributor to the elevated session-ritual invocation counts under investigation in #3921.**

**Evidence (field transcript, private).** Verbatim agent reasoning in a fresh container: *"There's an occupancy lease from another session. I need to steal it since we're unattended."* -> *"Need to steal with the correct occupant ID"* -> *"steal worked but session:start still says occupied by the new session."* The agent then continued past it. Turns spent on a lease conflict that cannot exist in a single-session container.

Two defects are visible in that sequence:
1. **A lease is present at all** in a first-run container — either a deposited artifact carries lease state, or the lease is being written and then read back as foreign within the same session.
2. **Post-steal state is still occupied** — the steal path does not leave the stealing session as the recognized occupant, so the condition survives its own remedy.

## Fix direction

- Identify why a fresh deposit presents an occupied lease (stale artifact in the payload vs self-collision on session identity — note #3399 changed session identity resolution, and #3921 is investigating repeated ritual invocation; these may share a root).
- After a successful steal, the stealing session must be the occupant; the current behavior makes the remedy non-terminating from the agent's point of view.
- Unattended/headless posture should not require a steal at all when no live foreign session exists (heartbeat/staleness check rather than presence check — the same presence-vs-provenance pattern flagged elsewhere).

## Acceptance sketch

- Fixture: fresh deposit + first `session:start` in a clean container -> no occupancy conflict reported.
- Fixture: successful steal -> subsequent `session:start` reports the stealing session as occupant, not "occupied".
- Regression: genuine two-session conflict still fails closed as designed.

Relates #3433 (occupancy lease), #3921 (repeated ritual invocation — possible shared root), #3399 (session identity resolution).

MScottAdams:
[...]

Issue #3873, opened 2026-08-28
bug(hooks): on hosts with no session derivation the write gate denies the lease holder, naming that holder's own session id


MScottAdams:
Bound by design-critique arc: successor lean 5466139968, verified-claims table 5466142398, synthesis accepted 5466142496. Round-1 critic 5457370740, Stop 4 retry critic 5466134250. This body carries the bound design; it is the bind product, not a recut.

## Summary

On this host the occupancy write gate denies edits by the session that holds the lease, and the denial names that holder's own session id back at it. Measured end to end and inverted at released `0.108.0` / master `c6761881`: claim a lease, a real tool-call write is denied naming the owner; release the lease, the identical write succeeds; re-claim, denied again.

Hit independently by two workers in the field on 2026-08-28, for the entire duration of both runs. Every `write` / `search_replace` call was refused; all file edits had to be routed through the shell.

## Mechanism, as measured

The host sets **`GROK_SESSION_ID`**. Directive reads **`DEFT_SESSION_ID`**. The names do not match, and `GROK_SESSION_ID` appears nowhere in this tree.

Grok is also absent from `HOST_IDENTITY_PROVIDERS` (`["codex","claude","cursor"]`), and `resolveHookHostIdentity` returns `unsupported` for any host not in that list, so there is no payload identity either. With both sources empty:

1. `evaluateOccupancyWriteGate` resolves the acting id through its own `|| ""` tail -- the **empty string**, not a minted UUID. `resolveOccupancySessionId` is not on this path.
2. `occupancyAdmission` classifies a zero-length id as `stranger`.
3. The live-stranger branch denies and calls `formatOccupancyRemediation`, which interpolates `record.sessionId` -- the occupant -- so the message names the holder to the holder.

The hook is invoked with no `env` block and no `--session-id`, and uses its own process environment.

## Two identity domains for one session

Measured on both paths:

- Shell-invoked CLI verbs **do** read `DEFT_SESSION_ID`: `occupancy:heartbeat` exits 2 unset, exits 0 when set.
- A shell export **cannot** reach the hook, which is a sibling process spawned by the host rather than a descendant of the agent's shell: a real tool-call write stayed denied after exporting in a prior shell.
- Synthetic hook probes confirm the variable is the input when it is genuinely in the hook child: unset denies, owner id admits at occupancy, wrong id denies with identical text to unset.

[...]

PR #4029, opened 2026-08-31
fix(session,hooks): one occupancy actor-resolution chain, four terminals (#3954)

### LINKED ISSUE #3954: bug(session,hooks): claim, release and grant resolve the occupancy actor by three different rules, so the printed recovery is unreachable and a parent locks out its own children
## Summary

The occupancy actor is resolved by three different rules on three surfaces of the same lease. The claim path resolves a host-published owner. The release path does not. The grant path prints a placeholder the party it addresses cannot compute. The result is that a session can claim a lease it cannot release, and a parent can lock out its own children in the same checkout, with every printed recovery unreachable from the state that printed it.

This is the residue measured after #3873 / PR #3948 (squash `41331dea`) landed the identity transport. It is filed separately rather than grown onto that scope, which is delivered.

Origin of the measurements: https://github.com/deftai/directive/issues/3873#issuecomment-5469266268

### 1. Claim and release disagree on who the actor is

`resolveOccupancySessionId` falls through to `ambientHostSessionOwner`, so a claim on a Grok host binds the lease to `host:grok:v1:<base64url(GROK_SESSION_ID)>`.

`releaseOccupancy` (`packages/core/src/session/occupancy.ts:742-743`) resolves the caller as:

```ts
const caller =
  input.sessionId?.trim() || (input.env ?? process.env).DEFT_SESSION_ID?.trim() || "";
```

It never consults `ambientHostSessionOwner`. `packages/cli/src/occupancy-release.ts:56-59` passes `sessionId: undefined`. `git show --stat 41331dea` confirms PR #3948 never touched that file.

Measured against the built modules at HEAD: `releaseOccupancy(root, { env })` returns `action: denied`, `code: 1`, `caller: ""`, and the lease survives. Passing the id explicitly releases cleanly.

Consequence on the installed 0.108.0 runtime, reproduced live: `task session:start` claimed a lease under a minted id, the next tool-call write was denied naming that same id, and `task occupancy:release` minted another id and was refused as a stranger. The only recovery that worked was `task occupancy:release -- --session-id <the-id-the-deny-reported>`. The deny text says "The occupant may release (`occupancy:release` / `session:end`)", which is the one thing the occupant cannot do.

### 2. A parent locks out its own children

`GROK_SESSION_ID` is per-agent-session and is **not** inherited by subagents. Two Grok sessions on the same workspace seven seconds apart derived different canonical owners, and the second session's write against the first session's live lease was denied.

This differs from Codex, where the host identity source is documented as shared between parent and subagents (`packages/core/src/hooks/classify/host-session-identity.ts`). So on Grok the lease is per-agent, while the worktree it guards is per-cohort.

The printed remediation is `occupancy:grant --child-session-id=<your-session-id>`, run by the occupant. The child cannot supply its own canonical string without base64url-encoding the environment variable itself, and the occupant does not have it.

### 3. An out-of-tree write is gated by the payload root's lease

A write to an OS temp path was denied by the primary checkout's lease. An out-of-tree target has no worktree of its own, so `effectiveRoot` falls back to the payload root. The lease is therefore not confined to the checkout it describes: it serialises writes anywhere the agent touches. With 100+ worktrees under `.deft-scratch/worktrees/`, that coupling widens as cohort size grows.

## Why this is not a #3873 regression

#3873's bound acceptance covers the claim-and-present loop, and that loop works: claiming under the host owner and presenting the same owner yields an occupancy admission, with an unidentified actor still denied. None of its six clauses covers the release path, the parent/child case, or out-of-tree gating. This issue is the part that was never in scope, surfaced by operating the fixed design.

1. Whether release should accept the recorded occupant id from the lease when the caller can present no identity, or whether that weakens the fail-closed posture the #3611 decision record protects.
2. Whether the parent/child case is an identity problem (children should derive a related owner) or a membership problem (`occupancy:grant` should be automatic within a nuclear-family cohort, #3155).
3. Whether an out-of-tree write target should be gated by the payload root at all, or whether `effectiveRoot` should decline to fall back.
4. Whether the three surfaces should share one actor-resolution function, and what that does to the locked unsupported-host tests.
5. Whether any of this is safe to change before the identity transport ships in a release; the fix is merged but unreleased, so operators are currently on the pre-fix runtime.

## Design-critique

`mechanism-shaped: true`

Refs #3873, #3611, #3755, #3433, #3794, #3155, #3156.

MScottAdams:
## What this fixes

Claim, release, heartbeat and grant/revoke resolved the occupancy actor by three different rules. On a host that publishes its own owner, `session:start` claimed under the host owner while `occupancy:release` resolved an empty caller — so a session was refused by the lease it had just taken, and the printed recovery ("the occupant may release") was the one thing the occupant could not do. The only working form was to copy the id out of the deny text.

The four surfaces now share one lookup order — explicit `--session-id`, then `DEFT_SESSION_ID`, then the owner the running host published — and keep four terminals.

## The eight acceptance items

**1. Shared lookup chain, per-surface terminal.** `resolvePresentedIdentity` is the shared chain. `resolveOccupancySessionId` delegates to it and adds the mint, so claim still establishes an identity where none exists. `releaseOccupancy`, `heartbeatOccupancy`, `grantOccupancyMembership`, `revokeOccupancyMembership` and `releaseSwarmOccupancy` take the empty terminal and keep their existing no-identity diagnosis. `resolveOccupancySessionId` was **not** adopted wholesale and the mint was **not** deleted — its terminal would have replaced "this process presented no session identity" with a plausible id no later hook presents, on every host that publishes no owner.

**Disagreement.** The chain also reports it. When `DEFT_SESSION_ID` and the host owner name different sessions, the documented order still stands — reordering it would break the injection path the lifecycle rewrite depends on — but a refusal now names both ids and the `--session-id` that would act as the host owner. That is the middle row of the measured matrix, which previously produced a state whose remediation the presenter could not act on.

**2. `occupancy:grant` identity transport.** Added to `EXACT_LIFECYCLE_VERBS` and `DIRECT_LIFECYCLE_VERBS` with a `LIFECYCLE_ARGUMENT_POLICIES` entry covering `--child-session-id`, `--role`, `--ttl-minutes`, `--host`, `--address`, `--join-protocol` and the `--revoke` boolean. `--worktree` rebinds the tree the grant covers, so it joins `--project-root` outside the auto-approved rewrite surface. It was the only occupancy lifecycle verb with zero transport on any host.

**3. Child-id validation.** A child under the reserved `host:` prefix must be a well-formed canonical owner, checked against the same `CANONICAL_OWNER_PATTERN` the rewrite bridge uses — the pattern moved beside the provider table in `session/host-session-owner.ts` so the identity surface, the rewrite surface and grant validation cannot drift. A canonical child that re-prefixes the owner's own host payload under a second provider is refused as the self-grant it is; that is the cross-prefix laundering case, which passes a shape check and names a child no session on that provider can present. Round-trip encoding is enforced, so one session cannot have two canonical strings.

An **opaque** child id (no `host:` prefix) stays admissible, deliberately: a child on a host with no identity contract presents whatever `DEFT_SESSION_ID` holds, so refusing every non-canonical id would deny a grant nothing has measured wrong. Revoke skips the check so a malformed grant written earlier stays withdrawable.

**4. Open question 1, answered against anonymous recorded-occupant release.** Unchanged authority, now deliberate and recorded: reading the occupant out of the lease file and releasing it would make possession of that path into authority to delete a live lease, and the cooperative bearer model has no second check behind that refusal. Recorded in `xbrief/decisions/2026-08-31-occupancy-actor-resolution-split.decision.json`, in `commands.md`, and on `releaseOccupancy` itself. The unreachable-recovery symptom is fixed by the ambient step in item 1 instead.

**5. Open question 2, answered per identity-source kind.** On `host-env` hosts parent and child are different actors, so the answer is **identity**, not automatic membership: each side resolves its own owner and claims its own worktree. Automatic membership was rejected on measured capacity — 32 grants at a four-hour expiry against a twenty-minute lease exhausts a busy parent's lease inside a day — and because it is half-built. The revocation trigger is the owner's `--revoke` or expiry; release-on-child-terminal is dispatcher lifecycle and stays on #3999. On `payload` hosts parent and subagents share one id, so there is nothing to grant and the live consequence is the inverse: a parent's release removes a working child's lease mid-flight with no denial. Recorded, not changed — it is a property of shared host identity.

Because the answer is not membership, claim and heartbeat stay owner-only. What did change is the diagnosis: a granted child refused at `session:start` now reads that claim is owner-only, instead of being offered the write grant it already holds. Heartbeat already routed through the member refusal; claim did not.

**6. Question 4's "locked unsupported-host tests" blocker is struck.** It does not exist as stated. `occupancy-release.test.ts`'s "without ambient inheritance" test never set a host variable and never asserted a denial; it now sets one and asserts the explicit id still wins, so the title and the body agree. The real condition is hermeticity: the three occupancy CLI harnesses scrub `HOST_ENV_IDENTITY_VARIABLES` (derived from the provider table, so a new host-env provider is scrubbed without a test edit) and core prove-surface call sites pass `env` explicitly. This was not theoretical — two CLI tests went red on a host publishing `GROK_SESSION_ID` the moment the ambient step landed, and green again after the scrub. #3877 is untouched and not made worse.

**7. Tests.** Bare release under an ambient host environment resolves the occupant; anonymous release is refused with the empty-caller diagnosis; a disagreement is named; grant refuses three shape-invalid children and the cross-prefix laundering while admitting an opaque one and a genuinely different session on the same provider; a malformed pre-existing grant stays revocable; member behaviour at claim is asserted per item 5. Baseline green: 84 files / 1127 tests across `packages/core/src/session`, `packages/core/src/hooks` and the occupancy CLI suites, plus `task check` exit 0.

**8. Out of scope, recorded.** No change to `admitEffectiveHookRoot` or the payload-root fallback (#4013). No dispatcher-side lease-release lifecycle (#3999). No release verdict.

## Gate integrity (#3156)

No refusal is weaker than before. Nothing that would have widened authority was taken: anonymous recorded-occupant release is refused, automatic membership is refused, claim and heartbeat stay owner-only, and the mint stays where it was. Two behaviour changes are additive and stricter — grant now refuses child ids it used to accept, and `occupancy:grant` on a wire-capable host now requires a resolvable owner like every other lifecycle verb.

One test title changed, and its assertion got **stronger**: `occupancy-release.test.ts` "parses and uses an explicit host session id without ambient inheritance" became "…ahead of every ambient source", and the body now sets a competing host variable and asserts the release actually removed the lease. Nothing was relaxed to go green; the two red CLI tests were fixed by making the harness hermetic, which is item 6's own acceptance condition.

## #3873 — what remains, and who must supply it

`Refs #3873`, not `Closes`. Its acceptance needs a clause that cannot be satisfied from this tree.

The merged #3873 transport makes the **CLI** resolve the host owner. The write gate runs in the `deft-hook` **sibling process**. On the deployment measured in #3954 comment 5482415424 and its #3873 mirror 5482789562, that process does not receive `GROK_SESSION_ID` and presents an inherited `DEFT_SESSION_ID` naming another host's session — so the deny names a *foreign* id, not the caller's own, and a test asserting the deny names the caller's own session passes in a fixture and misses the field case.

Nothing in this repository can set that process's environment: the hook is a sibling the host spawns, not a descendant of the agent's shell, and a shell export cannot reach it. The hook also writes no env or payload log, so the condition is inferred from the deny text rather than observed. **Who must supply it:** either the host (publish `GROK_SESSION_ID` into the `deft-hook` sibling process), or a Directive-side hook-process capture facility (`DEFT_HOOK_PAYLOAD_LOG` or equivalent) that makes the observation possible at all. Both are separate work.

What this PR does instead is make the prove-surface ambient step handle the divergence, which is the reachable half: the occupant can now release its own lease from its own shell without lifting an id out of the deny. The boundary is written down in `commands.md` § "Where the transport still does not reach".

## #3926

`Refs #3926`, never closed. Its identity and steal half is dominated by item 1; its remaining content is the fresh-container provenance question, which is not in this scope. Its measured facts were taken as inputs and none was "fixed": the mint is the terminal branch, a confirmed steal that names the occupant terminates correctly, one that does not is refused with a config code, and staleness is already TTL-plus-twelve-hour-cap with claim-over-expired working.

## Verification

- `task check` — exit 0
- `task verify:forward-coverage`, `verify:test-boundary`, `verify:scope-provenance`, `verify:encoding`, `verify:branch` — all exit 0
- `task xbrief:preflight` — exit 0
- Per-item `x-directive/evidence` stamped on the brief's eight acceptance items

Closes #3954
Refs #3873, #3926, #3999, #3755, #3611, #3156


MScottAdams:
Addressed the SLizard P1 on `HOST_ENV_IDENTITY_VARIABLES` in `00194a4d` rather than waiving it as advisory: `ambientHostSessionOwner` now iterates the same resolved host-env entries the exported list projects from, so the production scan and the surface a caller scrubs cannot name different providers, and the host-env filter is written once instead of twice. The test asserts the projection against the production scan.

Greptile on `00194a4d`: 5/5, P0=0, P1=0, advisory should-not-merge false, CI 25/25.

PR #4031, opened 2026-08-31
fix(session,swarm): release dispatched child occupancy on terminal

### LINKED ISSUE #3999: bug(session,swarm): a dispatched subagent claims the occupancy lease and exits without releasing it, stranding the parent's worktree

MScottAdams:
## Summary

A dispatched subagent that claims a worktree occupancy lease and exits no longer strands that lease until TTL. At dispatch the parent records the child occupancy owner and exact worktree root under .deft/child-occupancy/ (lease-gated; not .deft-scratch). The existing orchestration terminal observation (phase: terminal) compare-and-releases under the occupancy lock only when that recorded child still owns that recorded tree. A successor owner is left alone.

On host-env hosts this is the missing lifecycle event. On payload hosts parent and child share one id, so the same transition is a no-op (auto-release would drop a live parent lease mid-flight). Swarm close-out still releases only the launcher occupancy_session_id. No new reap verb.

Out of scope: the 20-minute TTL; #3987; #3873 (hook-process GROK_SESSION_ID observation); actor-resolution (#3954); payload-root fallback (#4013).

Closes #3999


MScottAdams:
ended_at: 2026-08-31T22:50:22.736Z

PR #4048, opened 2026-09-01
fix(lifecycle-visible): skip empty check-ignore pattern fields

### LINKED ISSUE #4010: [framework-gap] verify:lifecycle-visible reads a blank .gitignore line as a match-all pattern
## Summary

verify:lifecycle-visible reads a blank .gitignore line as a match-all pattern

## Context

Directive engine 0.107.0, content 0.107.0. Consumer project with a conventional multi-section `.gitignore` — blank lines between groups, comment headers, a trailing newline.

The lifecycle-visible check parses `.gitignore` to decide whether lifecycle artifacts are tracked.

## Expected behavior

Blank lines in `.gitignore` are skipped. Git's own rules are explicit that a blank line matches no file and serves only as a separator, so a `.gitignore` that is entirely valid and conventional should not affect the check's verdict.

## Actual behavior / gap

A blank line is treated as a pattern rather than skipped, and an empty pattern matches everything — so the check concludes that every lifecycle path is ignored.

Observed on `deft session:start` in a repo whose lifecycle roots are all tracked:

```
[deft lifecycle-visible] hidden xbrief/proposed/  (.gitignore:51:)
[deft lifecycle-visible] hidden xbrief/pending/   (.gitignore:51:)
[deft lifecycle-visible] hidden xbrief/active/    (.gitignore:51:)
[deft lifecycle-visible] hidden xbrief/completed/ (.gitignore:51:)
[deft lifecycle-visible] hidden xbrief/cancelled/ (.gitignore:51:)
[deft lifecycle-visible] hidden vbrief/proposed/  (.gitignore:51:)
...
[deft lifecycle-visible] ADVISORY: lifecycle roots must stay visible to git
  (warn-only; task verify:lifecycle-visible -- --enforce to fail closed).
```

Line 51 of that `.gitignore` is blank. The attribution string `(.gitignore:51:)` prints the empty pattern after the line number, which is the tell — every reported path is attributed to the same empty pattern. `git check-ignore -v xbrief/plan.xbrief.json` returns nothing and `git ls-files` lists the lifecycle artifacts, so git itself agrees they are tracked.

### Why it matters

The failure is silent and inverted: the check reports that lifecycle artifacts are invisible when they are tracked normally. Because blank lines are used for readability in essentially every non-trivial `.gitignore`, the trigger is common and unrelated to anything the consumer did wrong — which makes it hard to attribute. A consumer's first instinct is to look at their lifecycle folders, not at whitespace in an unrelated file.

It is warn-only today, so the cost is a false advisory on every mutation `session:start`. It would be a hard stop under `--enforce`, which is the mode the advisory text points at.

### Suggested fix

Skip blank and whitespace-only lines when parsing, and skip `#` comment lines if that is not already handled. Treat an empty pattern as a parse error rather than as match-all, so a future parsing gap fails loudly instead of silently swallowing the whole tree.

A regression fixture with a `.gitignore` containing blank lines, comments, a negation and a trailing newline would cover the family.

### Not classified as a blocker

The check is warn-only, so the flow completes.

### Dedup performed

Searched open issues in this repo on 2026-08-31 for `lifecycle-visible gitignore`, `gitignore blank line`, `verify:lifecycle-visible`. No match.

---
_Filed via `task feedback:file` from a directive consumer project (Refs #1709 value-feedback gap escalation)._

MScottAdams:
## Summary

`verify:lifecycle-visible` treated an empty `git check-ignore -v` pattern field as a match-all hide. The trigger is a `.gitignore` line of three ASCII spaces; empty and tab-only lines do not reproduce. Git emits `<source>:<line>:` with an empty pattern. The ignore-file reader already skips blank/whitespace/comment and never decides the verdict, so a parser skip-blank fix is a no-op.

`collectIgnoredRoots` now drops an empty pattern field instead of rendering it as a rule. Live fixture: three-spaces line plus empty and tab controls. Does not add a second in-process gitignore engine.

## Related Issues

Closes #4010

## Checklist

- [x] `/deft:change <name>` — N/A (3 files: product skip, tests, CHANGELOG; not cross-cutting or architectural)
- [x] `ROADMAP.md` — N/A (bug fix, not a tracked roadmap item)
- [x] Tests pass locally (`task check`: 1108 passed, 6 skipped)

## Post-Merge

- [ ] **Verify issue auto-close**: After squash merge, confirm referenced issues actually closed.
- [ ] Enable branch protection on `master` requiring CI status check (one-time setup, see #57) — N/A, already enabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Design-critique parent silence now waits on this round's siblings, not each critic EXIT (#4027).** Bind was already withheld while same-round siblings remained unposted; the lean and verb menu still talked at 1 of 3. Silence is the existing unqualified same-round-unposted conjunct. Spend stays permission. The critic brief gets no `parent-silence-until` field. Line 218 keeps the empty-lean verb-menu ban and waits on this round's lean. A lean that closes a round of two or more names the dispatched sibling count and those comment ids. Closes #4027. Refs #3850, #3741, #3962.
 
+- **`verify:lifecycle-visible` no longer treats an empty `git check-ignore -v` pattern as a match-all rule (#4010).** A `.gitignore` line of three ASCII spaces makes git emit `<source>:<line>:` with an empty pattern field; that is not a rule. Empty and tab-only lines do not reproduce. The ignore-file reader already skips blank/whitespace/comment and never decides the verdict, so a parser skip-blank fix is a no-op. `collectIgnoredRoots` now drops an empty pattern field instead of rendering it. Fixture: three-spaces line plus empty and tab controls. Does not add a second in-process gitignore engine. Closes #4010.
+
 ### Removed
 
 ## [0.109.1] - 2026-09-01
--- a/packages/core/src/lifecycle-visible/evaluate.test.ts
+++ b/packages/core/src/lifecycle-visible/evaluate.test.ts
@@ -95,6 +95,15 @@ describe("parsers (#3505)", () => {
     expect(parseCheckIgnoreVerboseLine(":1:\tpath")).toBeNull();
   });
 
+  it("parses an empty pattern field from check-ignore -v (#4010)", () => {
+    expect(parseCheckIgnoreVerboseLine(".gitignore:51:\txbrief/active/")).toEqual({
+      source: ".gitignore",
+      line: 51,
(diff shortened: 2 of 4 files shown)


MScottAdams:
owner: MScottAdams
monitor_agent_id: leaf-4010
head_sha:
started_at: 2026-09-01T15:01:00.189Z
expires_at: 2026-09-01T15:31:00.189Z
platform_primitive: spawn_subagent

Issue #3558, opened 2026-08-20
bug(banking): ac-pass bank writes regressed 0.104->0.105 and the serve path has never hit in the field (15/15 executed) - #3549 branch-1 fix; emit miss_reason on every executed walk


MScottAdams:
**TL;DR: the branch-1 fix issue from #3549. Two defects, one surface: (1) bank WRITES regressed 0.104 -> 0.105 — a task that banked on every verified pass in 0.104 banked zero times in 0.105 (its apps sibling banked only 3 of 8 scopes); (2) the SERVE path has never hit in the field — `served_from: "executed"` on 15/15 observed walks across both bundled trials. Until both land, #3387 is dead weight: all of the completion-walk cost, none of the reuse.**

**Evidence:** #3549 disposition comment (field JSONLs, internal benchmark, private). Code anchors at HEAD: `packages/core/src/session/ac-pass-reuse.ts` (miss conditions: no scope id, `hashed.complete === false`, epoch-zero bankedAt, hash mismatch), `product-first-done-gate/evaluate.ts:240,895` (hash passed at write), `session/product-state-hash.ts` (scope/exclusions correct).

## Fix
1. **Write regression:** bisect the 0.104 -> 0.105 delta on the bank-write path — prime suspects are the #3399 envelope migration (ac_pass_bank moved onto `RunSummaryEmitter`) and #3387's own landing. A verified-pass walk MUST produce a bank write; the 0.104 field stream is the behavioral fixture.
2. **Serve miss:** instrument-then-fix. The reuse gate returns a `reason` on every miss — ensure that reason reaches the acceptance event (`served_from: "executed", miss_reason: ...`) so the next field stream names the culprit outright. Fix the named cause; leading candidate is hash incompleteness in containers (if `git status` quirks force `complete:false`, the fallback hashing the glob work already built must engage rather than defaulting to miss-forever).
3. **Fixture (per #3362 discipline):** a trial-shaped test — activate -> verify pass (banks) -> complete (serves from bank) -> check (serves from cache) — asserting served_from values bank/cache, not executed, and exactly one command execution.

## Acceptance sketch
- Field-shaped fixture passes; a product edit after banking forces executed (fail-closed path intact).
- Next external run's streams show non-zero bank/cache serve rates on multi-scope tasks, with `miss_reason` on every remaining miss.

Relates #3549 (investigation this closes the loop on), #3387 (the shipped surface), #3285, #3399 (suspect), #3352.

PR #4651, opened 2026-09-16
fix(verify-ac): share resolveScopeIdForAcReuse at bank write and lookup (#4631)

### LINKED ISSUE #4631: bug(verify-ac,banking): acceptance walk runs without a scope id on derived none_stated walks - bank unkeyable, silent permanent miss
**Observed** (internal benchmark evidence, private; 0.119.0, run-level aggregation plus per-event detail from the operator): 3 of 63 acceptance walks missed the bank with `miss_reason: "no scope id"`, across two trials - one trial twice in the same session (seq 6 and 9, dual verify-ac), one trial once. Identical signature on all three: `served_from: executed`, `source_rung: derived`, `none_stated: true`, `resolved_command_count: 0`, `outcome: empty-pass`. The walk RAN to completion and still reported no scope id, so the bank was unkeyable: a walk in this state can never hit, silently, on every invocation.

**Why it matters beyond these three events.** Today the cost is nil - these are empty-pass walks with nothing bankable anyway. But the keying gap is independent of the emptiness: if derivation ever produces real commands on a scope-less walk, the result would execute, verify, and never bank - reproducing the #3549 starvation precisely on walks where the bank could pay. It also muddies the miss taxonomy: "no scope id" and "no reusable result" describe different failures (cannot look up vs looked up and found nothing), and only the second is expected on this class of walk.

**Suggested:** verify:ac resolves or synthesizes a scope id before the bank lookup, or refuses loudly when it cannot, and emits which path it took. A fixture asserting "every completed walk carries a scope id" would pin it.

**Related doc nit, same dataset:** walks ending `config-error` skip the bank lookup entirely and emit no miss_reason (observed twice, both at `source_rung: project_floor`, `none_stated: true`) - coherent behavior, but worth one schema doc line so stream consumers do not misread the executed-count vs miss_reason-count gap. Refs #3549, #3558, #3285, #3928, #4626.

MScottAdams:
## Summary

verify:ac FromPath write and lookup now share resolveScopeIdForAcReuse. Id-less walks synthesize the collision-aware oracle path (rel path, not basename stem), so a completed walk never misses no-scope-id. Empty-pass stays unbankable. #3558 bank-on-verified-pass remains.

Write site maybeAttachAcPassBank calls the same resolver. Sidecar resolveScopeIdForAcReuseDetailed reports which path ran.

## Test plan

- npx vitest run packages/core/src/session/ac-pass-reuse.test.ts
- npx vitest run packages/core/src/product-first-done-gate

## Documentation impact

change_class: none
surfaces: none
rationale: "CHANGELOG [Unreleased] Fixed records the miss_reason taxonomy; no registered command, skill, help key, or docs-site page changed."

Closes #4631


MScottAdams:
owner: MScottAdams
monitor_agent_id: 01a0ac7a-bc7b-7d62-a037-8fd88c33dd35
head_sha: e95815bbd5b3e6d9e58b157d36b2677dd7ceca38
started_at: 2026-09-16T23:21:31.915Z
expires_at: 2026-09-16T23:51:31.915Z
platform_primitive: spawn_subagent

MScottAdams:
Review-monitor batch 1 (Approach 1, spawn_subagent).

Class A fixed: Outside Paths Share Keys — collisionAwareOracleRelPath keeps `..` / other-volume paths instead of basename; resolveOracleScopeKey uses the same helper.

Class B parked: Completion Loses Scope Key (evaluate.ts FromPlan / scope:complete without path). Critic 5703424901: FromPlan without a path is not P1; do not widen harvest to FromPlan. Threading that lookup would expand into acceptance-evidence.ts.

P2 Integration Path Remains Untested: non-blocking.

MCP unavailable in this grok-build session — used `gh api` REST fallback for pulls/4651/comments.

Issue #4718, opened 2026-09-17
bug(doctor): after successful update, within-window engine>pin still recommends npx update (ignores pnpm; pin/lock is the real gap)


polfer:
## Summary

On a pnpm consumer after a successful global upgrade + `directive update`, doctor still prints a primary next action of `npx @deftai/directive update` even though:

1. deposited content already matches the running engine, and
2. the remaining gap is the committed `package.json` pin (+ lock follow-through), and
3. the repo declares `packageManager: "pnpm@…"` / has `pnpm-lock.yaml` (so the ephemeral form should be `pnpm dlx`, not `npx`).

This is a **wrong remediation loop**, not a missing deposit. Related leftover pin write is #4710; related beyond-window deadlock/print honesty is #4692. This issue is the **within-window + already-updated content + package-manager-blind Next command** shape.

### State after operator followed the documented path

| Layer | Version | Source |
| --- | --- | --- |
| Global engine on PATH | `0.119.2` | `npm i -g @deftai/directive@latest` |
| `.deft/core/VERSION` / `.deft/GENERATION.json` / `xbrief/.deft-version` | `0.119.2` | `directive update` (success, `deposit_completed: true`) |
| `package.json` pin `@deftai/directive` | `0.119.0` | unchanged by update |
| `pnpm-lock.yaml` | `@deftai/directive@0.119.0` | unchanged by update |
| Local `node_modules` (pnpm) | `0.119.0` | from prior `pnpm install` |

Doctor then printed:

```text
Version reconciliation: engine 0.119.2, content 0.119.2, pin 0.119.0 -- .deft/core/VERSION (0.119.2) is ahead of the package.json pin (0.119.0); engine 0.119.2 is ahead of pin 0.119.0 within the skew window; proceeding after content update

⚠ Primary next action ...
  Root cause: engine 0.119.2 is ahead of pin 0.119.0 within the skew window
Next command: npx @deftai/directive update
  Does / why safe: Forward-migrate content with `directive update`, then run the gate.
```

Re-running that Next command cannot clear the warning: content is already current; pin/lock were never the update writer's responsibility today.

### Code evidence (installed `directive-core@0.119.2`)

1. **Within-window always forces `update`, even when content is already ahead of pin**

`resolution/skew-policy.js`: `engine > pin` within window → `decision: "proceed-loud-update"`, `requiresUpdateFirst: true`.

`resolution/plan.js` (default branch):

```js
[...]

Issue #4710, opened 2026-09-17
bug(update): update must write the consumer pin; doctor-names-file is not reconstitution


MScottAdams:
# Update must write the consumer pin

Leftover of #4692 after blocking pain-audit 5716925951 and Handoff 5717045800. Operator confirmed the leftover split on 2026-09-17. Do not ingest #4692 as pin reconstitution closed.

## Summary

A Bound-remedy that says update writes the pin **or** doctor names `package.json` does not cover reconstitution. A next-build worker can keep the doctor-names arm, print a path, and leave every clone with a lagging pin after `directive update`. Pin reconstitution is: update writes the consumer `@deftai/directive` pin as part of the upgrade it performed. Fail-closed pin/engine skew is a doctor warning at edf1d698, not the write-gate unlock. Do not add a doctor-prose PreToolUse carve-out.

## Measured (dispatch-sha edf1d698aafcf956cff08ad3ad14ff608505861a)

- `ensurePackageJsonPin` lives in `packages/core/src/init-deposit/scaffold.ts` and is called from init. `runRefreshDeposit` (`refresh.ts`) writes hooks and does not write the pin.
- `classifyUpdateState` maps any non-proceed `plan.mode`, including fail-closed `blocked`, onto `updated`, so beyond-window skew does not stop the writer. The writer still leaves the pin behind.
- `packages/core/src/resolution/plan.ts` fail-closed `nextAction.command` is the literal `directive <gate> --accept-engine-jump`. packages/cli has zero matches. `DEFT_ACCEPT_ENGINE_SKEW` is a library escape and does not yield proceed.
- `WRITE_GATED_REQUIRED_STEPS` is agent_hooks then doctor. `WRITE_GATED_EXECUTE_STEPS` is agent_hooks only.

## Desired shape

- `directive update` / `runRefreshDeposit` writes the consumer `@deftai/directive` pin as part of the upgrade it performed. No doctor-names-file OR. Pin write is reconstitution identity (#2264 / #2269), not `lastErrorCount`.
- Printed recovery never proceeds without a pin match. If doctor prints a next command, it is parsed argv that yields a pin write, or it names `DEFT_ACCEPT_ENGINE_SKEW` without claiming proceed. Do not add a doctor/session-ready flag whose only effect is proceed-without-pin-match. Literal `<gate>` is not runnable argv.
[...]

Issue #4692, opened 2026-09-17
0.114.0 to 0.118.1 upgrade deadlocks: update does not bump its own pin, and the printed recovery flag does not exist


nheroux-bit:
Upgrading a consumer project from the 0.114.0 deposit to 0.118.1 left the install in a state where **every agent write was blocked and no documented recovery could clear it**. Four defects compounded; each is independently reproducible.

Environment: Windows 11, git-bash shell, node v24.18.0, engine `@deftai/directive-core@0.118.1`, hybrid vendored install (`.deft/core/` + managed AGENTS.md), consumer deposit.

## 1. `update` does not carry its own pin forward, so doctor can never clear

`directive update` brings the deposit to 0.118.1 — it rewrites `.deft/core/**`, `.deft/GENERATION.json`, `xbrief/.deft-version`, and the host hook files. It does **not** touch the `@deftai/directive` pin in the consumer's root `package.json`.

doctor then fails closed on precisely that mismatch:

```
Version reconciliation: engine 0.118.1, content none, pin 0.114.0
  -- engine 0.118.1 is a large jump ahead of pin 0.114.0; failing closed.
```

Because the gated `doctor` ritual step blocks writes, and the remedy is a write to `package.json`, `directive update` cannot fix the condition it creates. Ran it twice on a clean tree; the pin stayed `0.114.0` both times.

Bumping the pin by hand is what resolves it:

```
✓ Resolution: proceed -- engine 0.118.1 matches pin 0.118.1 and content is current.
```

**Expected:** `update` bumps the consumer's pin as part of the upgrade it just performed, or doctor names `package.json` explicitly as the file to edit.

## 2. The recovery flag doctor prints does not exist

doctor's own "next action" output:

```
Next command: directive <gate> --accept-engine-jump
  Does / why safe: Confirm the jump with --accept-engine-jump or DEFT_ACCEPT_ENGINE_SKEW=1 after reviewing the delta.
```

But no command accepts it:

```
$ directive doctor --accept-engine-jump
✗ Unknown flag(s): --accept-engine-jump
ℹ Allowed: --session, --fix, --repair, --repair-taskfile, --json, --quiet, --full,
           --network, --project-root, --force, --openclaw-all-agents, -h, --help

$ deft session:ready --session-id=... --accept-engine-jump
session_ready: unrecognized argument: --accept-engine-jump
```

[...]

PR #4734, opened 2026-09-17
fix(doctor): align-pin when content matches engine and pin lags

MScottAdams:
## Summary

After a successful `directive update`, within-window engine>pin still recommended another deposit update even when deposited content already matched the engine. The remaining gap is the committed pin plus lock.

This names ResolutionMode `align-pin` (schema lockstep), maps it so `directive update` treats that mode as current (no deposit refresh), threads doctor detectPackageManager (stripped env, field then lock, no user-agent) into `plan()`, and prints `renderProjectInstall` at the engine spec.

Pin write inside update stays #4710. Print honesty / hook one-schema stays #4692.

Bound harvest: successor lean 5719231055.

## Docs impact

change_class: change
surfaces: command:doctor
rationale: "Doctor primary Next command for content-at-engine plus pin lag is pin follow-through via the detected package manager, not another deposit update."

Tracking: #4718
Refs #4718


MScottAdams:
ended_at: 2026-09-17T22:20:41.023Z

